PR #99, opened 2015-02-25
Save currently selected element when page is changed

### LINKED ISSUE #81: Final value not updated after navigating to next student
If you specify some score in the text box, then click the link to go onto the last student, that last value entered is not preserved. No issues if you click somewhere else on the page, then click the navigation link.

jhamrick:
Fixes #81


jhamrick:
Travis thinks this has failed, but looking at the individual tests, they've all actually passed. Something is currently up with travis, I think, so I'm going to go ahead and merge this.

PR #100, opened 2015-02-25
Add buttons for full and no credit

### LINKED ISSUE #82: 'Full credit'  and 'No credit' buttons on grading bar
In addition to being able to add the numerical score, there should be two buttons that assign full credit and no credit. Default tabbing to the numerical score still makes sense though.

jhamrick:
Fixes #82 and changes the formgrader to look like this:

image

@smeylan is this what you were thinking of?


jhamrick:
FYI, travis thinks that this is still building but if you click on the tests then it appears to have passed.

PR #124, opened 2015-03-05
Build docs with travis

### LINKED ISSUE #117: Come up with a better solution for building docs
Docs that are in master should be just the manually written content, with nothing generated (i.e., IPython notebooks with *no* output from the cells). We should actually compile the docs and commit the output to a different branch, and maybe we can set up something on Travis to do this automatically every time there is a commit to master. This will help to minimize merge conflicts and also remove the problem of having to remember to regenerate the docs all the time.

jhamrick:
This is an attempt to close #117


.travis.yml
```diff
@@ -16,8 +19,10 @@ install:
...
   - cd ipython && git submodule update && pip install -e ".[notebook]" && cd ..
   - pip install -r requirements.txt
-  - pip install nose requests!=2.4.2
+  - pip install nose requests!=2.4.2 fabric
   - python setup.py install
   - if [[ $GROUP == 'js' ]]; then cp nbextensions/* $(ipython locate)/nbextensions/; fi
 script:
   - ./scripts/runtests $GROUP
+after_success:
+  - if [[ $GROUP != 'js' ]]; then fab travis_docs; fi
```
Carreau: Do you want to build only on non-PR and master branch like [here](https://github.com/ipython/ipython-website/blob/master/.travis.yml#L20) ?

Carreau:
Also you might want to have a look at [invoke](http://www.pyinvoke.org/) which is the local only part of fabric.

.travis.yml
jhamrick: Yeah, just updated it to only do not-PRs and I will update it to only do it on master just as soon as I have it working :smile:

jhamrick:
Yeah, it might be better at this point to use invoke, but I think I'll leave that for a different PR.

Carreau:
Isn't that like infinite loop ? 

builds docs branch, push on docs, triggers travis to build new commit ?

jhamrick:
Travis won't build the `docs` branch; there's a `.travis.yml` file on there already telling it to ignore that branch. The idea is that (once it actaully works) it will only build on master, and from there push the `docs` branch.

Carreau:
Oh, ok, I would have expected this `branch:only:master` to be in the travis.yml in all branches.

jhamrick:
I will add it to the other branches once this is working on this branch, then I don't have to make a whole bunch of commits to master while I'm getting it to work.

But, bah, it doesn't seem actually be executing the fab command, and I'm not sure why...

Carreau:
`$TRAVIS_PULL_REQUEST == false` ?

Carreau:
Or this is a PR ?

Carreau:
I had to mess with master also when I did that with IPython-website repo too.

jhamrick:
Oh, duh, yes, this is a PR. Clearly, I should be sleeping, and not coding right now :smile:

jhamrick:
Closing this to work on the jupyter/nbgrader branch instead.

Carreau:
At least the PR was helpful to review that this was indeed a PR.

PR #125, opened 2015-03-05
Build docs

### LINKED ISSUE #117: Come up with a better solution for building docs

jhamrick:
This is an attempt to close #117

Continuation from #124


Carreau:
OK, this is still technically a PR, so I don't know how it will behave. It seem to show two parallel travis build...

jhamrick:
Yep, this is a PR, but it's also a branch on the jupyter repo (rather than my fork), so it should build the docs on the push build but not the PR build.

Carreau:
> Yep, this is a PR, but it's also a branch on the jupyter repo (rather than my fork), so it should build the docs on the push build but not the PR build.

Nop. It choke on `rm -r *` try to remove the `.git` folder (weird) and no `-f` mean it's waiting for user to pres `y`...

Carreau:
Oh, it's actually choking on `ipython/.git` folder.... I guess we should clone in ~/ not in ~/nbgrader. I'll make a PR.

jhamrick:
Thanks Matthias! Almost there... now apparently it's dying because fabric doesn't do python 3? But all the docs are in python 3 notebooks, sigh. Ok, maybe it is time to switch to invoke, which I'm pretty sure does work on python 3.

jhamrick:
Yesssss it finally works!

Carreau:
:shipit: :ship: :+1:

PR #126, opened 2015-03-05
Clone ipython master in home and not subdir

Carreau:
Let see if this fix the docs building.


Carreau:
Oh, actually I have the rights to merge... Let see if this works.

PR #142, opened 2015-03-14
Optimize SQL queries

jhamrick:
This turns most of the `@property` methods in the declarative mapping into actual queryable columns. This adds a slight speed improvement. On top of that, this also moves all the methods for calculating average scores into the gradebook, which then uses proper joins and stuff for massive speed improvements. Additionally, I added a few helper functions for querying information about all the students or submissions at once, which dramatically speeds up load times in the formgrader.

### DIFF
--- a/nbgrader/api.py
+++ b/nbgrader/api.py
@@ -48,11 +48,8 @@ class Assignment(Base):
             "name": self.name,
             "duedate": self.duedate,
             "num_submissions": self.num_submissions,
-            "average_score": self.average_score,
             "max_score": self.max_score,
-            "average_code_score": self.average_code_score,
             "max_code_score": self.max_code_score,
-            "average_written_score": self.average_written_score,
             "max_written_score": self.max_written_score,
         }
 
@@ -77,12 +74,10 @@ class Notebook(Base):
         return {
             "id": self.id,
             "name": self.name,
+            "num_submissions": self.num_submissions,
             "max_score": self.max_score,
             "max_code_score": self.max_code_score,
             "max_written_score": self.max_written_score,
-            "average_score": self.average_score,
-            "average_code_score": self.average_code_score,
-            "average_written_score": self.average_written_score
         }
 
     def __repr__(self):
@@ -244,7 +239,7 @@ class SubmittedNotebook(Base):
         return {
             "id": self.id,
             "name": self.notebook.name,
-            "student": self.student.to_dict(),
+            "student": self.student.id,
             "score": self.score,
             "max_score": self.max_score,
             "code_score": self.code_score,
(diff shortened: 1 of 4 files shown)


nbgrader/api.py
```diff
@@ -2,11 +2,14 @@
...
 from sqlalchemy import (create_engine, ForeignKey, Column, String, Text, 
     DateTime, Interval, Float, Enum, UniqueConstraint)
-from sqlalchemy.orm import sessionmaker, scoped_session, relationship
+from sqlalchemy.orm import sessionmaker, scoped_session, relationship, column_property
 from sqlalchemy.orm.exc import NoResultFound, FlushError
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.ext.associationproxy import association_proxy
 from sqlalchemy.exc import IntegrityError
+from sqlalchemy.sql import and_
+from sqlalchemy.sql.expression import cast
```
Carreau: You don't seem to use cast.

nbgrader/api.py
```diff
@@ -443,6 +322,191 @@ def __repr__(self):
...
+        .correlate_except(Grade), deferred=True)
+
+Student.score = column_property(
+    select([func.coalesce(func.sum(Grade.score), 0.0)])\
+        .where(and_(
+            SubmittedAssignment.student_id == Student.id,
+            SubmittedNotebook.assignment_id == SubmittedAssignment.id,
+            Grade.notebook_id == SubmittedNotebook.id))\
+        .correlate_except(Grade), deferred=True)
+
```
Carreau: My SqlAlchemy Knowleged is ~0, but the 3 above column seem to be "just" one more `where` clause each time. 
Would there be a way to simplify ? (No, or not worth it is a good response)

nbgrader/tests/test_api.py
```diff
@@ -2,9 +2,12 @@
 
 from sqlalchemy import create_engine
 from sqlalchemy.orm import sessionmaker, scoped_session
+from nose.tools import assert_equal as _assert_equal
 
 from nbgrader import api
 
+assert_equal = lambda x, y: _assert_equal(x, y)
```
Carreau: o_O ? I don't see why you redefine it, I might be just tired.

Carreau: Oh, or just to throw on wrong number of args. Fine with me then.

nbgrader/api.py
jhamrick: Well, it slows it down a bit to have multiple sums -- so this way it first just selects all the grades that match, and then sum over them, rather than first summing over cells, then over notebooks, then over assignments.

nbgrader/tests/test_api.py
jhamrick: Oh, yeah, this is not necessary. I added that in there because my linter was complaining that `assert_equal` needed three arguments (even though it doesn't), and the visual noise was really distracting. I can remove it now that I'm done.

nbgrader/api.py
jhamrick: Good catch, thanks.

PR #151, opened 2015-03-16
Configure coverage so that it works with subprocesses

jhamrick:


jhamrick:
Gah. This is working locally, but apparently not on Travis? e.g. coverage for `assignapp` should be 100%, but Travis says it's not.

Carreau:
Isn't it just coverall that "count" differently ?

jhamrick:
If you look at the travis logs, it's apparently not being covered even before it gets to coveralls. Coverage on travis says 72% for assign, locally it says 100%.

jhamrick:
Yesssssss, it works! *Finally*. Debugging travis stuff is the worst.

For the record, there were two problems: 1) travis was not loading the `sitecustomize.py` for some reason (so I had to make it modify the sitecustomize in e.g. `/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages`), and 2) it was failing at combining the files when nbgrader was installed with `python setup.py install` rather than `pip install -e`. Probably something having to do with the paths; I don't exactly know for sure. But I am just happy that it's fixed finally!

jhamrick:
Apparently though now it's covering *everything*, hence the massive coverage decrease.

Sigh.

jhamrick:
Woohoo, it works!

Carreau:
:cake:  !! :cocktail::cocktail: ! Are you coming to BIDS ?

PR #155, opened 2015-03-18
Make launching single notebook server optional.

### LINKED ISSUE #145: Allow formgrader to run with JupyterHub auth
Add logic that allows the formgrader to run behind the config-http-proxy of JupyterHub. This will use the auth of the hub to make sure that the right set of users (the graders for that course) can use the formgrader. The lead instructor (who starts formgrader) will pick the set of graders in a config file.

jdfreder:
@jhamrick @ellisonbg 

closes #145

### DIFF
--- a/nbgrader/apps/formgradeapp.py
+++ b/nbgrader/apps/formgradeapp.py
@@ -57,10 +57,6 @@ class FormgradeApp(BaseNbGraderApp):
     directory_format = Unicode('{notebook_id}.ipynb', config=True, help="Format string for the directory structure of the autograded notebooks")
     start_nbserver = Bool(True, config=True, help="Start a single notebook server that allows submissions to be viewed.")
 
-    @property
-    def notebook_server_exists(self):
-        return self._notebook_server_exists
-
     def _classes_default(self):
         classes = super(FormgradeApp, self)._classes_default()
         classes.append(HTMLExporter)
@@ -115,9 +111,9 @@ class FormgradeApp(BaseNbGraderApp):
                     "--port", app.notebook_server_port
                 ],
                 cwd=self.base_directory)
-            self._notebook_server_exists = True
+            app.notebook_server_exists = True
         else:
-            self._notebook_server_exists = False
+            app.notebook_server_exists = False
 
         # now launch the formgrader
         app.notebook_dir = self.base_directory
--- a/nbgrader/apps/formgradeapp.py
+++ b/nbgrader/apps/formgradeapp.py
@@ -5,7 +5,7 @@ import socket
 from textwrap import dedent
 
 from IPython.config.loader import Config
-from IPython.utils.traitlets import Unicode, Integer, List, Dict
+from IPython.utils.traitlets import Unicode, Integer, List, Dict, Bool
 
 from IPython.nbconvert.exporters import HTMLExporter
 from IPython.config.application import catch_config_error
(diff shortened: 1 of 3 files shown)


nbgrader/html/formgrade.py
```diff
@@ -148,6 +148,9 @@ def view_submission(submission_id):
     if not os.path.exists(filename):
         abort(404)
 
+    if not app.notebook_server_exists:
+        abort(503, 'Submission viewing notebook server not started.')
+
```
jdfreder: @jhamrick one step further, I could modify the templates to not `a href` the `/submissions/#` url (just the formgrade.tpl?) so the grader wouldn't have to see the 503.  I can do that in this/another PR if you'd like.

jhamrick: That would be great! I think doing that in this PR is fine.

nbgrader/apps/formgradeapp.py
```diff
@@ -100,15 +105,19 @@ def start(self):
...
+            app.notebook_server_ip = self.ip
+            app.notebook_server_port = str(random_port())
+            app.notebook_server = sp.Popen(
+                [
+                    "python", os.path.join(os.path.dirname(__file__), "notebookapp.py"),
+                    "--ip", app.notebook_server_ip,
+                    "--port", app.notebook_server_port
+                ],
+                cwd=self.base_directory)
+            self._notebook_server_exists = True
```
jhamrick: I think you probably want this to be `app.notebook_server_exists`?

nbgrader/apps/formgradeapp.py
```diff
@@ -55,6 +55,11 @@ class FormgradeApp(BaseNbGraderApp):
     port = Integer(5000, config=True, help="Port for the server")
     base_directory = Unicode('.', config=True, help="Root server directory")
     directory_format = Unicode('{notebook_id}.ipynb', config=True, help="Format string for the directory structure of the autograded notebooks")
+    start_nbserver = Bool(True, config=True, help="Start a single notebook server that allows submissions to be viewed.")
+
+    @property
+    def notebook_server_exists(self):
```
jhamrick: This won't actually get called by the `app` in formgrade.py -- the "app" it's referencing there is the flask app, rather than the FormGradeApp

jdfreder:
@jhamrick comments addressed

nbgrader/html/templates/notebook_submissions.tpl
```diff
@@ -28,7 +28,9 @@
 <tbody>
   {%- for submission in submissions -%}
   <tr>
-    <td><a href="/submissions/{{ submission.id }}">Submission #{{ submission.index + 1 }}</a></td>
+    <td>{%- if notebook_server_exists -%}<a href="/submissions/{{ submission.id }}">{%- endif -%}
```
jhamrick: I don't think you want the if statement here -- otherwise this prevents people from doing any grading. The part that depends on the live notebook is in `formgrade.tpl` here: https://github.com/jupyter/nbgrader/blob/master/nbgrader/html/templates/formgrade.tpl#L65

nbgrader/html/templates/student_submissions.tpl
```diff
@@ -27,9 +27,9 @@
   {%- for submission in submissions -%}
   <tr>
     <td>
-      <a href="/submissions/{{ submission.id }}">
+      {%- if notebook_server_exists -%}<a href="/submissions/{{ submission.id }}">{%- endif -%}
```
jhamrick: Same thing here -- otherwise you can't open the student's submission for grading at all.

jdfreder: Ah I was confused.  Where I put the abort(503) is wrong.  I see what's going on now...  Another commit on it's way!

Carreau:
I've changed the coverall decrease failure test to 2%.

jdfreder:
@jhamrick sorry for the confusion, see latest commit :)

Also, I don't have nbgrader installed locally so I can't test what I'm doing.  Is there a quick way to deploy a commit for testing?  I have no docker experience- but I'm willing to learn!

jhamrick:
This looks great, thanks! I will wait until travis finishes then merge.

I have just been testing nbgrader locally, so I don't have anything set up to deploy commits for testing otherwise... is it a problem to install it locally?

PR #156, opened 2015-03-19
Use sudo=false for travis builds

jhamrick:
Hopefully will result in faster travis builds.


jhamrick:
Woohoo, changes to travis that work on the first time :-D

PR #157, opened 2015-03-19
WIP: nbgrader authentication

### LINKED ISSUE #145: Allow formgrader to run with JupyterHub auth

jdfreder:
Hey @jhamrick ,

Here's a simple extensible authentication scheme.  

FormgradeApp.authenticator_class can be used to change the authenticator.  `BaseAuth` is the base authenticator class.  A `NoAuth` class is provided as the default, and retains current behavior of `formgrade`.  I moved the `start_nbserver` traitlet and nbserver launching logic into `NoAuth`.

The `HubAuth`, empty now, will contain the logic for authenticating with JupyterHub.

Let me know what you think of this design.  I've tested it locally and everything seems to work as-is.  The not yet implement `HubAuth` can be tested quickly (when it's done) using `nbgrader formgrade --FormgradeApp.authenticator_class='nbgrader.auth.hubauth.HubAuth'`.

closes #145

### DIFF
--- a/nbgrader/apps/formgradeapp.py
+++ b/nbgrader/apps/formgradeapp.py
@@ -8,12 +8,11 @@ import time
 from textwrap import dedent
 
 from IPython.config.loader import Config
-from IPython.utils.traitlets import Unicode, Integer, List, Dict, Bool, DottedObjectName, \
+from IPython.utils.traitlets import Unicode, Integer, Dict, Type, \
     Instance
 
 from IPython.nbconvert.exporters import HTMLExporter
 from IPython.config.application import catch_config_error
-from IPython.utils.importstring import import_item
 
 from nbgrader.apps.baseapp import BaseNbGraderApp, nbgrader_aliases, nbgrader_flags
 from nbgrader.html.formgrade import app
@@ -62,7 +61,7 @@ class FormgradeApp(BaseNbGraderApp):
     base_directory = Unicode('.', config=True, help="Root server directory")
     directory_format = Unicode('{notebook_id}.ipynb', config=True, help="""Format
         string for the directory structure of the autograded notebooks""")
-    authenticator_class = DottedObjectName('nbgrader.auth.noauth.NoAuth', config=True, help="""
+    authenticator_class = Type('nbgrader.auth.noauth.NoAuth', config=True, help="""
         Authenticator used in all formgrade requests.""")
     authenticator_instance = Instance(BaseAuth, config=False)
     nbserver_port = Integer(config=True, help="Port for the notebook server")
@@ -137,8 +136,7 @@ class FormgradeApp(BaseNbGraderApp):
         super(FormgradeApp, self).start()
 
         # Init authenticator.
-        auth_class = import_item(str(self.authenticator_class))
-        self.authenticator_instance = auth_class(app, self.ip, self.port, self.base_directory, parent=self)
+        self.authenticator_instance = self.authenticator_class(app, self.ip, self.port, self.base_directory, parent=self)
         app.auth = self.authenticator_instance
 
         # first launch a notebook server
--- a/nbgrader/apps/formgradeapp.py
+++ b/nbgrader/apps/formgradeapp.py
@@ -138,7 +138,7 @@ class FormgradeApp(BaseNbGraderApp):
 
         # Init authenticator.
         auth_class = import_item(str(self.authenticator_class))
-        self.authenticator_instance = auth_class(app, self.ip, self.base_directory)
+        self.authenticator_instance = auth_class(app, self.ip, self.port, self.base_directory, parent=self)
         app.auth = self.authenticator_instance
 
         # first launch a notebook server
--- a/nbgrader/apps/formgradeapp.py
+++ b/nbgrader/apps/formgradeapp.py
@@ -1,9 +1,6 @@
 import os
-import subprocess as sp
-import socket
 import signal
 import sys
-import time
 
 from textwrap import dedent
 
@@ -17,7 +14,7 @@ from IPython.config.application import catch_config_error
 from nbgrader.apps.baseapp import BaseNbGraderApp, nbgrader_aliases, nbgrader_flags
 from nbgrader.html.formgrade import app
 from nbgrader.api import Gradebook
-from nbgrader.auth.base import BaseAuth
+from nbgrader.auth import BaseAuth, NoAuth
 
 aliases = {}
 aliases.update(nbgrader_aliases)
@@ -32,13 +29,6 @@ flags.update(nbgrader_flags)
 flags.update({
 })
 
-def random_port():
-    """get a single random port"""
-    sock = socket.socket()
-    sock.bind(('', 0))
-    port = sock.getsockname()[1]
-    sock.close()
-    return port
 
 class FormgradeApp(BaseNbGraderApp):
 
@@ -61,18 +51,14 @@ class FormgradeApp(BaseNbGraderApp):
     base_directory = Unicode('.', config=True, help="Root server directory")
     directory_format = Unicode('{notebook_id}.ipynb', config=True, help="""Format
         string for the directory structure of the autograded notebooks""")
-    authenticator_class = Type('nbgrader.auth.noauth.NoAuth', config=True, help="""
+    authenticator_class = Type(NoAuth, klass=BaseAuth, config=True, help="""
         Authenticator used in all formgrade requests.""")
     authenticator_instance = Instance(BaseAuth, config=False)
-    nbserver_port = Integer(config=True, help="Port for the notebook server")
 
     def _classes_default(self):
         classes = super(FormgradeApp, self)._classes_default()
         classes.append(HTMLExporter)
         return classes
-
-    def _nbserver_port_default(self):
-        return random_port()
 
     def init_server_root(self):
         # Specifying notebooks on the command-line overrides (rather than adds)
@@ -104,20 +90,7 @@ class FormgradeApp(BaseNbGraderApp):
 
     def _signal_stop(self, sig, frame):
         self.log.critical("received signal %s, stopping", sig)
-
-        if app.notebook_server_exists:
-            app.notebook_server.send_signal(sig)
-            for i in range(10):
-                retcode = app.notebook_server.poll()
-                if retcode is not None:
-                    app.notebook_server_exists = False
-                    break
-                time.sleep(0.1)
-
-            if retcode is None:
-                self.log.critical("couldn't shutdown notebook server, force killing it")
-                app.notebook_server.kill()
-
+        self.authenticator_instance.stop(sig)
         sys.exit(-sig)
 
     def build_extra_config(self):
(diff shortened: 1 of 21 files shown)


nbgrader/html/formgrade.py
```diff
@@ -169,20 +205,19 @@ def view_submission(submission_id):
...
     }
 
-    if app.notebook_server_exists:
-        resources['notebook_path'] = "http://{}:{}/notebooks/{}".format(
-            app.notebook_server_ip,
-            app.notebook_server_port,
-            os.path.relpath(filename, app.notebook_dir))
+    if server_exists:
+        relative_path = os.path.relpath(filename, app.notebook_dir)
+        resources['notebook_path'] = app.auth.get_notebool_url(relative_path)
```
jhamrick: I think this should be `get_notebook_url`

requirements.txt
```diff
@@ -1,4 +1,3 @@
-ipython==3.1.0.dev0
```
jhamrick: Is there a reason for removing this (rather than I guess changing the version number)?

jhamrick:
This looks awesome -- thanks for working on this! I think Travis is unhappy because you need to add some stuff to the `__init__.py` for auth, and I left a few small comments inline. Overall though I think this design works really well!

requirements.txt
jdfreder: No that's a Jon install hack, I accidentally checked it in.

jdfreder:
@jhamrick @minrk @ellisonbg this now works with JupyterHub auth.  I had to convert all of the URLs in the templates to relative paths, since the JHub auth nests the entire blueprint (we are using a blueprint now instead of registering routes directly on the flask app) in a sub path `hub/formgrade`.  @jhamrick I still haven't verified that all of the URLs map correctly, but most of them do.  Any path that's more than one level deep probably won't map correctly since I just preprended `..`  to the URLs.

To test, without writing any config files, run the following:
```bash
export CONFIGPROXY_AUTH_TOKEN=alice; jupyterhub
```

and in another terminal:

```bash
export JPY_API_TOKEN=`jupyterhub token`; export CONFIGPROXY_AUTH_TOKEN=alice; nbgrader formgrade --FormgradeApp.authenticator_class='nbgrader.auth.hubauth.HubAuth' --HubAuth.graders='["jon"]'
```

I want to do some cleanup, by using config to pass in the API token.

jdfreder:
Hehe, I'm not sure how coverage increased.  I didn't write a single test!

jhamrick:
Awesome, I will test this out later tonight or tomorrow. I am also working on some basic tests for the formgrader that just check that all the links work properly, so that should be perfect for verifying that everything still works -- I'll finish up with that soon and merge it so then we can test this PR against it.

nbgrader/apps/formgradeapp.py
```diff
@@ -54,8 +47,11 @@ class FormgradeApp(BaseNbGraderApp):
     ip = Unicode("localhost", config=True, help="IP address for the server")
     port = Integer(5000, config=True, help="Port for the server")
     base_directory = Unicode('.', config=True, help="Root server directory")
-    directory_format = Unicode('{notebook_id}.ipynb', config=True, help="Format string for the directory structure of the autograded notebooks")
-    start_nbserver = Bool(True, config=True, help="Start a single notebook server that allows submissions to be viewed.")
+    directory_format = Unicode('{notebook_id}.ipynb', config=True, help="""Format
+        string for the directory structure of the autograded notebooks""")
+    authenticator_class = DottedObjectName('nbgrader.auth.noauth.NoAuth', config=True, help="""
```
minrk: you can use `authenticator_class = Type('nbgrader.auth...')` and avoid needing the DottedObjectName + import_string.

nbgrader/auth/hubauth.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+class HubAuth(BaseAuth):
+    """Jupyter hub authenticator."""
+
+    graders = List([], config=True, help="List of JupyterHub user names allowed to grade.")
+
+    proxy_address = Unicode(None, config=True, allow_none=True, help="Address of the configurable-http-proxy server.")
+    proxy_port = Int(8001, config=True, help="Port of the configurable-http-proxy server.")
+    hub_address = Unicode(None, config=True, allow_none=True, help="Address of the hub server.")
```
minrk: These addresses don't need to allow None, they can have 

```python
def _hub_address_default(self):
    return self._ip
```

No need for the init logic below.

nbgrader/auth/hubauth.py
```diff
@@ -0,0 +1,123 @@
...
+        """Does the notebook server exist?"""
+        return True
+
+    def get_notebook_url(self, relative_path):
+        """Gets the notebook's url."""
+        # TODO
+        return "http://{}:{}/notebooks/{}".format(
+            self.hub_address,
+            self.hub_port,
+            relative_path)
```
minrk: This isn't the right URL, is it? Shouldn't it be the public (proxied) URL of the user's server?

nbgrader/html/templates/notebook_submissions.tpl
```diff
@@ -1,8 +1,8 @@
 {%- extends 'gradebook.tpl' -%}
 
 {%- block breadcrumb -%}
-<li><a href="/assignments">Assignments</a></li>
-<li><a href="/assignments/{{ assignment_id }}">{{ assignment_id }}</a></li>
+<li><a href="../assignments">Assignments</a></li>
```
minrk: Is this jinja? If so, a `{{base_url}}` might be more robust than `..` everywhere.

jhamrick:
@jdfreder #164 are the tests that I've been working on -- they're just about ready to merge, except that they keep hanging on travis after completing. I'll try to figure out why tomorrow and let you know once I've actually merged them. It's possible there might be a merge conflict with this (I had to add just a few things to `formgradeapp.py`) but probably nothing too major.

jhamrick:
Ok, #164 has been merged now!

nbgrader/auth/hubauth.py
jdfreder: Ah yes, you're right.  It should be /user/notebooks.  I haven't tested this with notebooks yet, but will test today.

jdfreder:
Thanks for the review @minrk and thanks for letting me know @jhamrick .  I'll rebase today and try testing the actual complete grading process.

nbgrader/html/templates/notebook_submissions.tpl
jdfreder: Yeah, it is, and yes that's a good suggestion.

nbgrader/apps/formgradeapp.py
jdfreder: Nifty!

jhamrick:
I think if you add `nbgrader.auth` to the list of packages in `setup.py` that will at least fix your errors with the docs.

jdfreder:
Thanks

jdfreder:
@jhamrick tests are passing now.  Commit d62ec2c
relys on the new JHub cookie auth API https://github.com/jupyter/jupyterhub/pull/207 which hasn't been merged yet.  Until it's merged, I suggest testing by checking out f122d7d .

jhamrick:
Awesome, I will test this out later today after the meeting!

jhamrick:
This is looking really awesome. Some comments as I'm going through:

1. Currently, unless the notebooks and formgrader are all rooted at $HOME, opening up the live notebook won't work. Maybe it would be worth it to add a config option to the `HubAuth` class that is a prefix for the path to the notebooks that the formgrader is trying to access? Then `get_notebook_url` would add that prefix to the beginning of `relative_path`.

2. It seems like the css/js/fonts resources aren't being loaded when using `HubAuth` (seems to be fine for `NoAuth`, though). Looks like they are trying to be loaded from e.g. `/hub/static/css/bootstrap.min.css`, but probably this should be `/hub/formgrade/static/css/bootstrap.min.css`?

3. Similarly, the breadcrumbs at the top of the formgrader page (the one rendered from `formgrade.tpl`) seem to have invalid links -- e.g. `/assignments` rather than `/hub/formgrade/assignments`.

I wonder how difficult it would be to add another set of tests that also launch JupyterHub and then verify that all the links are correct there, too... I feel like probably it would not be too difficult. If you want I can try to look into this today.

jdfreder:
> Currently, unless the notebooks and formgrader are all rooted at $HOME, opening up the live notebook won't work. Maybe it would be worth it to add a config option to the HubAuth class that is a prefix for the path to the notebooks that the formgrader is trying to access? Then get_notebook_url would add that prefix to the beginning of relative_path.

Good idea!  Yeah, I've been running both servers from ~/ so I'm not surprised a bug like that exists.



> It seems like the css/js/fonts resources aren't being loaded when using HubAuth (seems to be fine for NoAuth, though). Looks like they are trying to be loaded from e.g. /hub/static/css/bootstrap.min.css, but probably this should be /hub/formgrade/static/css/bootstrap.min.css?

Hmmm, strange.  I think the `{{base_url}}` template resource that @minrk suggested fixed it for me.  Again, there must be assumptions leaking into the implementation about my install&execution (paths)- I'll investigate.

> seem to have invalid links

Sounds like {{base_url}} is missing, I'll look.

Thanks again for testing this, it helps me a lot.  The regression tests are probably a good idea, and worth the price of admission- having Travis install JHub and friends (node, npm, etc.) and if you're willing to look into it, that'd be great!  In the meantime I'll resolve the multi-path problem.

jhamrick:
Ok @jdfreder I made a pull request against your branch for the tests!

jdfreder:
Ok @jhamrick , I think I addressed all of the URL problems.  The biggest problem was a missing `resources.` prefix on `base_url` for the formgrade.tpl (since it's going through nbconvert).  I also added a `notebook_url_prefix` configurable that allows you to set the relative path of formgrader to JHub.  i.e. if JHub is started in `~/` and formgrader in `~/Documents/tests`, `notebook_url_prefix="Documents/tests"`.

I merged your tests, thanks!  Let's see if all is good and sound..

jhamrick:
Awesome, I am also just about finished writing some tests that will check that the url prefix works, so I'll run those against your changes and submit another PR against your branch soon.

jhamrick:
Oops, I broke the tests on python 2. This diff should fix it, I think:

```
diff --git a/nbgrader/tests/test_nbgrader_formgrade.py b/nbgrader/tests/test_nbgrader_for
index ad98948..1489164 100644
--- a/nbgrader/tests/test_nbgrader_formgrade.py
+++ b/nbgrader/tests/test_nbgrader_formgrade.py
@@ -10,7 +10,8 @@ from nose.tools import assert_equal
 from textwrap import dedent

 try:
-    from urllib import urljoin, unquote # Python 2
+    from urllib import unquote # Python 2
+    from urlparse import urljoin
 except ImportError:
     from urllib.parse import urljoin, unquote # Python 3
```

jhamrick:
And as for the python 3 tests, I think that line that just does `npm install` in travis is the problem. I copied that from jupyterhub, but probably shouldn't have.

jhamrick:
I am working on fixing travis stuff on the `auth` branch, I'll make a PR to your branch when I'm done!

jdfreder:
Woohoo!  Looks like your new tests pass @jhamrick .

jhamrick:
Yay! Thanks for all the work you've done on this @jdfreder , nice job.

PR #160, opened 2015-03-20
Use Selenium for javascript/html tests

jhamrick:


nbgrader/tests/test_nbextension.py
```diff
@@ -0,0 +1,200 @@
...
+    def teardown_class(cls):
+        cls.nbserver.kill()
+
+        os.chdir(cls.origdir)
+        shutil.rmtree(cls.tempdir)
+        shutil.rmtree(cls.ipythondir)
+
+    def setup(self):
+        shutil.copy(os.path.join(os.path.dirname(__file__), "files", "blank.ipynb"), "blank.ipynb")
+        self.browser = webdriver.PhantomJS()
```
jdfreder: @takluyver it looks like @jhamrick 's implementation uses Phantom, which from what I understand is part of our problem with the current slimer system (CC @minrk).  @jhamrick have you tried the other drivers they have available?  http://docs.seleniumhq.org/projects/webdriver/  If so, were there any caveats?

jhamrick: I didn't try any of the other drivers as I didn't know whether they would work on travis, though @Carreau tells me that you can use saucelabs with travis to test the other drivers. I did initially try the firefox driver locally at the very beginning, and that seemed to work fine out of the box, so I would expect it to be pretty easy to switch drivers.

takluyver: There are two distinct problems we have:

- The old webkit used in Phantom doesn't support everything we want (current websockets protocol, PATCH requests, binary websockets). Being able to use alternative drivers should alleviate this.
- The general async-js awfulness causing things like sporadic failures, unclear error messages when failures do occur, and so on. We're never going to completely escape this, but I'm optimistic that it might be better with Selenium than with Casper, because the difference in language between the tests and the code running requires a cleaner abstraction layer. Also, Selenium seems rather more mature.

PR #164, opened 2015-03-20
Basic formgrader tests

### LINKED ISSUE #119: Figure out how to write tests for the formgrader

### LINKED ISSUE #163: Formgrader doesn't properly shut down notebook app

jhamrick:
This mostly just checks if all the links work -- more tests will still need to be written in the future to make sure that everything is being displayed properly, too, and that you can do things like update the grades in the database.

Also, Selenium is pretty nice!

* Fixes #163
* Fixes #119


jhamrick:
Bleh. I don't know why the tests keep hanging on travis; they finish fine on my machine :angry: It's almost like the formgrader is never actually getting the `SIGTERM` because there are no logs saying that it has.

jdfreder:
This is a great data point for using Selenium in master-  @takluyver .

takluyver:
Nice! Thanks Jess for trying out Selenium, we might look into switching to it for the notebook tests.

jhamrick:
Sure, it was nice playing around with a different framework. For reference, you may also want to look at #160, which actually tests the nbextension and so is a little closer to the stuff the notebook tests do.

takluyver:
Cool, thanks.

PR #168, opened 2015-03-21
[Work in progress] Command install activate nbgrader extension.

Carreau:
Rip off from jupyter-drive installation for now.
Need some modifications.


jhamrick:
Awesome! Though maybe it would be more consistent for this to go in `nbgrader.apps.nbgraderapp.py` as an additional subcommand?

Carreau:
> Awesome! Though maybe it would be more consistent for this to go in nbgrader.apps.nbgraderapp.py as an additional subcommand?

I was feeling that this was more a once in a blue moon thing, that you probably don't want to run everyday, and that only admins run, so pretty hidden. 

Up to you.

Carreau:
`python -m nbgrader --activate default` works now. Symlink by default :-)

Carreau:
I think I'll give up on installing the nbextension in the temporary folder created in test_nbextension...

jhamrick:
I think that's fine. Probably the test doesn't need to actually create a temporary directory -- I'm not entirely sure why I had it do that, so I think it would be fine if you just removed that functionality if you wanted to.

nbgrader/tests/test_nbextension.py
```diff
@@ -1,7 +1,10 @@
 import subprocess as sp
 import tempfile
 import os
+import os.path
```
jhamrick: I think this import is redundant

Carreau: Yep, sys also I guess. I need to clean a bit...

nbgrader/tests/test_nbextension.py
```diff
@@ -12,28 +15,46 @@
...
+from nbgrader.install import install
+
+
 class TestCreateAssignmentNbExtension(object):
 
     @classmethod
     def setup_class(cls):
         cls.tempdir = tempfile.mkdtemp()
         cls.ipythondir = tempfile.mkdtemp()
+        cls.ipythondir = os.path.expanduser('~/eraseme/temp')
```
jhamrick: There are now two `ipythondir`s here -- which one is the one you want to keep? Does it work now if you use the tempdir created from `tempfile`?

nbgrader/tests/test_nbextension.py
```diff
@@ -12,28 +15,46 @@
...
+            )
 
     @classmethod
     def teardown_class(cls):
         cls.nbserver.kill()
 
         os.chdir(cls.origdir)
-        shutil.rmtree(cls.tempdir)
-        shutil.rmtree(cls.ipythondir)
+        #shutil.rmtree(cls.tempdir)
```
jhamrick: Should these two lines be uncommented?

setup.py
```diff
@@ -32,9 +32,9 @@
         'nbgrader.exporters'
     ],
     package_data={
-        '': [
-            'nbextensions/*.js',
-            'nbextensions/*.css'
+        'nbgrader': [
```
jhamrick: Is it necessary to move these into an additional subfolder? Is it just to keep the `nbextensions` folder in the profile less cluttered? If so, then is it possible to just create the folder at install time, rather than having an extra directory in the nbgrader repository itself?

Carreau: It's on my todo list. I didn't acchieve to access it from with the python module while it was outside of `/nbgrader`. But now the it works, I can tackle that again.

nbgrader/install.py
```diff
@@ -0,0 +1,132 @@
...
+    try:
+        with io.open(json_file, 'r') as f:
+            config = json.loads(f.read())
+    except IOError:
+        # file just don't exists yet. IPython might have never been launched.
+        config = {}
+
+    if not config.get('load_extensions', None):
+        config['load_extensions'] = {}
+    config['load_extensions']['nbgrader/nbgrader'] = True
```
jhamrick: It seems weird to me to have to activate it with `nbgrader/nbgrader` rather than just `nbgrader`. Perhaps something clearer would be to rename the extension itself to something like `create_assignment` (as technically that is what it's called), so then when you load it you would be loading `nbgrader/create_assignment`.

setup.py
jhamrick: Ok, awesome!

nbgrader/install.py
Carreau: We could do purely `nbgrader` but then if ever you want multiple file, it becomes annoying. Usually we can do a `main` file so you activate `nbgrader/main` as you prefer.

nbgrader/install.py
```diff
@@ -0,0 +1,132 @@
...
+
+    if not os.path.exists(json_dir):
+        os.mkdir(json_dir)
+
+    print('...', json_file)
```
Carreau: OOps, extra print.

nbgrader/tests/test_nbextension.py
```diff
@@ -12,28 +15,46 @@
...
+        cls.ipythondir = os.path.expanduser('~/eraseme/temp')
         cls.origdir = os.getcwd()
         os.chdir(cls.tempdir)
 
+        # ensure IPython dir exists.
+        sp.call([
+            'ipython', 'profile', 'create', '--ipython-dir', cls.ipythondir])
+
+
+        install(path=os.path.join(cls.ipythondir), symlink=False, activate=True)
```
Carreau: join nothing.

Carreau:
`cov <installscript>` was not working, I removed it in favor a actually calling `main(<arglist>)` in install nbextension tests.

nbgrader/install.py
jhamrick: Sure, that sounds reasonable.

Carreau:
o_O that's a bit scary how things wrk or not... if I do the following with data_files:
 
```
/
/nbgrader/
/nbextensions/
```

And move things around at install time, it should work on user machine. 

Though for development, If you `python setup.py install`, PY3 is smart enough to "link" to the cloned repo and so the layout of file it sees is not the same that to one that would make the wheels or be installed...

So you cannot run things fro inside dev dir and/or you have to special case in code ...

3rd possibility, is I do smth really wrong.

I think I'll revert to move the nbextension data files in package tomorrow.

jhamrick:
Ok, I think this is working the way it's supposed to -- I guess the only other thing would be to implement the deactivation part? I can do that in a bit, too.

jhamrick:
Ok @Carreau you should take a look over this -- is there anything else you wanted to do with it? If not, then let me know, and I'll merge it!

Carreau:
Ahhhhh the comma after install.... why did I missed it ! Will have a look.

nbgrader/install.py
```diff
@@ -0,0 +1,176 @@
...
+    except IOError:
+        # file doesn't exist yet. IPython might have never been launched.
+        return
+
+    if 'load_extensions' not in config:
+        return
+    if 'nbgrader/create_assignment' not in config['load_extensions']:
+        return
+
+    config['load_extensions']['nbgrader/create_assignment'] = False
```
Carreau: I don't think this will deactivate the ext. Actually the JS config use a dict as a set, but set do not exist in json, the value dont' have effet yet. You we might have to delete the key. 
Let me check.

jhamrick: Ahh, ok, good catch.

jhamrick: If that's the case, then probably the tests should be updated too to make sure the key doesn't exist rather than checking for `False`.

Carreau: Will do.

Carreau:
Yeahhhh works !

@jhamrick I'll let you push Little Greeny

jhamrick:
Woohoo! :tada: :cake: Thanks @Carreau !

Carreau:
Hey ! Go take care of your relatives !

jhamrick:
Doing that now :)

PR #176, opened 2015-03-26
Actually prune if empty dict, not other way around.

Carreau:
Or you would just deactivate all extension, which much likely not what
you want to do :-)

---- 

I think that make sens. (haven't tried though) 

Can comme up with a test also.

### DIFF
--- a/nbgrader/install.py
+++ b/nbgrader/install.py
@@ -101,7 +101,7 @@ def deactivate(profile=None, ipython_dir=None):
     del config['load_extensions']['nbgrader/create_assignment']
 
     # prune if last extension.
-    if config['load_extensions']:
+    if not config['load_extensions']:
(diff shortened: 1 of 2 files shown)


Carreau:
Ok, the first commit should fail, and the second pass.

Carreau:
Travis is happy but coveralls refuse to post the new version :(

jhamrick:
Oof, good catch!

Carreau:
> Oof, good catch!

Yeah, I was doing completely something else totally unrelated and far from my computer when my brain  told me "Hey ! Hey! There is no `not`, you should put one". One of these moment where your wondering how you think of that.

jhamrick:
Hooray for cognitive processes that are still doing stuff in the background even when the rest of our brains are not :smile:

PR #178, opened 2015-03-27
Make the hub_base_url configurable.

jdfreder:
@jhamrick 

This is required for complex deployments.

I have an SSL terminating nginx sitting in front of the Hub and nbgrader.  I need to direct clients to nginx's HTTPS domain.


jhamrick:
Thanks!

jonathanmorgan:
I believe the problem we are having with nbgrader and jupyterhub SSL is related to this same file - looks like "http" is hard-coded in the base URL.  Would be good to be able to configure it to use "https" instead, if needed.

jhamrick:
Aha, yes, that is probably the case. Can you open a new issue for that and I'll submit a fix?

jonathanmorgan:
certainly.  Thanks.

PR #188, opened 2015-04-01
make profile ags mandatory + docs

Carreau:
make `profile` first positional arg, + add docs on installing.

It occurred to me that install might not actually require a profile name as it can install globally on the system.

### DIFF
--- a/docs/user_guide/1 - Installing.ipynb
+++ b/docs/user_guide/1 - Installing.ipynb
@@ -30,7 +30,14 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "To additionally install the assignment toolbar notebook extension:"
+    "To additionally install, activate or deactivate the assignment toolbar notebook extension:"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "To get help and see all the options you can pass while installing/activating nbgrader, use:"
    ]
   },
   {
@@ -38,8 +45,23 @@
    "metadata": {},
    "source": [
     "```bash\n",
-    "cp nbextensions/* $(ipython locate)/nbextensions/\n",
+    "python -m nbgrader --help\n",
     "```"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "You can install and activate nbgrader assignment toolbar extension with:\n",
+    "\n",
+    "```bash\n",
+    "python -m nbgrader --install --activate <profilename>\n",
+    "```\n",
+    "\n",
+    "Where `<profilename>` is the name of the profile you want to use for creating assignments. \n",
+    "By default the extension will by symlink, so you shouldn't need to reinstalling it after upgrading nbgrader. \n",
+    "use `--no-symlink` to actually copy the file."
    ]
   }
  ],
--- a/docs/user_guide/1 - Installing.ipynb
+++ b/docs/user_guide/1 - Installing.ipynb
@@ -72,15 +72,6 @@
     "%%bash\n",
     "python -m nbgrader --help"
    ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {
-    "collapsed": true
-   },
-   "outputs": [],
-   "source": []
   }
  ],
  "metadata": {},
--- a/docs/user_guide/1 - Installing.ipynb
+++ b/docs/user_guide/1 - Installing.ipynb
@@ -30,7 +30,28 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "To additionally install, activate or deactivate the assignment toolbar notebook extension:"
+    "You can install and activate nbgrader assignment toolbar extension with:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "%%bash\n",
+    "python -m nbgrader --install --activate default"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Where `default` is the name of the profile you want to use for creating assignments. \n",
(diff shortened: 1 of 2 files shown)


Carreau:
Got a weird error telling me output not cleared on only-markdown-cells notebook...

jhamrick:
Can you also make the default be `default` in the command line parsing stuff (line 122)?

docs/user_guide/1 - Installing.ipynb
```diff
@@ -30,7 +30,14 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "To additionally install the assignment toolbar notebook extension:"
+    "To additionally install, activate or deactivate the assignment toolbar notebook extension:"
```
jhamrick: I think this cell is redundant

docs/user_guide/1 - Installing.ipynb
```diff
@@ -30,7 +30,14 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "To get help and see all the options you can pass while installing/activating nbgrader, use:"
```
jhamrick: I would actually put the `python -m nbgrader --install --activate <profilename>` example first, then the `--help` example, since more people are going to just want to know how to install it and don't care about the other options.

docs/user_guide/1 - Installing.ipynb
```diff
@@ -38,9 +45,24 @@
...
+   "source": [
+    "You can install and activate nbgrader assignment toolbar extension with:\n",
+    "\n",
+    "```bash\n",
+    "python -m nbgrader --install --activate <profilename>\n",
+    "```\n",
+    "\n",
+    "Where `<profilename>` is the name of the profile you want to use for creating assignments. \n",
+    "By default the extension will by symlink, so you shouldn't need to reinstalling it after upgrading nbgrader. \n",
+    "use `--no-symlink` to actually copy the file."
```
jhamrick: 'use' should be capitalized

docs/user_guide/1 - Installing.ipynb
```diff
@@ -38,9 +45,24 @@
    "metadata": {},
    "source": [
     "```bash\n",
-    "cp nbextensions/* $(ipython locate)/nbextensions/\n",
+    "python -m nbgrader --help\n",
```
jhamrick: It might be nice to actually make this a code cell, so the output gets displayed in the docs. You can use the `run` command from `nbgrader.utils` so that the doc checker machinery will catch any errors if it fails to run.

docs/user_guide/1 - Installing.ipynb
```diff
@@ -38,9 +45,24 @@
...
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "You can install and activate nbgrader assignment toolbar extension with:\n",
+    "\n",
+    "```bash\n",
+    "python -m nbgrader --install --activate <profilename>\n",
```
jhamrick: Same thing here about making this a code cell.

docs/user_guide/1 - Installing.ipynb
Carreau: OK

docs/user_guide/1 - Installing.ipynb
Carreau: Ok

docs/user_guide/1 - Installing.ipynb
Carreau: Done

docs/user_guide/1 - Installing.ipynb
Carreau: Done.

docs/user_guide/1 - Installing.ipynb
Carreau: Done

Carreau:
One annoying things about making `default` the default is that `python -m nbgrader` succed with no error (well error message that `IPython.config` is deprecated here... but irrelevant). Making no default have it print help.

jhamrick:
You could add a check that they've specified at least one of `--activate`, `--deactivate`, or `--install`, and print the help and exit if they're missing any of them.

docs/user_guide/1 - Installing.ipynb
```diff
@@ -30,17 +30,57 @@
...
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "%%bash\n",
+    "python -m nbgrader --help"
+   ]
+  },
+  {
+   "cell_type": "code",
```
jhamrick: Extra cell here.

Carreau:
> You could add a check that they've specified at least one of --activate, --deactivate, or --install, and print the help and exit if they're missing any of them.

done, and extra cell removed.

jhamrick:
Thanks @Carreau !

nicola:
This is so magic, I think you should keep the standard way of installing extensions

jhamrick:
What do you mean by the standard way of installing extensions? There isn't really one, yet -- every extension has its own way of installing stuff.

nicola:
I am new to jupyter, and what I have seen ofter has been `ipython install-nbextension url`

jhamrick:
Ah, I see. Yes, that would be a somewhat more standard way of installing it, though we are actually using that machinery underneath (this is actually an outdated PR, see #207 and #225). There's a few additional arguments for having our own version of it:

1.The way we're doing it in nbgrader additionally makes it possible to *activate* the extension (which `ipython install-nbextension` does not), which is something I really want to have. There are plans to move this functionality into ipython as well (see ipython/ipython#8262).
2. I do think there is something to be said though for just doing `python -m nbgrader --install` rather than `ipython install-nbextension https://raw.githubusercontent.com/jupyter/nbgrader/master/nbgrader/nbextensions/nbgrader`
3. Once the first release of nbgrader happens, we won't want people installing nbextension from master, and the path to the files will differ from system to system, so it will be better for someone to do `python -m nbgrader --install --activate` after they've installed nbgrader. I suppose they could install from a tag on github, but it seems better that after you've installed nbgrader, you can install the nbextension without needing to download anything else.

PR #248, opened 2015-04-16
Fix timeouts when testing nbextension/formgrader

jhamrick:
Continuing to debug test failures...


jhamrick:
Yaaaaay no more timeouts! At least hopefully. The last several builds have all timed out at the same place, except for the very last one, where I think I've (finally) managed to fix it.

PR #263, opened 2015-05-25
Fix hubauth to work with current jupyterhub

jhamrick:
Recent changes to JupyterHub have renamed the `user` key to `name` in the response json. This updates the JupyterHub authentication to use `name` instead of `user`.

Ping @ellisonbg and @jdfreder , just FYI. I need to make this change so the tests on travis will pass, but wanted you to be aware of it.


jdfreder:
Thanks for the ping WRT this!

PR #300, opened 2015-06-23
Validation of cell ids

jhamrick:
As per a discussion with Brian, this changes the notebook extension so that cell ids are only validated on save, rather than as soon as they are modified. It also adds in better validation in the `CheckCellMetadata` preprocessor.


jhamrick:
Also, cc @ellisonbg

PR #319, opened 2015-07-11
Fixing installation requirements and documentation.

lphk92:
Because conda does not preserve package information using ipython[notebook],
developers will need to install ipython notebook separately from other
requirements


jhamrick:
I think `pip install ipython[all]` needs to be added to the `before_install` task in `tasks.py` for the install to complete correctly on travis.

jhamrick:
Thanks!

PR #324, opened 2015-07-11
fixes warning about points

### LINKED ISSUE #314: Warning about different point values
nbgrader will warn if the points are different, which is ok in most cases except when it's just an instance of the points being a string in the notebook and a float in the database, in which case this should be ok and ignored (or, even better, nbgrader assign should make sure all the points get converted to floats).

ellachao:
Fixes #314


jhamrick:
The failing test is due to something unrelated, so I'm going to go ahead and merge. Thanks!

PR #323, opened 2015-07-11
try to get rid of filenames

Carreau:
Working on #321, Travis, please test.

PR #329, opened 2015-07-12
Package with Flit

Carreau:
On top of https://github.com/jupyter/nbgrader/pull/323

Use `flit wheel`, `flit wheel --upload` once ready. A `--debug` flag might be useful to see what things it is doing (sending PR on flit to have even more info).


nbgrader/apps/extensionapp.py
```diff
@@ -16,8 +16,16 @@
...
+print(install_aliases)
+try:
+    del install_aliases['destination']
+except KeyError:
+    pass
+try:
+    del install_aliases['ipython-dir']
+except KeyError:
+    pass
+
```
Carreau: These raise keyerrors on recent enough IPython.

Carreau:
it works !

LICENSE
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
```
jhamrick: I think there is already a license in `COPYING.md`.

flit.ini
```diff
@@ -0,0 +1,16 @@
+[metadata]
+module = nbgrader
+author = Jessica Hamrick
```
jhamrick: Thanks for putting me as the author :-) But I think maybe it would be better to do "Jupyter Development Team" since that's what all the other Jupyter projects (like JupyterHub) have.

jhamrick:
Thanks Matthias! One question I have is how should I recommend to people to install the dev requirements? Currently, they would pip install with them, but with flit that won't be possible.

Also the installation instructions in README.md will need to be updated, but I can do that myself when it's actually released.

minrk:
A local `flit install` should install dependencies now that takluyver/flit#42 is merged.

LICENSE
Carreau: Oh, missed that. removed.

Carreau:
Yes, you can also `flit install --symlink` for dev install.

takluyver:
As it stands, dev install would be two steps:

```
pip install -r dev-requirements.txt
flit install --symlink
```

jhamrick:
@Carreau can you rebase? Currently testing this pr doesn't work because the symlinks are still there.

Carreau:
Done.

jhamrick:
:+1: looks good, I will merge just before I release!

jhamrick:
Sigh, I realized another issue this is going to cause. I need to have a `setup.py` in order for Read The Docs to be able to build... so I'm not sure what to do about that. I'm tempted to just say screw it and use standard setuptools, but maybe there's a better option I don't know about? Maybe I could do something hacky that uses a `setup.py` which isn't a real `setup.py` but just calls `flit install`?

Thoughts @takluyver @Carreau ?

takluyver:
You can make it importable without installing it in the `conf.py`, like this:

https://github.com/takluyver/astsearch/commit/5eaecccf331efb2f3d14317e22ed729ad3f3b621

If it needs dependencies to import the things you want, though, they have to get installed. I often have a requirements.txt inside the docs folder, and tell RTD to use that:

https://github.com/takluyver/astsearch/blob/master/doc/requirements.txt

Carreau:
Flit docs are on rtfd, so @takluyver find a way to do it. 
Though I don’t remember how.

I’ll try to dig in.

jhamrick:
Ok thanks, I'll have a look at those examples!

takluyver:
RTD are apparently [working on](https://github.com/rtfd/readthedocs.org/issues/1336) an in-repo config file as an alternative to the web configuration interface. I hope this might also lead to more generalised options for setting stuff up, so that we can do `pip install flit; flit install`, rather than just the option to use setup.py or not.

jhamrick:
Ok, so it's not enough for me to just make it importable -- I really do need it to be installed because my setup has RTD executing some notebooks that call the `nbgrader` command. Hopefully the in-repo config thing will happen soon, but in the meantime I'm including a fake `setup.py` that just calls `flit install`, which should hopefully work.

Carreau:
that's one more datapoint for thomas to have a flit command to create a fake `setup.py` that actually call flit :-)

takluyver:
Myeh, that sucks. Sorry about that.

PR #334, opened 2015-07-12
Auto-resize textboxes when giving feedback in formgrader

alope107:
Auto-resizes textboxes when giving feedback. Sets max size at 150px, but this can be tweaked in CSS to taste.


nbgrader/html/templates/formgrade.tpl
```diff
@@ -64,6 +65,9 @@ window.MathJax = {
 {%- endblock body %}
 
 {% block footer %}
+<script>
+    autosize(document.querySelectorAll('textarea'));
```
jhamrick: I think a better place to actually do the autosize would be in the `initialize` function for the `CommentUI` (which is in `models.js`).

alope107:
Moved js, ready for re-review.

nbgrader/html/static/js/models.js
```diff
@@ -121,6 +121,9 @@ var CommentUI = Backbone.View.extend({
         this.listenTo(this.model, "sync", this.animateSaved);
 
         this.$comment.attr("placeholder", this.model.get("auto_comment") || "Comments");
+
+        autosize(document.querySelectorAll('textarea'));
```
jhamrick: I think you can use `this.$comment` to use just the specific textarea rather than querying all of them. This `initialize` function will get called once for each comment, so if you select them all here it'll end up doing n^2 resizes :-)

alope107:
Ah, good catch. Changed and ready for review.

alope107:
Fixed .gitignore so dist js is pushed.

jhamrick:
Hmm, travis seems to be failing though I'm not really sure why. It seems like the autosize javascript isn't loaded? It seemed to load fine though when I tested it locally. It might be that you need to add something to the tests to make sure that it waits for everything to load...

jhamrick:
@alope107 do you think you'll be able to try to fix the travis tests soon? I'd like to cut the 0.1.0 release tonight. If not, that's fine -- I can push this to the next release.

alope107:
I think this one is best pushed to the next release. There are some weird issues with my environment that I need to work through before I can get the tests to run properly.

jhamrick:
Ok, sounds good. Thanks for working on it!

jhamrick:
@alope107 If it's alright with you, I am going to go ahead and fix this last failing test, as I'd like to include this feature in the next release (which I am planning to make in the next few days).

alope107:
Yes, please go ahead. Sorry I didn't get around to this earlier.

jhamrick:
PR recreated as #364

PR #337, opened 2015-07-12
some windows work

svurens:


svurens:
I guess a quick rundown of all the changes: 

run_command ends up starting bash from command line to run the nbgrader command. It's likely that you want to create an nbgrader.exe and then add it to the path.

The .replaces are to fix os.path.join. Windows (or at least mine) can understand forward slashes in the path. Backslashes are an issue with regexes and command line arguments, since the backslash escapes out. Some of the .replaces are probably unnecessary, but it might be good to just use forward slashes for everything.

The os.chmod is to allow deleting of files that are read only. The default permissions are either 666 or 644

jhamrick:
Closing in favor of #413. Thanks for all the work you did on this @svurens , it was really helpful! I spent way less time debugging stuff than I would have had to otherwise since you had already discovered most of the major issues, like stuff dealing with backslashes in paths, permissions issues, trying to remove files while they were still open, etc. Just for reference, the only other big thing I ran into had to do with signaling subprocesses, and I learned that for windows you want to handle the CTRL_BREAK_EVENT rather than SIGINT.

PR #356, opened 2015-08-12
Add cache to nbgrader submit and nbgrader list

jhamrick:
This modifies the behavior of `nbgrader submit` so that it additionally copies a version of the assignment to a cache directory (by default, `~/.nbgrader/cache`). Then, the `--cached` flag is added to `nbgrader list` so that you can see the list of cached assignments.

This is kind of a workaround to #295. It's not a perfect or ideal solution, but it'll work for the time being until we can implement something more robust like an authenticated REST api for releasing and submitting assignments.

FYI @ellisonbg

### DIFF
--- a/nbgrader/apps/baseapp.py
+++ b/nbgrader/apps/baseapp.py
@@ -229,7 +229,21 @@ class TransferApp(BaseNbGraderApp):
         help="Format string for timestamps"
     )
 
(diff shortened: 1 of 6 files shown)


ellisonbg:
Will this handle student enrolled multiple courses at the same time?

jhamrick:
Yes, it creates a different subdirectory in the cache folder for each course id.

PR #357, opened 2015-08-12
Add assignment list extension

### LINKED ISSUE #181: Create a "validate" notebook extension
Rather than having people create a new cell to run `nbgrader validate`, if they have nbgrader installed, it would be nice for there to be a notebook extension that adds a toolbar button (or menu option) to validate the notebook and display any errors. Even better, this could actually cause the page to scroll to the errors and maybe highlight them, or something, to make it really obvious.

### LINKED ISSUE #182: Create a "submit" notebook extension
Rather than requiring students to run `nbgrader submit` from the command line (or by creating a new code cell with `!nbgrader submit`), there could be a new button added to the toolbar (or a menu option) that would allow them to submit their assignment. The tricky bit with this, I think, is that I think it would be confusing to have them be able to submit each notebook individually -- it would really be better to have one place to submit the entire assignment. Perhaps something like this could be added to the dashboard?

jhamrick:
This adds a new extension for fetching, validating, and submitting assignments, mainly for use by students. Here's a gif preview of it:

Currently there are ~no tests but I wanted to make a PR with what I had in case anybody has feedback on it!

Ping @ellisonbg 

Fixes #181, fixes #182


ellisonbg:
Wow!!!

jhamrick:
@ellisonbg this extension won't properly handle having different classes as the moment if those classes have assignments with the same name. I think the correct solution is going to be to have something like nbgrader profiles (similar to ipython profiles) that point to different root directories for the class and which have different class names. But, that's not something I am going to have time to implement before classes start, so I am going to go ahead and merge this once travis passes. I'll create an issue in the meantime about creating profiles or something similar.

PR #362, opened 2015-08-18
Switch to travis container builds

### LINKED ISSUE #345: Switch to container builds on travis
Travis is complaining that we're using legacy infrastructure, so we should at some point switch to the new container builds. I tried using them a while back, but had issues with the tests hanging a lot. Maybe with #341 I won't have issues anymore, so we should try the switch and see.

jhamrick:
Fixes #345

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: python
+sudo: false
 python:
   - 2.7
   - 3.4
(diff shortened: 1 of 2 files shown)


Carreau:
Travis seem happy ! +1

jhamrick:
Ugh, of course it's happy now -_- I couldn't get it to pass earlier, which is why I turned on debugging output... but I'll go ahead and merge, and hopefully it'll continue to work. If not I'll need to revisit it again later...

PR #364, opened 2015-08-18
Auto-resize textboxes when giving feedback in formgrader

### LINKED ISSUE #253: Comment box should grow larger for larger comments
If you're leaving a long comment for someone, it would be nice if the comment box automatically got bigger, as it's pretty small to begin with. I don't want to make all the comment boxes bigger, because that would take up too much space -- so having it get bigger only when there's a need for it would be great.

jhamrick:
Fixes #253, original PR #334


jhamrick:
Aha, apparently the issue is that the bind operation isn't supported on PhantomJS until version 2 (see jackmoore/autosize#211), but travis runs less than version 2. I am updating the travis config to install phantomjs 2.0.0 at least for the time being.

PR #389, opened 2015-08-29
Add `requests` to dependencies.

### LINKED ISSUE #338: Add example "boilerplate" notebooks for use by instructors
It may be helpful to add notebooks for instructors to use as a baseline example for setting up and grading assignments. For example, take the examples that are spread out across the user guide and make them into "boilerplate" notebooks without all the additional documentation for instructors. The benefit would be that the instructor would have a quick start to setting up a demo.

Carreau:
Needed for extension installing.
and

closes #338

(it was present in tests because its a flit dependency)

### DIFF
--- a/flit.ini
+++ b/flit.ini
@@ -13,6 +13,7 @@ requires = sqlalchemy
     jupyter_core
     tornado
     six
+    requests
 dev-requires = nose
     pytest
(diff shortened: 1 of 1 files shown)


jhamrick:
Thanks!

PR #396, opened 2015-08-30
Add some clarifications to the docs

### LINKED ISSUE #382: Document the config system better
I don't think it says anywhere where you can actually put config files. You can put them in the same places that Jupyter looks for config files, which are given in the "config" section when you run `jupyter --paths`.

cc @jonathanmorgan

### LINKED ISSUE #391: Questions on documentation/suggestions for improvements
More specific questions from documentation I've looked over:

- http://nbgrader.readthedocs.org/en/stable/user_guide/01_philosophy.html

    - concept of nbgrader file system structure and hierarchy, but no indication I see of where that needs to live and how you reference it across multiple students or multiple instructors.

- http://nbgrader.readthedocs.org/en/stable/user_guide/03_generating_assignments.html

    - mention of needing a "gradebook.db" database, but not specific about where one would put it (implies it is in course folder at same level as "source" folder, I think).
    - code to add assignments to database, and how to then release an assignment, and implication, I believe, that an instructor should run this code as well as the "nbgrader" command in the directory where the "source" directory lives (and it looks like that is the expectation throughout, but I don't think I've seen that explicitly stated).

- http://nbgrader.readthedocs.org/en/stable/user_guide/11_jupyterhub_config.html

    - here, it tells you how to configure an nbgrader_config.py file for formgrade app, but I don't see where one is supposed to put it or how one tells nbgrader where it is.  Makes more sense if each user gets one, or at least you need one in the same folder where you run nbgrader, but doesn't say that.
    - And, if multiple graders, can multiple users log in to an instance of formgrade?  It takes up a port, so if not, per user, would you need separate ports for different instructors, or to coordinate so only one person using at a time?
    - c.HubAuth.notebook_url_prefix - the notes on this imply that it will be relative to the current user's home directory - is this to guide the formgrade app to the course folder?

        - So does this imply that each grader who might use this app will need at least a symbolic link to the course folder in their home folder?
        - Could I make a standard nbgrader_config.py that I put in the course folder that anyone with access could use to launch the formgrader?

- http://nbgrader.readthedocs.org/en/stable/user_guide/07_releasing_assignments.html

    - here it talks about more nbgrader_config.py that will get automatically loaded when we run nbgrader, but it contains different information than is in the nbgrader_config.py in 11_jupyterhub_config.
    - it also doesn't say explicitly where that file needs to live, or where you are running the nbgrader command from (from the directory that holds the nbgrader file system hierarchy?).
    - Sounds like it assumes a single instructor, also?

- http://nbgrader.readthedocs.org/en/stable/user_guide/08_fetching_assignments.html

    - what is the default exchange directory?
    - looks like each student will need an nbgrader_config.py file - will it just live in their home directory?  Can you change it so that they at least are in a child directory, so they don't clutter the home directory?
    how do the commands a student runs know where the database is?

- http://nbgrader.readthedocs.org/en/stable/user_guide/10_collecting_assignments.html

    - how would an instructor/grader's nbgrader_config.py file need to be configured for this to work?  Would they need the same explicit definition of the exchange directory, and then need to run the nbgrader command in the course directory?

jhamrick:
* Closes #391 
* Closes #382

@jonathanmorgan could you have a look at this and tell me if these changes answer your questions?

### DIFF
--- a/docs/source/build_docs.py
+++ b/docs/source/build_docs.py
@@ -66,7 +66,15 @@ def autogen_config(root):
         Configuration options
         =====================
 
-        These options can be set in ``nbgrader_config.py``, or at the command line when you start it.
+        These options can be set in ``nbgrader_config.py``, or at the command
+        line when you start it.
+
+        Note: the ``nbgrader_config.py`` file can be either located in the same
+        directory as where you are running the nbgrader commands (which is most
+        likely the root of your course directory), or you can place it in one of
+        a number of locations on your system. These locations correspond to the
+        configuration directories that Jupyter itself looks in; you can find out
+        what these are by running ``jupyter --paths``.
 
         """
     )
--- a/docs/source/user_guide/01_philosophy.rst
+++ b/docs/source/user_guide/01_philosophy.rst
@@ -12,7 +12,7 @@ assignments will be organized with the following directory structure:
 
 By default, the ``course_directory`` variable is wherever you run the nbgrader commands.
 This means that you can place your class files directory wherever you want.
-However, this location can also be customized (see the :doc:`configuration options <../config_options>`) so that you can run the nbgrader commands from anywhere on your system, but still have them operate on the same directory.
+However, this location can also be customized (see the :doc:`configuration options </config_options>`) so that you can run the nbgrader commands from anywhere on your system, but still have them operate on the same directory.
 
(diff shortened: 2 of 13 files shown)


jhamrick:
Updated to include an example `HubAuth.hub_address`.

jhamrick:
I am going to go ahead and merge this but I'm happy to make more changes later if you think some of your questions aren't fully answered in this PR.

PR #406, opened 2015-09-01
Fix bad celltoolbar flex layout

### LINKED ISSUE #394: Create Assignment toolbar not rendering right in safari
The Create Assignment toolbar is not rendering right in safari at the moment.
screen shot 2015-08-29 at 5 37 19 pm

Carreau:
I think it should do the trick, but my installation is broken for whatever reason and safari give weird result even with the fix I did  yesterday, so I'm not sure. 

Fixes #394

### DIFF
--- a/nbgrader/nbextensions/create_assignment/create_assignment.css
+++ b/nbgrader/nbextensions/create_assignment/create_assignment.css
@@ -35,14 +35,13 @@ div.nbgrader-highlight div.text_cell_render, div.nbgrader-highlight div.input_ar
     border: 1px solid #337ab7 !important;
 }
 
-
 .celltoolbar {
-    // safari fix, we cannot use -webkit-flex on hbox
-    // and vbox either or all css get borked
-    // cf https://github.com/jupyter/nbgrader/issues/394.
-    // manual workaround while #370 is not released. 
+    /* safari fix, we cannot use -webkit-flex on hbox
+    and vbox either or all css get borked
+    cf https://github.com/jupyter/nbgrader/issues/394.
+    manual workaround while #370 is not released. */
     display: -webkit-flex;
-};
+}
 
 div.lock-cell-container {
     margin-right: auto;
--- a/nbgrader/nbextensions/create_assignment/create_assignment.css
+++ b/nbgrader/nbextensions/create_assignment/create_assignment.css
@@ -35,6 +35,15 @@ div.nbgrader-highlight div.text_cell_render, div.nbgrader-highlight div.input_ar
     border: 1px solid #337ab7 !important;
 }
 
+
+.celltoolbar {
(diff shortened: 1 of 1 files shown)


jhamrick:
This seems like it works, but it seems like it has the same effect as just removing the CSS that sets the margin for the icon to `auto` -- which isn't exactly what I want because I'd like that icon to be on the left while everything else is on the right. I'd post a screenshot but apparently I can't taken them while I have bluejeans open :-(

Carreau:
> This seems like it works, but it seems like it has the same effect as just removing the CSS that sets the margin for the icon to auto

:rage: it was working for me the other day :rage: not today.

jhamrick:
This is what it looks like for me :-/

image

nbgrader/nbextensions/create_assignment/create_assignment.css
```diff
@@ -35,6 +35,15 @@ div.nbgrader-highlight div.text_cell_render, div.nbgrader-highlight div.input_ar
...
 }
 
+
+.celltoolbar {
+    // safari fix, we cannot use -webkit-flex on hbox
+    // and vbox either or all css get borked
+    // cf https://github.com/jupyter/nbgrader/issues/394.
+    // manual workaround while #370 is not released. 
+    display: -webkit-flex;
+};
```
jhamrick: Ahhhh I think this semicolon is the culprit. Seems to work if it is removed!

jhamrick: Oops no that only then fixes it for Chrome, Safari is still broken :disappointed:

jhamrick: Ok, got it: you're also using javascript comments rather than CSS, I think, so safari is breaking because of the comment syntax. If it's fixed to `/*...*/` then it seems to work in safari too :smile:

jhamrick:
@Carreau I pushed a commit to your branch and will merge this once tests pass. Thanks for finding the fix for this!

Carreau:
> Ok, got it: you're also using javascript comments rather than CSS, I think, so safari is breaking because of the comment syntax. If it's fixed to /*...*/ then it seems to work in safari too.

This image seem to be an appropriate representation of me in front of CSS. 

angry-y-u-no

Carreau:
hum I need an extra

```
div.lock-cell-container{
    flex-grow: 2;
    margin-left: 9px;
}
```

to have the pad-lock back on the left side .

jhamrick:
Hmm, it seems to look fine to me... did you remove the `margin-right: auto` and replace it with the `flex-grow: 2`? Maybe they both work?

Carreau:
Hum, I though I was on master of both notebook and nbgrader. 

Looks fine in Chrome and Safari for you ?

PR #407, opened 2015-09-01
Add --no-metadata option to nbgrader assign

### LINKED ISSUE #403: nbextension: Don't force cells to have id's
Hi, I'm basically just using nbgrader to make a bunch of manually graded questions followed by answer cells.  I can "assign --no-db" and not have nbgrader care about whether I have added a grade or not (great!).  However, if I don't give my cell "ID"s I can't save my notebook ("Invalid nbgrader cell ID").  It seems to me that this should be an error at the "nbgrader assign" stage, *not* at the editing stage, and that it should be ignored if we do "assign --no-db".

FWIW, I don't think this "error" really stops the notebook from being saved, but it is pretty annoying.

jhamrick:
Fixes #403

This allows you to run `nbgrader assign --no-metadata` and it will not perform any metadata validation or modify it either. Can be used in conjunction with `--no-db` to just replace solution regions and not use the "Create Assignment" toolbar or save anything into the database.

cc @jklymak

### DIFF
--- a/nbgrader/apps/assignapp.py
+++ b/nbgrader/apps/assignapp.py
@@ -27,8 +27,19 @@ flags = {}
 flags.update(nbconvert_flags)
 flags.update({
     'no-db': (
-        {'SaveCells': {'enabled': False}},
-        "Do not save information about grade cells into the database."
+        {
+            'SaveCells': {'enabled': False},
+            'AssignApp': {'no_database': True}
+        },
+        "Do not save information into the database."
+    ),
+    'no-metadata': (
+        {
+            'ClearSolutions': {'enforce_metadata': False},
+            'CheckCellMetadata': {'enabled': False},
+            'ComputeChecksums': {'enabled': False}
+        },
+        "Do not validate or modify cell metatadata."
     ),
     'create': (
         {'AssignApp': {'create_assignment': True}},
@@ -102,6 +113,15 @@ class AssignApp(BaseNbConvertApp):
         )
     )
 
+    no_database = Bool(
+        False, config=True,
+        help=dedent(
+            """
+            Do not save information about the assignment into the database.
+            """
+        )
+    )
+
     def _permissions_default(self):
         return 644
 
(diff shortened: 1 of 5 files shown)


jklymak:
Cool, thanks!  That seems useful.  

To recap for those who didn't follow #403  this means you can just use ### BEGIN SOLUTION and ### END SOLUTION in your cells and nbgrader will just strip those regions.  Lightweight use of nbgrader for sure, but useful for manually graded problem sets.

jklymak:
... well kind of cool.  

I can't install flit, so I can't install the development version of nbgrader.  Probably because I still use python2.7?  So I can't test this.

jhamrick:
Oh, bleh, yeah, because flit is python 3 only. I should create an installation script or something that makes this a bit more friendly to do a development install. In the meantime, you can do:

```
pip install flit-install-py2
flit-install-py2 --symlink
```

and hopefully that should work (that's how I do the installation on travis). I'll make another PR now that reduces the complexity of this installation, sorry again about that!

jklymak:
Hmmm, same error:
```
valdez:nbgrader jklymak$ pip install -e ./
Obtaining file:///Users/jklymak/Downloads/nbgrader
    Complete output from command python setup.py egg_info:
    /Users/jklymak/anaconda/bin/python: No module named flit
    Warning: this setup.py uses flit, not setuptools.
    Behavior may not be exactly what you expect. Use at your own risk!
    /Users/jklymak/anaconda/bin/python -m flit install --deps production
      File "<string>", line 20, in <module>
      File "/Users/jklymak/Downloads/nbgrader/setup.py", line 9, in <module>
        sp.check_call(cmd)
      File "/Users/jklymak/anaconda/lib/python2.7/subprocess.py", line 540, in check_call
        raise CalledProcessError(retcode, cmd)
    subprocess.CalledProcessError: Command '['/Users/jklymak/anaconda/bin/python', '-m', 'flit', 'install', '--deps', 'production']' returned non-zero exit status 1
```

jklymak:
... also I'm not sure how this works for markdown cells (which I also expect answers in).  ### BEGIN SOLUTION renders as heading 3 in markdown.  Even if it is properly stripped by `nbgrader assign` it is inelegant in the "key".

I guess I'm fine w/ using the meta data and just ignoring the warning about not naming the cells (or in my case, removing that check from the main.js source code).

jklymak:
OK, downloaded the individual files that you changed, at it works fine for me.  I can still use the cell metadata to specify read-only or manual-grading, which I like.  ### BEGIN SOLUTION also works in both code and markdown cells.  So, I think this is a decent solution for my needs.  Thanks a lot!

jhamrick:
@jklymak yeah, unfortunately python comment markers are the same as headings in markdown :-/ I could change it so that you don't need to include the comment mark in the markdown cells, but I'm worried that's going to make things too complicated.

One option would be to configure the ClearSolutions preprocessor so that it uses multiline strings instead of comments:

```
c.ClearSolutions.comment_mark = '"""'
c.ClearSolutions.begin_solution_delimeter = 'BEGIN SOLUTION"""'
c.ClearSolutions.end_solution_delimeter = 'END SOLUTION"""'
```

and then in your cells (both code and markdown) you would use `"""BEGIN SOLUTION"""` and `"""END SOLUTION"""` rather than `### BEGIN SOLUTION` and `### END SOLUTION`.

PR #451, opened 2016-03-03
Execute documentation notebooks manually

jhamrick:
This changes the docs workflow so that docs must be executed manually. RTD will still convert notebooks to rst and generate things like the configuration options automatically.


jhamrick:
FYI @willingc . This should make things easier for users, but will mean the executable notebooks can get out of date. Hopefully nbgrader is stable enough at this point that this won't be a huge issue, but I am still a bit worried about it. I think it's complex enough at the moment though that it's worth trying this and seeing how it goes.

jhamrick:
Docs seem to build ok on my fork on RTD, hooray!

willingc:
@jhamrick Cool on the build. For now, I think that this works for the notebooks. Though, I'm hoping that we'll have a better solution sooner vs. later.

PR #464, opened 2016-03-08
Speed up tests

### LINKED ISSUE #463: Make nbgrader apps less slow
Currently, it takes about 1s just to run `nbgrader` with no other arguments. I think this is largely due to overhead with traitlets and the Jupyter application, so I'm not entirely sure how much I can improve on it just with nbgrader.

cc @willingc

jhamrick:
Fixes #463 by having tests create nbgrader apps in the same python session rather than calling out to subprocess. This was a little tricky to do because it involved a lot of monkeypatching of things like sys.stdout, but ultimately I think I have gotten it to work correctly!

### DIFF
--- a/nbgrader/apps/baseapp.py
+++ b/nbgrader/apps/baseapp.py
@@ -19,7 +19,7 @@ from nbconvert.exporters.export import exporter_map
 from nbconvert.nbconvertapp import NbConvertApp, DottedOrNone
 from textwrap import dedent
 from tornado.log import LogFormatter
-from traitlets import Unicode, List, Bool, Instance, Dict, Integer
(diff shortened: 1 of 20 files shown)


Carreau:
Looks good to me. 

How much faster?

jhamrick:
About twice as fast for the python tests (which means 4-5 minutes faster!)

willingc:
Whoa! Nicely done. Does this mean you get all difficult projects done in 5 hours, @jhamrick? ;)

jhamrick:
Haha, this took me a lot longer than 5 hours :-)

Carreau:
Remeber:

https://xkcd.com/1205/

is_it_worth_the_time

Assuming 5min daily. You could have spend a week on it.

PR #467, opened 2016-03-09
unpin jupyter-client

minrk:
Not meant to merge as-is. This is to verify that jupyter/jupyter_client#141 actually works.

Once that's confirmed and the fix released, I'll update this to just unpin jupyter_client.

### DIFF
--- a/flit.ini
+++ b/flit.ini
@@ -10,7 +10,7 @@ requires = sqlalchemy
     nbformat
     traitlets
     jupyter_core
-    jupyter_client<4.2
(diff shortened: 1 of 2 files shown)


jhamrick:
Yay! :cake: 

When you update it to just unpin this, would you mind also unpinning it in the `flit.ini` file?

minrk:
Forgot about flit. Will do.

minrk:
Should be set.

jhamrick:
Thanks!

PR #473, opened 2016-03-28
More helpful default comment box message

### LINKED ISSUE #462: Document that comments can be formatted as markdown
cc @dsblank

jhamrick:
Fixes #462 by changing the default message to:

image

@willingc @dsblank do you think this is clear enough that it will render in the feedback (given that it doesn't actually render in the formgrader interface)? Or do you think I should also add something indicating specifically that markdown added here will render properly in the feedback?

### DIFF
--- a/nbgrader/formgrader/static/js/models.js
+++ b/nbgrader/formgrader/static/js/models.js
@@ -120,7 +120,8 @@ var CommentUI = Backbone.View.extend({
         this.listenTo(this.model, "request", this.animateSaving);
         this.listenTo(this.model, "sync", this.animateSaved);
 
-        this.$comment.attr("placeholder", this.model.get("auto_comment") || "Comments");
+        var default_msg = "Type any comments here (supports Markdown and MathJax)";
+        this.$comment.attr("placeholder", this.model.get("auto_comment") || default_msg);
 
         this.render();
         autosize(this.$comment);


dsblank:
@jhamrick Sure, that is better; thanks!

But, I'm thinking big changes :) For example:

* I'd like to put comments on each cell. Sometimes just commenting here (in the graded part) doesn't allow me to put the comment where I want to put it. And the notebook should have comments on cells anyway!
* I'd like to add more cells when grading
* the feedback should be a notebook (had to jump through hoops to get feedback to students)

jhamrick:
@dsblank yes, I definitely would like to have some of those things, though that is a much bigger change that is going to have to wait until after 0.3.0. Do you want to open some issues for these things?

dsblank:
Perhaps a JEP would be more appropriate?

jhamrick:
I don't think these things need to be in a JEP because they're pretty nbgrader specific, unless you specifically mean the thing for having comments on cells in the live notebook? But for (1) comments on other cells in the formgrader and (3) exporting feedback to notebook format, those don't need to be in a JEP, I think. For (2) adding new cells during grading, I don't think I can support that in the formgrade interface itself, but I could allow an option to view notebooks in the live notebook as editable (currently it exposes them as read-only) so that you could add new cells via that interface.

PR #485, opened 2016-04-11
Update README gif links after docs move into nbgrader

willingc:


jhamrick:
Looks like this is failing due to unrelated changes in dependencies. I opened #486 which should fix that and then after I merge that if you want to rebase this I think it should pass.

jhamrick:
Also thanks for making these changes!

jhamrick:
Ok needs rebase.

jhamrick:
Actually I will just go ahead and merge this, as it doesn't make any other changes so I'm confident the tests will pass 😄

Issue #233, opened 2015-04-09
Configurable plugins for late penalties


jhamrick:
Every class is going to want to institute late penalties differently, so we shouldn't actually include anything in nbgrader itself that calculates the late penalties. However, it would be nice to for example have the ability to write your own plugin to calculate late penalties, which could then be displayed in the formgrader and also when generating feedback.

mandli:
Seconded.

jhamrick:
Closed by #512

PR #512, opened 2016-05-05
Late penalty plugin

lgpage:
Framework for loading plugin functions.
Adds enhancement for #233 

TODO list from the discussions below:
- [x] CSS style class for late penalty in feedback template
- [x] Create separate preprocessor for assigning penalties
- [x] Utilize traitlets to import plugin customizations
- [x] Add tests and fix issues
- [x] Check / add docstrings
- [x] Update docs

### DIFF
--- a/nbgrader/formgrader/templates/feedback.tpl
+++ b/nbgrader/formgrader/templates/feedback.tpl
@@ -42,14 +42,14 @@ div#notebook-container {
   div.cell {
     display: block;
     page-break-inside: avoid;
-  } 
-  div.output_wrapper { 
-    display: block;
-    page-break-inside: avoid; 
   }
-  div.output { 
+  div.output_wrapper {
     display: block;
-    page-break-inside: avoid; 
+    page-break-inside: avoid;
+  }
+  div.output {
+    display: block;
+    page-break-inside: avoid;
   }
 }
 
@@ -126,6 +126,9 @@ span.nbgrader-label {
             <li><a href="#comment-{{ cell.metadata.nbgrader.grade_id }}">Comment</a></li>
             {% endif %}
           {% endfor %}
+          {% if resources.nbgrader.late_penalty %}
+            <li style="color: #d2413a;">Late submission penalty (Score: -{{ resources.nbgrader.late_penalty | float | round(2) }})</li>
+          {% endif %}
           </ol>
         </div>
       </div>
@@ -171,7 +174,7 @@ span.nbgrader-label {
   <span class="nbgrader-label">Grade cell: <code>{{ cell.metadata.nbgrader.grade_id }}</code></span>
   {{ score(cell) }}
 {%- endif -%}
-</div>  
+</div>
 {%- endmacro %}
 
 {% macro nbgrader_footer(cell) -%}
@@ -225,7 +228,7 @@ span.nbgrader-label {
   </div>
 
   {%- else -%}
-  
+
   <div class="inner_cell">
     <div class="input_area">
       {{ cell.source | highlight_code(metadata=cell.metadata) }}
--- a/nbgrader/formgrader/templates/feedback.tpl
+++ b/nbgrader/formgrader/templates/feedback.tpl
@@ -92,6 +92,10 @@ div.nbgrader_cell .input_area {
     color: black;
 }
 
+li.late-penalty {
+    color: #d2413a;
+}
+
 span.nbgrader-label {
     line-height: 25px;
 }
@@ -126,8 +130,8 @@ span.nbgrader-label {
             <li><a href="#comment-{{ cell.metadata.nbgrader.grade_id }}">Comment</a></li>
             {% endif %}
           {% endfor %}
-          {% if resources.nbgrader.late_penalty %}
-            <li style="color: #d2413a;">Late submission penalty (Score: -{{ resources.nbgrader.late_penalty | float | round(2) }})</li>
+          {% if resources.nbgrader.late_penalty > 0 %}
+            <li class="late-penalty">Late submission penalty (Score: -{{ resources.nbgrader.late_penalty | float | round(2) }})</li>
           {% endif %}
           </ol>
         </div>
--- a/nbgrader/plugins/latesubmission.py
+++ b/nbgrader/plugins/latesubmission.py
@@ -5,7 +5,7 @@ from .base import BasePlugin
 
 
 class LateSubmissionPlugin(BasePlugin):
-    """Predifined methods for assigning penalties for late submission"""
+    """Predefined methods for assigning penalties for late submission"""
 
     penalty_method = Enum(
         ('none', 'zero'),
--- a/nbgrader/plugins/latesubmission.py
+++ b/nbgrader/plugins/latesubmission.py
@@ -1,14 +1,10 @@
 from textwrap import dedent
-from traitlets import List
 from traitlets import Unicode
-from traitlets import validate
 
-from .base import BasePluginLoader
+from .base import BasePlugin
 
 
-class LateSubmissionPlugin(BasePluginLoader):
-
-    supported_methods = List(['none', 'zero', 'custom'])
+class LateSubmissionPlugin(BasePlugin):
 
     penalty_method = Unicode(
         'none',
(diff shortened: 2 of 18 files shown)


lgpage:
@jhamrick this is what I've come up with so far (still a bit to add), let me know what you think.

jhamrick:
This is really awesome! I'm super excited about this 🎉 

I do wonder though if we want to necessarily define a custom plugin loader class, rather than relying on traitlets' built-in functionality. With traitlets, you can define class variables so that instances of the class are loaded, and the type of that instance is configurable. A good example is in nbconvert, where things like the writer and postprocessor are configurable instances: https://github.com/jupyter/nbconvert/blob/master/nbconvert/nbconvertapp.py#L181 e.g. for the writer:

```python
    # Writer specific variables
    writer = Instance('nbconvert.writers.base.WriterBase',  
                      help="""Instance of the writer class used to write the 
                      results of the conversion.""", allow_none=True)
    writer_class = DottedObjectName('FilesWriter', config=True, 
                                    help="""Writer class used to write the 
                                    results of the conversion""")
    writer_aliases = {'fileswriter': 'nbconvert.writers.files.FilesWriter',
                      'debugwriter': 'nbconvert.writers.debug.DebugWriter',
                      'stdoutwriter': 'nbconvert.writers.stdout.StdoutWriter'}
    writer_factory = Type(allow_none=True)

    def _writer_class_changed(self, name, old, new):
        if new.lower() in self.writer_aliases:
            new = self.writer_aliases[new.lower()]
        self.writer_factory = import_item(new)
```

this will by default use the `FilesWriter` class for the writer but it can be customized by passing in any class name. So if you had a file in the current directory called `my_custom_writer.py` with a class called `MyWriter`, you could set `writer_class = 'my_custom_writer.MyWriter'` and then the `writer` variable will get automatically populated with an instance of the given class. So one option is that we could do something similar with late plugins. What do you think?

I have a few smaller comments that I'll leave inline, too.

nbgrader/formgrader/templates/feedback.tpl
```diff
@@ -126,6 +126,9 @@ span.nbgrader-label {
             <li><a href="#comment-{{ cell.metadata.nbgrader.grade_id }}">Comment</a></li>
             {% endif %}
           {% endfor %}
+          {% if resources.nbgrader.late_penalty %}
+            <li style="color: #d2413a;">Late submission penalty (Score: -{{ resources.nbgrader.late_penalty | float | round(2) }})</li>
```
jhamrick: This `li` tag should probably be given a class and then define the color using that class in the CSS at the top of the page.

nbgrader/preprocessors/getgrades.py
```diff
@@ -1,9 +1,10 @@
 from .. import utils
 from ..api import Gradebook
+from ..plugins import LateSubmissionPlugin
 from . import NbGraderPreprocessor
 
 
-class GetGrades(NbGraderPreprocessor):
+class GetGrades(NbGraderPreprocessor, LateSubmissionPlugin):
```
jhamrick: Rather than inheriting from the plugin here, I would add a new preprocessor whose job it is to handle late penalties -- maybe called `AssignLatePenalties` -- and then add that preprocessor to the autograde app. That way late penalties are assigned during autograding, rather than during feedback (since not everyone might use nbgrader-generated feedback, but might want to rely on nbgrader to calculate late penalties).

jhamrick: Also I should note if we do the traitlets thing I mentioned in the main comments, the late submission plugin would then be a trait of the preprocessor, rather than having the preprocessor inherit from it.

nbgrader/formgrader/templates/feedback.tpl
lgpage: Agreed. Did this just to quickly test it.

lgpage:
> I do wonder though if we want to necessarily define a custom plugin loader class, rather than relying on traitlets' built-in functionality... So one option is that we could do something similar with late plugins. What do you think?

This makes sense. I'll definitely look into it.
This is the first time I've used traitlets and unfortunately the documentation is lacking. It has been slow going trying to figure things out, so thanks for pointing out this example.

jhamrick:
No problem, and yeah, the traitlets stuff can be pretty confusing. I remember being incredibly confused when I was first learning about it maybe a year ago. There's at least a bit more documentation now than there was then! These links might also be helpful, and definitely let me know if you have any questions...

* http://traitlets.readthedocs.io/en/stable/using_traitlets.html#dynamic-default-values
* http://traitlets.readthedocs.io/en/stable/using_traitlets.html#callbacks-when-trait-attributes-change
* http://traitlets.readthedocs.io/en/stable/trait_types.html#classes-and-instances

Note that nbgrader still has some of the traitlets 4.0 syntax, while there was new syntax introduced in 4.1: http://traitlets.readthedocs.io/en/stable/migration.html which reminds me, I should open an issue to remind myself to migrate the traitlets stuff!

jhamrick:
Aha, I also tracked down an example of using the `Instance` traitlet in nbgrader itself, if that's helpful too:

https://github.com/jupyter/nbgrader/blob/master/nbgrader/apps/formgradeapp.py#L78

This one is a little different (and simpler) from the writer example. I'm actually not totally sure why the writer example is so complicated, as it is possible to also pass just the string of a class name for `authenticator_class` which seems much simpler (see https://nbgrader.readthedocs.io/en/latest/configuration/jupyterhub_config.html#configuring-nbgrader-formgrade )

lgpage:
@jhamrick I still need to add tests to check this, but it seemed rather easy to utilize traitlets as you suggested.

nbgrader/plugins/latesubmission.py
```diff
@@ -0,0 +1,24 @@
+from textwrap import dedent
+from traitlets import Unicode
+
+from .base import BasePlugin
+
+
+class LateSubmissionPlugin(BasePlugin):
+
+    penalty_method = Unicode(
```
jhamrick: There is an [enum trait type](http://traitlets.readthedocs.io/en/stable/trait_types.html#traitlets.Enum) which might be good to use here to ensure user typos throw an error.

nbgrader/plugins/latesubmission.py
```diff
@@ -0,0 +1,24 @@
...
+            Predefined methods:
+                'none':
+                'zero':
+            """
+        ),
+    ).tag(config=True)
+
+    def late_submission_penalty(self, student_id, score, total_seconds_late):
+        self.log.info("Using late submission penalty method: {}".format(self.penalty_method))
+        if self.penalty_method == 'zero':
```
jhamrick: Should this also check the `total_seconds_late`, and only assign the penalty is the assignment is in fact late?

Ah, I see in the preprocessor you check for that already. Well, I know it's a little redundant, but it might be good to throw an error as a safeguard if `total_seconds_late` is zero.

jhamrick:
👍 This looks great! Let me know when you are ready for me to review the tests and docs.

nbgrader/plugins/latesubmission.py
lgpage: Thanks :)  I was looking for the best way to do this validation.

nbgrader/plugins/latesubmission.py
lgpage: Will do.

lgpage:
@jhamrick ready for review.

jhamrick:
Just wanted to say really quickly that I think this looks great -- I've skimmed over it and will have a longer look at it on Monday (I am about to leave for a conference in Singapore, so I'll be offline for the rest of the weekend).

jhamrick:
Another quick update -- so sorry that I haven't had time to finish the review for this. I am definitely going to try to get to it this week (hopefully tomorrow)!

lgpage:
No worries. I have been swamped with work lately, so probably would not have been able to get round to any updates had you done it earlier :P

nbgrader/tests/apps/test_nbgrader_autograde.py
```diff
@@ -9,6 +9,17 @@
 from .. import run_nbgrader
 from .base import BaseTestApp
 
+PLUGIN = """
```
jhamrick: Since this is only used in one place, I would define it in the function where you use it rather than having it be a global variable.

jhamrick:
This looks really great. I only have the one comment about that plugin variable -- if you address that then I'll merge! Looks like this will make it in to the 0.3.0 release, too 😄

lgpage:
Awesome :D 
I'll try update this PR this evening or over the weekend.

lgpage:
Done

Issue #511, opened 2016-05-01
Pre-processing of downloaded submission files


lgpage:
I would like to propose adding a new feature of being able to pre-process submission files.

For example, institutions using BlackBoard can download assignment submission files in a `.zip` archive with the following filename format (give or take some extra garbage in the filename):
`{couse_id}{assignment_id}{timestamp}.zip`

Each of the files in this archive will have the following filename format (give or take some extra garbage in the filename):
`{assignment_id}{student_id}{timestamp}{upload_filename}.{ext}`

I propose allowing for the following (optional) directories in the nbgrader directory structure:
`{course_directory}/downloaded/archive/{zipfiles}`
`{course_directory}/downloaded/{assignment_id}/extracted/{raw_files}.ipynb`
`{course_directory}/downloaded/{assignment_id}/invalid/{raw_files}.{ext}`

The pre-processing step will do two things:
1. extract the archive file/s to `{course_directory}/downloaded/{assignment_id}/extracted/`
2. process these extracted files into `{course_directory}/downloaded/{assignment_id}/invalid/` or the submitted path `{course_directory}/submitted/{student_id}/{assignment_id}/`

This needs to be done in a pluggable way (similar to what is mentioned in #233) as filenames will vary from user to user. There may also the possible option of automatically being able to update the students in the database, based on this processing.

jhamrick:
This seems like a reasonable thing to have. I guess this would call for a new command, maybe something along the lines of `nbgrader import` (I am thinking "import" specifically because I've been thinking of having some sort of "export" application related to #233 and also #436, so having it be "import" would give it a nice symmetry).

Definitely it should be pluggable -- I'm thinking maybe we can define a new type of `ImportPlugin` class (or something along those lines) that does all the heavy lifting. We could maintain a few popular ones (e.g. blackboard and canvas) but still allow other people to write their own if they are using something different.

Would you like to take a stab at a PR for this?

lgpage:
A new command would be needed, I was thinking something along the lines of `nbgrader stage`, although `import` does give nice symmetry to `export`

For pluggability, my initial thought was to allow for specific functions in the `nbgrader_config.py` file that allows for the customized behaviour. Wrapping this in a `PluginManager` (of sorts) for default behaviours and/or supported behaviours (e.g for blackboard and canvas) is a good idea :)

With respects to #233, allow something like this in the `nbgrader_config.py` file:
```python
def late_penalty(student_id, score, duedate, extension, timestamp):
[...]

PR #526, opened 2016-07-04
Processing of LMS downloaded submission files

lgpage:
This PR address the enhancement discussed in  #511

**TODO list:**
- [x] Agree on app name
- [x] Agree on work-flow and directory structure
- [x] Create app and plugin
- [x] ~Create LMS specific plugins (Blackboard, Canvas, etc.)~ Generic plugin should suffice
- [x] Create tests
- [x] Update docs

At the moment there is no code to speak of for this PR, but I'm creating the PR early to more easily discuss the app name, work-flow, and directory structure.

**App Name:**
At the moment this is called `StageApp` (`nbgrader stage`) which I do not like, but can't think of anything better yet. @jhamrick you made mention of calling it `ImportApp` (`nbgrader import`). Also not sure I like this as it may be confused with Pythons import machinery.

Can sub-commands be hyphenated or underscored? So something like `nbgrader lms-collect` or `nbgrader lms-unstage` ??

**Work-flow and Directory Structure:**
My initial though is to add 2 directories:
- `{course_id}/LMS/submitted/{assignment_id}/archive/`
  - for downloaded `{archive.zip}` files
  - the contents of which would be extracted to the 2nd directory
  - `--unzip` option to first extract these archives
  - `--force` option to overwrite existing files when extracting
- `{course_id}/LMS/submitted/{assignment_id}/files/`
  - for the `{rawfiles}.{ext}` extracted from the `.zip` files

The extracted files will then be moved / copied into the `nbgrader` submission directory:
- `{course_id}/submitted/{student_id}/{assignment_id}/{notebook_id}.ipynb`
  - Plugin to do the heavy lifting and processing of filenames
  - `{student_id}` (at least) must come from the extracted LMS file filename
  - There can also be the option to auto-update the database with the `{student_id}`
  - Same `--force` option to overwrite existing files

A few questions come to mind:
- Should these directories be created by `nbgrader` at some point? (`nbgrader assign` maybe?) Or purely rely on the instructor to create them?
- What to do with, what I will call, the garbage `.ext` files?
  - Blackboard, for example, archives 2 files per student (the file submitted by the student; and a text file containing comments from the submission web-page)
  - Personally I would just have the app delete these text files, but others may want to keep or even use them (possible even have the comments embedded in the notebook during processing?)
- What to do with invalid submission files?
  - Students, for example, uploading `.pdf`, etc files in place of `.ipynb`


jhamrick:
> At the moment this is called StageApp (nbgrader stage) which I do not like, but can't think of anything better yet. @jhamrick you made mention of calling it ImportApp (nbgrader import). Also not sure I like this as it may be confused with Pythons import machinery.

Hmm, true, I hadn't thought about that. I kind of liked the duality of `nbgrader import/export` (where `export` would be like exporting to CSV, or something) but I agree it could be confusing. 

I believe commands can be hyphenated. I quite like `nbgrader lms-collect`, actually, as it mirrors the existing `nbgrader collect`.

The workflow and directory structure sound good to me!

> Should these directories be created by nbgrader at some point? (nbgrader assign maybe?) Or purely rely on the instructor to create them?

Hmm, I think probably we should just leave it up for the instructor to create them (with detailed instructions in the docs for what to call the directories) because not everybody using nbgrader will be using a LMS.

> What to do with, what I will call, the garbage .ext files?

I agree that some other instructors might want to keep those files, so I'd be tempted to keep them. We could have an option that is like `--remove-extras` or something, which would optionally remove them if you didn't want to keep them.

> What to do with invalid submission files?

Hrm, that is tough because it depends on how instructors wanted to have students submit their files, I guess. What I mean by that is, obviously if they submit a PDF we can't have nbgrader handle that -- but what if instructors asked students to submit a zip archive of all their files? How are you having students turn in their assignments -- do you just have one notebook per assignment, or are you having them turn in the notebooks separately? I guess the issue with doing it separately too is that some instructors might want students to turn in something else (e.g. a data file that they collected or something) in which case an archived file might really be more appropriate...

lgpage:
Awesome. Thanks for the feedback Jess.

> I quite like nbgrader lms-collect, actually, as it mirrors the existing nbgrader collect.

I also like `lms-collect` for the same reason, so I'll go with this.

> I guess the issue with doing it separately too is that some instructors might want students to turn in something else (e.g. a data file that they collected or something) in which case an archived file might really be more appropriate...

I'll play around with a few options of how to handle this once I've done a bit more work on this PR.

At the moment I have been caught up with other work (converting course material from Latex to Notebooks for next year) so have not had time to work on this PR yet. This is however something I will in any case need for next year, so I plan on having it done before the end of this year.

jhamrick:
> At the moment I have been caught up with other work (converting course material from Latex to Notebooks for next year) so have not had time to work on this PR yet. This is however something I will in any case need for next year, so I plan on having it done before the end of this year.

Sounds great! And no worries, I've also been quite busy this summer. Good luck with converting all your assignments; I did that for my class too (converting from a combination of latex and matlab to notebooks) and can't say it was the most enjoyable experience 😩

lgpage:
@jhamrick when you get some time could you please start reviewing this.

I changed the name of this app again :wink: ``lms-collect`` may give the impression there is some sort of LMS integration, so I called it ``zip-collect``.

I still need to add CI tests.

jhamrick:
@lgpage sorry I haven't had a chance to look at this yet -- but I just wanted to let you know it is on my todo list! I will try to look at it tomorrow.

lgpage:
@jhamrick awesome thanks, but no rush I know its quite a bit of code.

nbgrader/docs/source/user_guide/managing_assignment_files_manually.ipynb
```diff
@@ -0,0 +1,558 @@
...
+    "%%bash\n",
+    "\n",
+    "cat submitted_zip/hacker/ps1/timestamp.txt"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This is an issue with the underlying ``dateutils`` package used by ``nbgrader``. But not to worry, we can easily create a custom collector plugin to correct the timestamp strings when the files are collected:"
```
jhamrick: Add a link here to the documentation page on custom zipcollect plugins.

nbgrader/apps/zipcollectapp.py
```diff
@@ -0,0 +1,522 @@
...
+        default_value=False,
+        help="Force overwrite of existing files."
+    ).tag(config=True)
+
+    strict = Bool(
+        default_value=False,
+        help="Skip submitted notebooks with invalid names."
+    ).tag(config=True)
+
+    auto_update_database = Bool(
```
jhamrick: This option seems like it is redundant with `--create` in `nbgrader autograde`.

... which I realize is technically deprecated. I opened #668 to fix this. But I think this is probably an option that is more robust to have in  `nbgrader autograde` rather than here.

nbgrader/apps/zipcollectapp.py
```diff
@@ -0,0 +1,522 @@
...
+        "UTC",
+        help="Timezone for recording timestamps"
+    ).tag(config=True)
+
+    timestamp_format = Unicode(
+        "%Y-%m-%d %H:%M:%S %Z",
+        help="Format string for timestamps"
+    ).tag(config=True)
+
+    collector_plugin = Type(
```
jhamrick: I think unfortunately we're going to have to add support for another type of plugin, which is an extractor plugin. I think in a lot of cases the filenames in people's submissions won't have their usernames in them -- rather, the username will be at a higher level of the directory structure. For example, for Berkeley, we download a giant zip file of all submissions which when extracted has the directory structure like this:

```
Problem Set 1/LAST, FIRST(123456)/Submission attachment(s)/name_of_submission.zip
```

Of course, I could write a separate script to copy these zip files into the `downloaded` directory and rename them. But even them, the files inside won't have the right files names. Even better would be if there was an extractor plugin that could handle both (1) associating the student names with the zipfiles even though there is a directory hierarchy and (2) rename the files within the archive to have student names.

nbgrader/apps/zipcollectapp.py
```diff
@@ -0,0 +1,522 @@
...
+        extracted = 0
+        for _, _, extracted_files in os.walk(extracted_path):
+            extracted += len(extracted_files)
+
+        if cnt_files != extracted:
+            self.log.warn(
+                "File count mismatch. Processed or extracted {} files, but "
+                "only found {} files in {}\nThis may be due to the archive "
+                "(zip) file/s either containing duplicates or sub-directories"
+                "".format(number_of_files, extracted_file_count, extract_to)
```
jhamrick: number_of_files and extracted_file_count are not defined

jhamrick:
It's really nice that you've structured it already to have the two steps of extract and collect, because I do think that separation is perfect. However, I think some of the logic that is in the collect step is perhaps more appropriate for the extract step. Here's my suggestion:

1. The extract step comes with a plugin that will examine the full path (before extraction) and determine the student id from that. By default maybe we can make it just like you have it---i.e. based on the filename or the name of the archive.
2. The extract step will then extract files (or copy them, if they are not an archive) to a path that includes the student id, such as `{downloaded}/{assignment_id}/{extracted_directory}/{student_id}`.
3. The collect step will now be able to determine the student id from the path. By default again it can just assume the files are flat in the extracted directory, but you can write a plugin to change this.

lgpage:
@jhamrick thanks for this awesome review. I'll go through your comments and change requests in detail :)

nbgrader/apps/zipcollectapp.py
lgpage: I added this here mainly for the reason that a new student entry could be created - or an existing student entry could be updated - with information like first name, last name, and email that could be pulled from the filename or looked up from an external file in a custom collect plugin.

jhamrick: I see. I think though that it is in that case also redundant with `nbgrader db student import`, which can also read in students from an external file. I just want to avoid duplicating too much functionality across the different parts of nbgrader.

lgpage: Makes sense :smile: I will remove it.

lgpage:
So the main thing to address here is the extract step, in that it needs to be able to handle more generic cases (If not out-of-the-box at least via plugins). I typing out few cases [here](https://gist.github.com/lgpage/17e85800687b0aeac3b5b824c4b92b02) (that come from the comments) just so I can get things straight in my head and I realized this could get unnecessarily messy.

So I propose a different approach:
Modify the default extract process to "walk" all ``.zip`` files and keep the ``.zip`` file structure, for example:
- ``downloaded/assignment_id/archive/main_download.zip``
  - ``first_last_{student-id}_timestamp.zip`` (for each student - for each attempt)
    - {file-id}.ext (per release file)
    - timestamp.txt
should get extracted to:
- ``downloaded/assignment_id/extracted/main_download/``
  - ``first_last_{student-id}_timestamp/`` (for each student - for each attempt)
    - {file-id}.ext (per release file)
    - timestamp.txt

Then the collect process remains as is and the ``student_id`` only needs to provided in the collect step.

>Currently it seems that ``zip_collect`` makes the assumption that for every archive:
>- It is a flat archive and contains no subdirectories
>- All the files in the archive are named in a way that the student id can be recovered from the filename. You can write your own configurable plugin, but it still is only given the filename which means it can't do much else (e.g. get the student id from the original archive name instead).

The collect process does walk ``downloaded/assignment_id/extracted/`` and passes the absolute path to the default plugin. Granted the default plugin only applies the regexp to the base filename, but this can easily be changed.

> I think unfortunately we're going to have to add support for another type of plugin, which is an extractor plugin.

I do agree the extract code should be in a plugin just in case the default is not good enough.

jhamrick:
Yes, that seems reasonable to me, I think that's pretty similar to how I had envisioned it. I think even in the case of "archive of archives" (the second in your list of examples) I would be fine requiring instructors to unzip the big archive by hand as long as they can configure nbgrader to take care of the rest (then you wouldn't need to run the extract step twice).

Just to clarify though, in your first example you said that the student id can only be determined in the collect step. Under your proposed solution, though, it can be determined on the extract step since you're going to walk the zip archive?

> The collect process does walk downloaded/assignment_id/extracted/ and passes the absolute path to the default plugin. Granted the default plugin only applies the regexp to the base filename, but this can easily be changed.

Yes, but if the student id was only present in the name of the archive (and not in the names of the extracted files), it will no longer be available at the extracted step.

nbgrader/docs/source/user_guide/managing_assignment_files_manually.ipynb
```diff
@@ -0,0 +1,558 @@
...
+   "metadata": {},
+   "source": [
+    "If this is not the case and you are using nbgrader in a shared server environment, you can do this with the ``nbgrader release`` command (see :doc:`managing_assignment_files`)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Collecting assignments"
```
lgpage: Added subheadings in the docs

lgpage:
>Just to clarify though, in your first example you said that the student id can only be determined in the collect step. Under your proposed solution, though, it can be determined on the extract step since you're going to walk the zip archive?

Technically yes it would be possible, but a true pain if doing so inside an open ``zipfile`` handle.

>Yes, but if the student id was only present in the name of the archive (and not in the names of the extracted files), it will no longer be available at the extracted step.

Agreed, but now it is available in the collect step as the name of the archive is kept in the file path.

lgpage:
I have no idea why Travis is failing :cry:

lgpage:
@jhamrick Travis was being screwy is that it was pointing to the wrong commit in its build :confused: 
Deleting and restoring this branch seems to have sorted it out.

I believe I have implemented all changes requested, if you wouldn't mind checking it again. Thanks :smile:

jhamrick:
Ok, I will look at this again as soon as I can! (It may not be until next week though, sorry---I am totally swamped until then)

lgpage:
> Ok, I will look at this again as soon as I can! (It may not be until next week though, sorry---I am totally swamped until then)

No worries, I understand, whenever you get a chance. Thanks :smile: 
#681 is also be ready for review for whenever you get a chance.

jhamrick:
Ok, going through this now! I am going to try to use this to collect some old assignments I have from Berkeley's LMS, and will report how well it goes.

nbgrader/apps/zipcollectapp.py
```diff
@@ -0,0 +1,486 @@
...
+            self.log.warn("Clearing existing files in {}".format(path))
+            rmtree(path)
+            os.makedirs(path)
+        else:
+            self.fail(
+                "Directory not empty: {}\nuse the --force option to clear "
+                "previously existing files".format(path)
+            )
+
+    def _find_student(self, info):
```
jhamrick: I wouldn't require students to be in the database at this step -- I would allow the collector to collect any and all submisisons (which is the same behavior as `nbgrader collect`) and then have the autograder step be the one the throws an error when the student isn't in the database.

nbgrader/apps/zipcollectapp.py
```diff
@@ -0,0 +1,486 @@
...
+            submitted_path = self._format_path(
+                self.submitted_directory, info['student_id'], self.assignment_id)
+            dest_path = os.path.join(submitted_path, submission)
+
+            timestamp = self.get_timestamp()
+            if 'timestamp' in info.keys():
+                timestamp = info['timestamp'] or timestamp
+                if isinstance(timestamp, six.string_types):
+                    try:
+                        timestamp = parse_utc(timestamp)
```
jhamrick: There is a bug here -- `self.get_timestamp()` returns a string, but this returns a datetime object. I think probably you want `self.get_timestamp()` to return a datetime object instead.

nbgrader/apps/zipcollectapp.py
```diff
@@ -0,0 +1,486 @@
...
+            )
+
+    def _find_student(self, info):
+        try:
+            with open_gradebook(self.db_url) as gradebook:
+                return gradebook.find_student(info['student_id'])
+        except MissingEntry:
+            return None
+
+    def get_timestamp(self):
```
jhamrick: I think here you want to get the datetime in UTC format, because when you parse the timestamp below (line 371) it is converted into UTC format without a tzinfo. So by making it have a tzinfo here, you will later get an error like this:

```
TypeError: can't compare offset-naive and offset-aware datetimes
```

jhamrick: Or alternately you can just call `parse_utc` here as well, e.g.:

```
    def get_timestamp(self):
        """Return the timestamp using the configured timezone."""
        tz = gettz(self.timezone)
        if tz is None:
            self.fail("Invalid timezone: {}".format(self.timezone))
        return parse_utc(datetime.datetime.now(tz).isoformat(' '))
```

nbgrader/apps/zipcollectapp.py
```diff
@@ -0,0 +1,486 @@
...
+            dest_path = os.path.join(submitted_path, submission)
+
+            timestamp = self.get_timestamp()
+            if 'timestamp' in info.keys():
+                timestamp = info['timestamp'] or timestamp
+                if isinstance(timestamp, six.string_types):
+                    try:
+                        timestamp = parse_utc(timestamp)
+                    except ValueError:
+                        self.log.warn(
```
jhamrick: I am not sure if this is the right behavior. I would be inclined to say if there is no timestamp associated with a file, then it doesn't get a timestamp at all.

In particular, if the individual files don't have timestamps associated with them but there is one file in the submission, e.g. `timestamp.txt` that contains a timestamp, then what will happen is all the files get the current time. Then below (line 440), the max of all the timestamps is taken. This means that even if a user writes a custom plugin to parse the timestamp out of `timestamp.txt`, the default timestamp will always be used since it is guaranteed to be newer.

nbgrader/apps/zipcollectapp.py
```diff
@@ -0,0 +1,486 @@
...
+                    data[student_id]['dest_files'].append(dest_path)
+                else:
+                    ind = data[student_id]['file_ids'].index(submission)
+                    old_timestamp = data[student_id]['timestamps'][ind]
+                    if timestamp >= old_timestamp:
+                        data[student_id]['timestamps'][ind] = timestamp
+                        data[student_id]['src_files'][ind] = _file
+                        data[student_id]['dest_files'][ind] = dest_path
+                    else:
+                        self.log.warn("Skipped. Older timestamp found.")
```
jhamrick: I don't fully understand this behavior. Why would you keep the oldest version of a file, rather than the newest?

Regardless, I would perhaps suggest throwing an error if multiple files with the same name are found instead. It seems to me that different instructors would have different ways they'd want to handle this, and I don't think nbgrader should try to be too smart in guessing which it is.

nbgrader/apps/zipcollectapp.py
```diff
@@ -0,0 +1,486 @@
...
+                self.log.warn(
+                    "Skipped. Student {} not found in gradebook."
+                    "".format(info['student_id'])
+                )
+                invalid_files += 1
+                continue
+
+            submitted_path = self._format_path(
+                self.submitted_directory, info['student_id'], self.assignment_id)
+            dest_path = os.path.join(submitted_path, submission)
```
jhamrick: It would be nice if this is something that could be returned in `info` so it could be configured by a plugin.

Also, if I have subdirectories in my assignments, then this will cause all files to be moved out of the subdirectories into a flat directory, which is undesirable. I think the default behavior should be to preserve directory structure.

jhamrick:
For reference, here is my custom collector plugin for reading timestamps from a file:

```
class CustomPlugin(FileNameCollectorPlugin):
    def collect(self, submission_file):
        if os.path.basename(submission_file) == "timestamp.txt":
            regex = ('.*/'
                '(?P<student_id>[^/]+)/'
                '(?P<file_id>\w+)')
            info = re.match(regex, submission_file).groupdict()
            with open(submission_file, "r") as fh:
                timestamp = datetime.strptime(fh.read()[:-3], "%Y%m%d%H%M%S")
            info['timestamp'] = timestamp.isoformat(' ')
        else:
            info = super(CustomPlugin, self).collect(submission_file)

        return info
```

And corresponding `nbgrader_config.py` options (these assignments contain matlab files, too):

```
c.FileNameCollectorPlugin.named_regexp = (
    '.*/'
    '(?P<student_id>[^/]+)/'
    'Submission attachment\(s\)/'
    '.*/'
    '(?P<file_id>\w+)'
)
c.FileNameCollectorPlugin.valid_ext = [".m", ".mat", ".txt"]
```

nbgrader/plugins/zipcollect.py
```diff
@@ -0,0 +1,190 @@
...
+    :class:`~nbgrader.apps.zipcollectapp.ZipCollectApp`.
+    Extractor plugin subclasses MUST inherit from this class.
+    """
+
+    force = Bool(
+        default_value=False,
+        help="Force overwrite of existing files."
+    ).tag(config=True)
+
+    zip_ext = List(
```
lgpage: I added support for tarfiles as well as zip files.

nbgrader/apps/zipcollectapp.py
lgpage: I have removed this requirement.

nbgrader/apps/zipcollectapp.py
lgpage: I agree with your review lower down:
>I would be inclined to say if there is no timestamp associated with a file, then it doesn't get a timestamp at all.

So this is no longer needed, and I will remove it.

nbgrader/apps/zipcollectapp.py
lgpage: ``self.get_timestamp()`` will be removed so this shouldn't be a problem.

nbgrader/apps/zipcollectapp.py
lgpage: I do agree "if there is no timestamp associated with a file, then it doesn't get a timestamp at all". I will update for this.

This also made me think what if ``.txt`` is added to ``valid_ext`` to collect the ``timestamp.txt`` files, with the current behaviour, it would be overwritten. I will update for this as well.

nbgrader/apps/zipcollectapp.py
lgpage: In this case I think it is the warn message that is misleading. Submission files with older timestamps are skipped.

This timestamp comparison is something I would like to keep. For example if the archive file contains multiple attempts for a given submission then there will be duplicate files. I can update the code so the logic is more readable and easier to follow.

nbgrader/apps/zipcollectapp.py
lgpage: >Directory structure within assignments should be preserved

I need to double check this, but the structure should be preserved by including the subdirectories (the relative path to the assignment) in ``file_id``.

lgpage:
A summary of changes made based on your review are:
A) Don't require students to be in the database until the ``nbgrader autograde`` step.
B) Added support for tarfiles
C) Preserved the directory structure within assignments by providing the relative path to the assignment in ``file_id``
D) Updated timestamp logic:
  1. If no timestamp is provided one is not created by defaulting to the current time
  2. Allowed for the collection of ``timestamp.txt`` without overwriting it
  3. Made the timestamp comparison logic more verbose*
  4. Fail on duplicate submission files if no timestamps are provided


>Regardless, I would perhaps suggest throwing an error if multiple files with the same name are found instead.

I have done this for duplicate file submissions if no timestamps are provided. *I still think the timestamp comparisons are still needed, for example if the archive file contains multiple attempts for a given submission then there will be duplicate files.

jhamrick:
Awesome! I will have a look at it again this weekend.

> I still think the timestamp comparisons are still needed, for example if the archive file contains multiple attempts for a given submission then there will be duplicate files.

That makes sense.

nbgrader/apps/zipcollectapp.py
```diff
@@ -0,0 +1,492 @@
...
+        any non-archive files will be copied, to the `extracted_directory`:
+
+            {downloaded}/{assignment_id}/{extracted_directory}
+
+        After which the files in the `extracted_directory` will be collected
+        and copied into the students `submitted_directory`:
+
+            {submitted_directory}/{student_id}/{assignment_id}/{notebook_id}.ipynb
+
+        By default the collection of files in the `extracted_directory` is
```
jhamrick: These docs are out of date (I think should be `FileNameCollectorPlugin`?)

lgpage: Thanks :smile: I'll make this change and also one day I will learn to use grep :wink:

jhamrick: I think you're still missing a reference to `FileNameProcessor` (and `--processor`) in the paragraph immediately after this one 😕 

I don't know what text editor you use, but if you happen to use Sublime, you can right click a folder in the left-hand file browser and select "Find in folder" to search all files within a folder. It's super useful, and a lot less painful than grep. grep in this particular case isn't too bad though:

```
grep -R FileNameProcessor .
``` 

and 

```
grep -R "\-\-processor" .
```

(both of those only reveal this one instance, so once that's fixed I think it should be good to go 😄 )

lgpage: Should be fixed now. Thanks for catching this.

jhamrick:
Awesome. Feel free to merge once the tests pass again.

lgpage:
Yay :smile: Thanks again @jhamrick for your time reviewing :+1:

PR #538, opened 2016-08-27
Use official HubAuth from JupyterHub

### LINKED ISSUE #520: Rely on official HubAuth in JupyterHub
See https://github.com/jupyterhub/jupyterhub/pull/554

jhamrick:
Fixes #520

cc @minrk -- no need for review or anything, I just wanted to let you know I've finally done this! I ended up not using the tornado mixin because my handlers aren't *always* authenticated with JupyterHub. But I am now using `jupyterhub.services.auth.HubAuth` rather than my own thing.


minrk:
Great! Since you are the first user of this outside the single-user server itself, I'd be happy for some feedback if there's anything that should be better.

jhamrick:
I think for the most part it works pretty well. One suggestion might be to have the default login url just be `'/hub/login'` (rather that including the full url at the beginning) because otherwise the `?next=...` redirection doesn't work (since JupyterHub will refuse absolute redirects, only relative ones).

minrk:
Great, thanks. I think that's something to fix on both ends - the redirect filter should recognize its own URL.

PR #546, opened 2016-09-09
Make it clearer how to set due dates

### LINKED ISSUE #540: Document how to set assignment due dates
Update the example in the user guide (http://nbgrader.readthedocs.io/en/stable/user_guide/creating_and_grading_assignments.html#configure-assignments-and-students-in-the-config-file) so that it also includes due dates.

jhamrick:
Fixes #540


willingc:
Nice! LGTM

PR #562, opened 2016-09-24
Add better support to transfer apps for multiple courses

### LINKED ISSUE #554: Fetching assignments should use ~/course_id/assignment
Currently, if a student is in two (or more) courses, they can switch back and forth between by editing their nbgrader_config.py file. However, if two courses both have "Assignment1" then that causes problems when fetching because they both try to copy the files into ~/Assignment1.

A better solution would be for fetched assignments to go to ~/course_id/Assignment1

jhamrick:
This adds a new option, `path_includes_course`, to the TransferApps so that they work a little better with multiple courses. If this option is set to True, then `nbgrader fetch` will fetch assignments to `<course_id>/<assignment_id>` (rather than just to `<assignment_id>`). Similarly, if the option is set to True, then `nbgrader submit` will look for assignments at `<course_id>/<assignment_id>`. I have also made sure that this functionality works from the assignment list extension.

This fixes #554 though it doesn't fully address the full issue with multiple courses yet (i.e. #544). I am going to try to address that fully in a separate PR, this is just a first step.

cc @dsblank

### DIFF
--- a/nbgrader/apps/baseapp.py
+++ b/nbgrader/apps/baseapp.py
@@ -465,6 +465,18 @@ class TransferApp(NbGrader):
         config=True,
         help="Local cache directory for nbgrader submit and nbgrader list. Defaults to $JUPYTER_DATA_DIR/nbgrader_cache")
 
+    path_includes_course = Bool(
+        False, config=True,
+        help=dedent(
+            """
+            Whether the path for fetching/submitting  assignments should be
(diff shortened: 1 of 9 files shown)


dsblank:
Fantastic! This goes a long way to making it more straightforward to use nbgrader for jupyterhub institutions with many instructors. Thanks!

PR #563, opened 2016-09-25
Add ability to switch between courses in assignment list extension

jhamrick:
This partially (but not fully) address #544. It adds a dropdown menu to the assignment list extension that allows students to switch between different classes:

I still need to add tests for this (and current tests are probably going to fail).

cc @dsblank @randy3k @kthyng @fcurrin


jhamrick:
I should mention that this doesn't do anything regarding access control -- so all students will still be able to see all classes. This should just hopefully make it easier for students to be able to ignore classes that they're not enrolled in. Doing the access control is going to be trickier, so I am going to try to address that in a separate PR. I think this will still be useful even after that though: if a student is in multiple classes they will be able to switch between them and can see individually what they've submitted for each class.

dsblank:
Great! I'm going to use this as soon as it is ready... I have 50% of my students in multiple classes, and they are constantly having to edit their config file. This will lessen the cognitive load.

jhamrick:
Ok, should be ready to use! Let me know if there's anything about it that doesn't quite work the way you expect, or if you run into any issues with it.

fcurrin:
Hi, I was wondering if this works with the newest release of nbgrader. We just updated to 0.3.3 and would like to retain this capability but don't want to replace the files if that would interfere with other changes made in the update.

jhamrick:
This will work with the upcoming release (which you can install with the `--pre` flag), which is 0.4.0.

> don't want to replace the files if that would interfere with other changes made in the update.

Sorry, I'm not sure I understand the issue. Could you elaborate?

PR #584, opened 2016-12-15
Changes for Notebook v4.3 tests

### LINKED ISSUE #582: Travis and local tests fail with notebook v4.3
Gets stuck at a login page:
selenium screenshot

Also gets stuck at a login page

Need to update to handle token authenticaiton:
xref: https://github.com/jupyter/notebook/pull/1831

lgpage:
Turns out it is actually easy to disable token auth for the nbextension tests and for the FormgradeNotebookApp.

It will be a bit more work if you actually want to use token auth for the FormgradeNotebookApp.

Closes #582

### DIFF
--- a/nbgrader/apps/notebookapp.py
+++ b/nbgrader/apps/notebookapp.py
@@ -1,9 +1,11 @@
 from tornado import ioloop
 from notebook.notebookapp import NotebookApp
 
+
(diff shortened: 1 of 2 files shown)


jhamrick:
Thanks! I think this is fine for now, probably we should handle token authentication at some point but that can be some point in the future 😄

PR #588, opened 2016-12-19
`basename` to wordslist

lgpage:
It seems `nbconvert` was updated to v5.0 and released perfectly between my PR's and merge, causing the docs builds to fail on master.


jhamrick:
Haha, I was wondering how that happened. Thanks for looking into it! 👍

PR #590, opened 2016-12-22
Processing of invalid `notebook_id`s

### LINKED ISSUE #577: Don't include notebooks in database that don't exist on disk
c.f. https://groups.google.com/forum/#!topic/jupyter-education/w261tgt43fs

lgpage:
Fixes #577 

cc @hkarl


lgpage:
This is more or less done. So feel free to start reviewing the code.

I need to
- fix what I currently broke in the tests,
- add new tests and docs, and
- see what happens in `JupyterHub` when the submit assignment fails with the `--strict` flag.

jhamrick:
This is really great, thanks! One thing that might be useful to include in this is an actual notification in the assignment list extension that gives a visual error to the student if they are using the extension. Otherwise it seems like currently the error will only be visible from the command line?

lgpage:
>One thing that might be useful to include in this is an actual notification in the assignment list extension that gives a visual error to the student if they are using the extension. Otherwise it seems like currently the error will only be visible from the command line?

Yes at the moment this warning is only seen on the command line. I will update the assignment list extension a bit as suggested.

jhamrick:
@lgpage Do you think you'll have time to work on this more before we release 0.4.0 (in maybe a week or two)? No pressure if not, I just want to know whether I should move it to the next milestone or not.

lgpage:
Got back from holiday yesterday, so I will be able to work on this.

jhamrick:
Awesome, thanks!

nbgrader/nbextensions/assignment_list/assignment_list.js
```diff
@@ -413,7 +432,15 @@ define([
...
                     dataType : "json",
-                    success : $.proxy(that.on_refresh, that),
+                    success : function (data, status, xhr) {
+                        if (!data.success) {
+                            that.submit_error(data);
+                            button.text('Sumbit');
+                            button.removeAttr('disabled');
+                        }
+                        //$.proxy(that.on_refresh, that);
+                        //GRRRR
```
lgpage: @jhamrick I am a little bit stuck with this. How do I incorporate this refresh when the submission is successful?

lgpage:
Example of error in `assignment_list` extension:
selection_006

nbgrader/nbextensions/assignment_list/assignment_list.js
jhamrick: I *think* you should be able to do `that.on_refresh()`, rather than using the `$.proxy`. The way you're doing it here you're creating a closure with this function you've defined -- which is sort of what `$.proxy` is doing in the first place. JavaScript is confusing and I don't entirely understand why that's necessary, but if you don't create the closure (e.g. `success : that.on_refresh`), then when the function is called it won't be associated with the relevant instance of the class. So that's why the `$.proxy` is there to begin with, but I think if you remove it here then it shouldn't be a problem.

nbgrader/nbextensions/assignment_list/assignment_list.js
```diff
@@ -413,7 +432,15 @@ define([
                     },
                     type : "POST",
                     dataType : "json",
-                    success : $.proxy(that.on_refresh, that),
+                    success : function (data, status, xhr) {
+                        if (!data.success) {
+                            that.submit_error(data);
+                            button.text('Sumbit');
```
jhamrick: Typo here

jhamrick:
This looks awesome! 🎉 

Let me know if you have any questions in getting the tests written for the js frontend stuff -- those tests are kind of a nightmare. I think you should be able to take the `test_validate_failure` test as a starting point, since that already checks for a modal dialog and stuff.

jhamrick:
Oh, though I would say it should not break if there are `.ipynb_checkpoints` -- those should be allowed to be submitted. Actually in general it might be best for it to always throw an error if there are missing files, and just give a warning if there are new (unexpected) files. I can imagine situations in which we actually ask that students submit a new file along with their assignment (e.g. a dataset they collected or something) and we shouldn't give an error in those cases.

nbgrader/nbextensions/assignment_list/assignment_list.js
lgpage: Thanks :)

lgpage:
At the moment only an error (with the `strict` flag set) is shown in the frontend. I don't see an easy way to show a warning in the frontend other than some output string searching, which I'm not a real fan of.

jhamrick:
Yeah, I wouldn't do the string searching either. I think it's fine to just show an error in the frontend and quietly ignore the warnings.

lgpage:
@jhamrick ready for review :)

nbgrader/apps/submitapp.py
```diff
@@ -18,8 +21,25 @@
...
 flags.update(transfer_flags)
 flags.update({
+    'strict': (
+        {'SubmitApp': {'strict': True}},
+        "Fail if the submission is missing notebooks for the assignment"
+    ),
 })
 
+
+# XXX Move to utils.py maybe?
```
jhamrick: Yes, should be fine to move to utils.py.

Issue #599, opened 2016-12-26
Fix running of invoke on windows


jhamrick:
cf https://github.com/conda-forge/staged-recipes/pull/2125

We can try this fix of adding an invoke configuration file: https://github.com/pyinvoke/invoke/issues/345#issuecomment-225557831

This should probably be fixed in a patch release, i.e. 0.3.2

lgpage:
I will try verify this first with a rudimentary `echo` config stuff to file hack in staged recipes.

lgpage:
@jhamrick I tested the patch in PR #601 on a local branch for the staged recipe [here](https://github.com/lgpage/staged-recipes/blob/nbgrader-patch/recipes/nbgrader/tasks.patch).

The only problem now is Python 2.7 (x86) on windows regularly times out in the `AssignApp`. Appveyor logs [here](https://ci.appveyor.com/project/lgpage/staged-recipes/build/1.0.2). Any ideas on fixing this?

Also if you would prefer to target this change to the `0.4.0` milestone instead of a patch release we can simply leave the patch in the staged recipe until then.

jhamrick:
Hmm, I am not totally sure, but it's possible it is just a fluke -- sometimes the tests randomly time out :-/ When you say they regularly time out, does that mean you've tried restarting them and they still fail?

If so, then I don't really know without being able to look at it more closely. The error `RuntimeError: Timeout waiting for IOPub output` usually means that it has successfully execute the cell and is waiting for the cell outputs (see https://github.com/jupyter/nbconvert/blob/master/nbconvert/preprocessors/execute.py#L282 ). I don't know why that would be an issue for this particular test, though, especially when the previous ones seem to have run successfully. And it's particularly puzzling because that's not the sort of thing I would expect to differ between python versions 😕 

> Also if you would prefer to target this change to the 0.4.0 milestone instead of a patch release we can simply leave the patch in the staged recipe until then.

Sure, that sounds fine to me.

lgpage:
Regularly might be overstating it. There were 2 jobs where both failed only for Python2.7 (x86) within 5min into the tests.

Ill re-run the jobs a few times and see if "regularly" is indeed the case.

lgpage:
This is somewhat of a certainty. I re-ran the AppVeyor job another 5 times and for all runs Python2.7 (x86) failed, yet Python2.7 (x64) passed :confused:

lgpage:
[...]

PR #601, opened 2016-12-28
Add shell config for invoke on windows

lgpage:
Partially for #599

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -1,7 +1,7 @@
 import os
 import re
 
-from invoke import task
+from invoke import task, collection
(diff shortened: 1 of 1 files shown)


jhamrick:
👍 Seems good to me. I'll wait on merging this just until we can verify that the issues we're having with the windows tests aren't related (I highly doubt they are, but just in case).

lgpage:
I am going to go ahead and merge this as the issues we're having with the windows tests aren't related.

willingc:
Thanks @lgpage. I'm glad you are contributing here ☀️

PR #614, opened 2017-01-06
Don't be as strict about solution delimeters

### LINKED ISSUE #422: Fuzzy string checking for begin/end solution regions
If there's a typo, then they won't be detected, which can cause problems. If there's something close, a warning should at least be printed.

cc @jonathanmorgan

### LINKED ISSUE #551: Be more flexible in finding solution delimeters
See the discussion in #542

This would essentially revert #523, which is unfortunate in that it will result in yet another API change, but I think it will ultimately make things simpler as @dsblank suggests.

jhamrick:
Fixes #551 and fixes #422 

I didn't explicitly include fuzzy string checking for #422 but that's because the delimeters are actually pretty close, so this is difficult to do -- for example, "BEGIN SOLUTION" and "END SOLUTION" actually share a large amount of text because of the "SOLUTION" part. But, I think being able to be flexible about everything else (e.g. comment marks) will still be sufficient for that.

cc @dsblank

### DIFF
--- a/nbgrader/preprocessors/clearsolutions.py
+++ b/nbgrader/preprocessors/clearsolutions.py
@@ -1,3 +1,5 @@
+import re
+
 from traitlets import Dict, Unicode, Bool, observe
 from textwrap import dedent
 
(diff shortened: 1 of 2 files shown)


dsblank:
Less code, more power! Win win! Thanks... look forward to trying this!

PR #630, opened 2017-01-11
Don't fail `test_check_version` test on `'import sitecustomize' failed error`

lgpage:
On the conda forge linux image the stupid `'import sitecustomize' failed` error pops up which causes the `test_check_version` test to fail, eg: https://circleci.com/gh/lgpage/staged-recipes/48

This PR updated the test to only check the last line output, I.e. only checks the `nbgrader` version info and not anything else.

### DIFF
--- a/nbgrader/tests/apps/test_nbgrader.py
+++ b/nbgrader/tests/apps/test_nbgrader.py
@@ -30,6 +30,10 @@ class TestNbGrader(BaseTestApp):
 
     def test_check_version(self, capfd):
         """Is the version the same regardless of how we run nbgrader?"""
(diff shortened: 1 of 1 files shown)


lgpage:
Is it necessary to check the Python version info?

jhamrick:
It's not strictly necessary, but I'd like to keep the check in there as it's an additional check that the installed nbgrader executable is the same as the python that's currently being executed.

lgpage:
Ok, will update this PR.

lgpage:
@jhamrick updated, please check again :)

jhamrick:
Looks great! Thanks!

PR #637, opened 2017-01-13
Update conda channel to conda-forge

lgpage:
I just noticed I forgot to update the conda channel in the README :(


jhamrick:
Good catch!

PR #644, opened 2017-01-14
Add AppVeyor CI files

lgpage:


appveyor.yml
```diff
@@ -0,0 +1,68 @@
...
+  CMD_IN_ENV: "cmd /E:ON /V:ON /C obvci_appveyor_python_build_env.cmd"
+
+  # Workaround for https://github.com/conda/conda-build/issues/636
+  PYTHONIOENCODING: "UTF-8"
+
+  matrix:
+    - TARGET_ARCH: "x64"
+      CONDA_PY: "34"
+      PY_CONDITION: "python >=3.4,<3.5"
+    - TARGET_ARCH: "x64"
```
jhamrick: Does python 3.6 fail to build? Do you know why?

appveyor.yml
```diff
@@ -0,0 +1,68 @@
...
+    - cmd: conda update --yes --quiet conda
+    - cmd: set PYTHONUNBUFFERED=1
+    - cmd: conda config --add channels conda-forge
+    - cmd: conda info
+    - cmd: conda create -n nbgrader-dev "%PY_CONDITION%"
+    - cmd: conda env update
+    - cmd: activate nbgrader-dev
+
+    # Install non-conda packages
+    - cmd: pip install pytest-rerunfailures
```
jhamrick: Rather than installing this explicitly maybe we can do `pip install -r dev-requirements-windows.txt` below?

environment.yml
```diff
@@ -0,0 +1,25 @@
+name: nbgrader-dev
```
jhamrick: Hmm, I'm tempted to say I'd prefer to try to install everything via pip to make sure that that method of installation still works on windows (I'm more confident that conda will work reliably than pip, TBH). And then we don't have to maintain two separate files with dependencies.

appveyor.yml
```diff
@@ -0,0 +1,68 @@
...
+    - cmd: npm install phantomjs-prebuilt
+    - cmd: set "PATH=%CD%\\node_modules\\phantomjs-prebuilt\\lib\\phantom\\bin;%PATH%"
+
+# Skip .NET project specific build phase.
+build: off
+
+test_script:
+    - cmd: python setup.py develop
```
jhamrick: This line could be replaced with `pip install -r dev-requirements-windows.txt -e .` which will install all regular dependencies and dev dependencies and install the package itself.

appveyor.yml
```diff
@@ -0,0 +1,68 @@
...
+    - cmd: jupyter nbextension install --sys-prefix --py nbgrader
```
jhamrick: These lines shouldn't be necessary, *I think*. In the tests I believe temporary directories are created to test the nbextension installation.

appveyor.yml
lgpage: Yes python 3.6 fails to build, at least with the conda env approach I have taken here, as there are a huge amount of missing python 3.6 conda dependencies. They are in the process of re-rendering current feedstocks to get python 3.6 packages out so it would be long until this can be added.

appveyor.yml
lgpage: Can do. I did this as it is the only package needed for the tests :smile:

environment.yml
lgpage: Either way, If we take a ``pip`` approach or ``conda`` approach I would really like to have this file in the repo as it would make it easier to setup a fresh conda dev env for nbgrader.

lgpage:
@jhamrick I followed a ``conda`` env approach here, as it was easy to copy and paste from what I had lying around :wink: and I know for previous experience AppVeyor can sometimes be a bitch when it comes to missing system packages not installed when using ``pip``.

lgpage:
In any case I'll play around with a ``pip`` approach version tomorrow.

appveyor.yml
jhamrick: Ah, I see. Yeah, I would say let's build with pip then if that's the case, as I'd ideally like to keep the 3.6 builds.

environment.yml
jhamrick: That's true. How about not including it for this PR, and then make another PR with an environment file and then also update the dev installation docs telling people how they can create a new conda dev env with it?

jhamrick:
> AppVeyor can sometimes be a bitch when it comes to missing system packages not installed when using pip

Ah, I see. Hmm. Well, let's try it with pip, and if you can't get it to work then maybe we can use conda but it would be great if people were able to install with pip. That's how I was installing it on Jenkins, so it *should* be able to work.

appveyor.yml
lgpage: Do you also want x86 tests?

jhamrick: No, I think x64 is fine.

lgpage:
@jhamrick please kill this build: https://ci.appveyor.com/project/jupyter/nbgrader/build/1.0.4

jhamrick:
Done. Do you not have permission to kill/restart builds yourself? Let me see if I can fix that...

lgpage:
I don't, thanks :smile:

jhamrick:
OK, try now?

lgpage:
Nope, still not.

jhamrick:
Hmm. Are you logged in with GitHub, or with a username and password? I made you an admin on the repo, so if you're logged in with GitHub, it should work...

lgpage:
I'm logging in with GitHub. Restarted my browser, clear cache and cookies and still not :cry:

jhamrick:
Ok, that was confusing and complicated. I think I figured it out now -- I figured access would just be inherited from GitHub (like is the case on Travis) but not so. Let me know if it still isn't working...

lgpage:
I have access now thanks :D and the tests should be passing. I hate phantomjs !! Using the AppVeyor system version leads to that completely unrelated failure occurring with the fixture finalizer :angry:

lgpage:
@jhamrick well the tests are passing, sort of... It looks like we will have to fix the underlying IOPub TimeOut issue for this to be useful. At the moment it is failing more often than not due to this issue.

jhamrick:
This is really, awesome, thanks for getting this working! I am going to go ahead and merge it despite the appveyor tests not fully passing, and we can try to address the IOPub issue in a separate PR. Pretty annoying that that's happening, but I guess some slow continuous integration that causes sporadic issues is better than no continuous integration at all 😆

lgpage:
The best would be to post a PR for ``nbconvert`` to make the iopub timeout configurable instead of hard coded here https://github.com/jupyter/nbconvert/blob/master/nbconvert/preprocessors/execute.py#L278.

jhamrick:
Yeah, good idea, I'll do that.

PR #649, opened 2017-01-14
Bump the number of test reruns even higher

jhamrick:
This is an interim effort to deal with the IOPub timeouts we're getting in the windows tests, until https://github.com/jupyter/nbconvert/pull/513 is merged and released. Not sure if this will actually help all that much, but let's see!

cf #644 
cc @lgpage

### DIFF
--- a/nbgrader/preprocessors/execute.py
+++ b/nbgrader/preprocessors/execute.py
@@ -21,14 +21,29 @@ class Execute(NbGraderPreprocessor, ExecutePreprocessor):
         """)
     ).tag(config=True)
 
-    def preprocess(self, nb, resources):
+    execute_retries = Integer(0, help=dedent(
+        """
+        The number of times to try re-executing the notebook before throwing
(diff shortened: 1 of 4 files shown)


jhamrick:
Meh. So far, does not seem to really help.

jhamrick:
Woo! Ok, the combination of bumping the number of reruns, and also re-executing notebooks if necessary, seems to have done the trick. Hacky solution, but good enough for now at least! 🎉

lgpage:
Found [this](http://phantomjs.org/troubleshooting.html) while trying to sort some other issue I have with running tests:
>On Windows, the default proxy setting may cause a massive network latency (see Known Issues in the release note). The workaround is to disable proxy completely, e.g. by launching PhantomJS with --proxy-type=none command-line argument.

I wonder if this would help with running tests?

jhamrick:
Hmm, interesting. I don't know if it will fix the problem but it's worth a try!

PR #655, opened 2017-01-15
Fix saving notebook in create assignment tests

jhamrick:
Sometimes when saving the notebook in the create assignment tests, the save operation doesn't actually complete before we try to read it. This makes sure the save operation has fully finished first.

### DIFF
--- a/nbgrader/tests/nbextensions/test_create_assignment.py
+++ b/nbgrader/tests/nbextensions/test_create_assignment.py
@@ -117,9 +117,17 @@ def _save(browser):
 
 def _save_and_validate(browser):
     _save(browser)
-    _wait(browser).until(
-        EC.visibility_of_element_located((By.CSS_SELECTOR, "#notification_notebook")))
 
+    # wait for notebook_saved.Notebook event
+    def notebook_saved(browser):
+        elem = browser.find_element_by_css_selector("#notification_notebook")
+        if not elem.is_displayed():
+            return False
+        if elem.text == "Notebook saved":
+            return True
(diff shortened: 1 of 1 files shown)


jhamrick:
@MeeseeksDev hello

jhamrick:
@meeseeksdev hello

jhamrick:
Hmm, not sure this is yet working, as I am still seeing `RERUN` in the nbextensions tests. Hard to debug locally, since it is a race condition that only seems to really occur on slow CI systems. I'm not sure what else I can do, though, since theoretically this event only occurs after the atomic save is complete... going to go ahead and merge this, hopefully it will be fine for now.

lgpage:
Ok that bot is awesome !!!

jhamrick:
I know!! @Carreau wrote it. So nice to just have that work automatically!

I think it will only respond to certain users at the moment, but if you want to use it maybe @Carreau can give you access too?

Carreau:
> I think it will only respond to certain users at the moment, but if you want to use it maybe @Carreau can give you access too?

Sure, let me know, there is a [whitelist of org and users](https://github.com/MeeseeksBox/MeeseeksDev/blob/master/meeseeksdev/__init__.py#L4-L22) for now until it's reviewed enough to make sure there are no security flaws.

lgpage:
@Carreau yes please, that would be awesome :D

Carreau:
> @Carreau yes please, that would be awesome :D

You should be whitelisted, it will may take a few minutes to deploy.

PR #665, opened 2017-01-20
Preventing URI Encoding of the base-url in the assignment_list extension

whitead:
The base-url is not expected to be encoded in the server-side jupyter web handlers, so certain character (e.g., @)  were causing 404s due to mismatch between the nbextension URLs and the server extension API URLs.

You can see an example of how this is dealt with in the jupyter project [here](https://github.com/jupyter/jupyter_server/blob/dd4b052403cc0ce70dbde9bfc6c71d6b09e814ec/notebook/static/services/config.js#L26) and [here](https://github.com/jupyter/jupyter_server/blob/dd4b052403cc0ce70dbde9bfc6c71d6b09e814ec/notebook/static/services/contents.js#L45).

This is an issue because in my JupyterHub deployment, usernames are email addresses which contain `@`, which becomes `%40` when encoded. This means the URLs for the AJAX calls are invalid.

I left the URI encoding of the file paths, as is done in other jupyter extensions. 

This is a very helpful project, by the way. Thank you.

### DIFF
--- a/nbgrader/nbextensions/assignment_list/assignment_list.js
+++ b/nbgrader/nbextensions/assignment_list/assignment_list.js
@@ -35,7 +35,6 @@ define([
         this.data = undefined;
     };
 
-
     CourseList.prototype.bind_events = function () {
         var that = this;
         this.refresh_element.click(function () {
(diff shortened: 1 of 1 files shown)


jhamrick:
Thanks! I will wait for the spurious test failure to rerun and then will merge.

jhamrick:

PR #670, opened 2017-01-22
Fix deprecation warnings for config options

### LINKED ISSUE #669: Deprecation warnings aren't being printed
I think this might have to do with migrating to the new traitlets syntax in #612 . But for example, if you run `nbgrader assign --create`, it does not print a warning, even though it should.

jhamrick:
Fixes #669


jhamrick:

Issue #390, opened 2015-08-29
Hide autograder test cells


dementrock:
Is it possible to have the autograder test cells hidden for the release notebook? If not, where in the code should I change to make this possible?

jhamrick:
No, unfortunately, it's not possible to do that at this time because there
is no way to hide cells in the notebook itself. One possibility that might
work is to introduce another special syntax for selectively removing some
tests (eg '### BEGIN TEST' and '### END TEST') kind of how there is for
solutions, but this would still require you to include *some* tests so that
students don't just see a bunch of blank cells.

On Saturday, August 29, 2015, dementrock <notifications@github.com> wrote:

> Is it possible to have the autograder test cells hidden for the release
> notebook? If not, where in the code should I change to make this possible?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jupyter/nbgrader/issues/390>.
>


-- 
Cheers,
Jess

hkarl:
I'm facing the same issue...  Let's assume we are content with having autograder cells in the released notebook and just want to remove some of the tests in each cell. 
 
From what I understand so far, there are various options: 

1. Simple and ugly: post-process the files in release/../*.ipynb by a "sed s/..//g " and remove all content between marker lines. Should work nicely, but clearly not elegant. 
2. Try to fiddle with the Preprocessors, but that does not seem trivial as AssignApp first calls ComputeChecksum and then SaveCell. When stripping content after SaveCell, we would have to recompute Checksums, it seems? Else, a final Preprocessor in this chain shouldn't be difficult to write. 
2. Subclass FilesWriter and use a new class instead that strips out all tests between marker lines. That's perhaps the cleanest solution. I'm trying this, but stumbling across nbconvert issues (needs checking, see https://github.com/CN-UPB/nbgrader/blob/master/nbgrader/writers/nbgraderwriter.py ). 

I'd appreciate any pointers here; what do you think is the most promising approach? 

Thanks a lot, 

Holger

jhamrick:
Yes, the tricky part is definitely going to be making sure that validation doesn't break. Your second suggestion (using preprocessors) is the way I would do it. Your proof-of-concept I think is on the right track, and I like the idea of having two separate checksums -- that's cleaner I think than the way I was originally thinking of doing it, which was actually adding a new field into the database for the cell contents with and without hidden tests.

hkarl:
[...]

Issue #556, opened 2016-09-13
Error when creating assignment


whositwhatnow:
Finally figured out how to get this going but am running into an issue when creating my first assignment.
I copy assignment i created into the correct directory.  when i run nbgrader assign assignement i get hte following error.  I have zero clue what im loking for here 

[root@jupyterhub lab1]# nbgrader assign ps1
[AssignApp | INFO] Updating/creating assignment 'ps1': {}
[AssignApp | INFO] Converting notebook /var/courses/BIO362Assignments/lab1/source/./ps1/Exercise 1 - DNA is a string.ipynb to notebook
[AssignApp | ERROR] There was an error processing assignment: /var/courses/BIO362Assignments/lab1/source/./ps1
[AssignApp | ERROR] Traceback (most recent call last):
      File "/usr/local/lib/python3.5/site-packages/nbgrader/api.py", line 1615, in find_source_cell
        .filter(SourceCell.name == name, Notebook.name == notebook, Assignment.name == assignment)\
      File "/usr/local/lib/python3.5/site-packages/sqlalchemy/orm/query.py", line 2724, in one
        raise orm_exc.NoResultFound("No row was found for one()")
    sqlalchemy.orm.exc.NoResultFound: No row was found for one()

    During handling of the above exception, another exception occurred:

    Traceback (most recent call last):
      File "/usr/local/lib/python3.5/site-packages/nbgrader/api.py", line 1644, in update_or_create_source_cell
        source_cell = self.find_source_cell(name, notebook, assignment)
      File "/usr/local/lib/python3.5/site-packages/nbgrader/api.py", line 1618, in find_source_cell
        raise MissingEntry("No such source cell: {}/{}/{}".format(assignment, notebook, name))
    nbgrader.api.MissingEntry: No such source cell: ps1/Exercise 1 - DNA is a string/Read_16

    During handling of the above exception, another exception occurred:

    Traceback (most recent call last):
      File "/usr/local/lib/python3.5/site-packages/sqlalchemy/engine/base.py", line 1139, in _execute_context
        context)
      File "/usr/local/lib/python3.5/site-packages/sqlalchemy/engine/default.py", line 450, in do_execute
        cursor.execute(statement, parameters)
    sqlite3.IntegrityError: constraint failed

    The above exception was the direct cause of the following exception:

    Traceback (most recent call last):
      File "/usr/local/lib/python3.5/site-packages/nbgrader/api.py", line 1587, in add_source_cell
        self.db.commit()
[...]

PR #681, opened 2017-01-25
Hide tests in "Autograder tests" cells

lgpage:
Address the enhancement discussed in #390 

See also: #476, [#556](https://github.com/jupyter/nbgrader/issues/556#issuecomment-246806864) and #685 

Closes: #390 

cc @hkarl @ellisonbg @jharbig


nbgrader/docs/source/configuration/student_version.rst
```diff
@@ -102,3 +107,69 @@ can be configured through the ``ClearSolutions.text_stub`` option:
...
+include notebooks of different languages within the same assignment:
+
+.. code:: python
+
+    c = get_config()
+    c.ClearHiddenTests.begin_test_delimeter = {
+        "python": "### BEGIN HIDDEN TESTS",
+        "javascript": "// BEGIN HIDDEN TESTS"
+    }
+    c.ClearHiddenTests.end_test_delimeter = {
```
lgpage: This is wrong, `text_delimeter` configs are Unicode not Dict, need to update docs.
I think this may also be wrong in the docs for the clearsolutions preprocessor.

nbgrader/tests/apps/test_nbgrader_autograde.py
```diff
@@ -568,6 +568,72 @@ def test_update_newer_single_notebook(self, course_dir):
...
+            join(course_dir, "source", "ps1", "p3.ipynb")
+        )
+        run_nbgrader(["assign", "ps1", "--db", db])
+
+        # make sure hidden tests are removed in release
+        with open(join(course_dir, "release", "ps1", "p2.ipynb"), 'r') as nb:
+            source = nb.read()
+        assert "BEGIN HIDDEN TESTS" not in source
+
+        self._copy_file(
```
jhamrick: This is buggy test -- it should copy the files from the release directory, not from the original `files` directory.

jhamrick:
Thinking about this some more, I think actually the change would be pretty trivial -- if you add `ComputeChecksums` to the list of preprocessors a second time (so it appears both before the cells are saved into the database, and then after the hidden tests are removed) then I think that should work.

nbgrader/tests/apps/test_nbgrader_autograde.py
lgpage: Yup, that is right, not quite sure how I missed this.

lgpage:
> you have a test for this, but the test is buggy -- if you fix the test you'll find that it fails

I have clean up this test a bit and fixed the buggy nature

> if you add ComputeChecksums to the list of preprocessors a second time (so it appears both before the cells are saved into the database, and then after the hidden tests are removed) then I think that should work.

Yup, this works like a dream :D

jhamrick:
👍 looks good to me!

I am going to release 0.4.0 and then merge this after that's done.

jhamrick:
Actually, I think everything has been merged that needs to be for 0.4.0, so it should be fine to merge this. First PR of 0.5.0! 🎉

nadimmajed:
Hello,  I did a test like the following : 

```
assert squares(1) == [1]
### BEGIN HIDDEN TESTS
assert squares(2) == [1, 4]
assert squares(10) == [1, 4, 9, 16, 25, 36, 49, 64, 81, 100]
assert squares(11) == [1, 4, 9, 16, 25, 36, 49, 64, 81, 100, 121]
### END HIDDEN TESTS
```

The first assert was evaluated but the rest was not. 
My question is, did you intend to make the assert between the delimiters not evaluable?
Thank you in advance.

jhamrick:
In the student version of the assignment, the tests in the hidden tests region won't be evaluated -- they are no longer present in the assignment. The way the hidden tests work is that they are removed for the student version, and added back in for autograding.

nadimmajed:
How can I modify the source code to make them hidden and evaluable in the student version?

lgpage:
When you say "evaluable in the student version" do you mean (a) by the student running the test cell or do you mean (b) via ``nbgrader validate``?

nadimmajed:
I mean (a).

lgpage:
Unfortunately, the short answer is, there is no way to make tests both hidden and evaluable in this manner.

nadimmajed:
Thank you for everything.

lgpage:
@nadimmajed you could perhaps post an example of the assignment question, and what you are trying to achieve, on the [Jupyter education forum](https://groups.google.com/forum/#!forum/jupyter-education). Someone may have an alternative method or solution; or alternative way of structuring the assignment, etc.

nadimmajed:
I can hide the test cell with "hide_code" extension and then the tests will be evaluable and hidden in the same time. But I like the fact that nbgrader present a notebook for the instructor where the tests are displayed and a notebook for the student where the tests are hidden.
Thank you for the advice, I will post a question in the forum.

lgpage:
I am not sure which extension exactly you are referring to, but I am 99% sure it achieves this by simply toggling the `display` property of a HTML div element, which means:
1) Yes the test cell may be hidden for you but you will have to rely on your students to also have this extension installed and activated, otherwise it won't be hidden for them.
2) The students in any case could simply open up the browsers inspector and change this `display` property manually to reveal the test cell.

nadimmajed:
I will try it, thank you very much for your kind cooperation.

Szepi:
@nadimmajed : Precompute a hash of the desired answer, and then check the hash of the student's answer against the precomputed answer in the nonhidden part. The test will be performed in the student's notebook, yet, the student's cannot easily get the answer (unless the range of possible responses is limited).

PR #695, opened 2017-02-06
Use sys.executable when executing nbgrader

jhamrick:
Just in case it is somehow finding another version of nbgrader...


lgpage:

Issue #509, opened 2016-04-30
Investigate why subprocess don't always get killed


jhamrick:
Occassionally, when the tests fail, the subprocesses they have started (such as JupyterHub, the formgrader, or the notebook server) don't exit. I am not sure why this would be. This is definitely an issue that needs to get resolved at some point (it is not good to clutter up peoples machines with abandoned processes!) though it is not super urgent. I think it is going to take a bit of sleuthing to figure it out as it only happens sometimes; I would love if someone who knows more about process management than me could look into it...

jhamrick:
I haven't seen this occur in quite a while, so I am going to close this issue.

PR #698, opened 2017-02-06
Force phantomjs service to terminate in Linux

lgpage:
See https://github.com/SeleniumHQ/selenium/issues/767

This gave me grief while messing around with ``nbextension`` tests, managed to freeze up my PC after running out of memory.


jhamrick:
Ah! Nice find. I wonder if this is the solution to #509 that I could never figure out.

PR #700, opened 2017-02-07
use check_call for extension installation in tests

### LINKED ISSUE #694: Investigate why nbextension tests keep failing
The nbextension tests seem to have a much higher rate of spurious failures recently. They aren't real failures, as restarting the tests seems to fix it, but it is highly annoying.

minrk:
rather than Popen, which just launches the program.

check_call adds:

- wait for it to finish
- raise error on nonzero exit code

should close #694, if slow extension enabling was indeed the culprit.


jhamrick:
Thanks!

Well, there are still spuriously failing tests, but this does seem to have enforced at least that the nbextensions are properly installed. I think the other failing tests are probably due to other errors, sigh. Progress, at least! 😄

jhamrick:

PR #701, opened 2017-02-07
Set proxy-type=none in phantomjs

jhamrick:
Hopefully this will make the windows tests less flaky?

cf #649 

cc @lgpage


jhamrick:

PR #704, opened 2017-02-08
Fix nbextension tests

### LINKED ISSUE #694: Investigate why nbextension tests keep failing

lgpage:
This sets the course element (`course_list_default`) text before loading assignments and splits the test wait into two part:
- wait for courses to be loaded, then
- wait for assignments to be loaded

Should fix #694 properly.

### DIFF
--- a/nbgrader/nbextensions/assignment_list/assignment_list.js
+++ b/nbgrader/nbextensions/assignment_list/assignment_list.js
@@ -118,6 +118,7 @@ define([
             this.default_course_element.text(course);
         }
         this.current_course = course;
+        this.default_course_element.text(this.current_course);
         var success = $.proxy(this.load_assignment_list_success, this);
         this.assignment_list.load_list(course, success);
     };
@@ -139,7 +140,6 @@ define([
             this.data = undefined;
         }
 
-        this.default_course_element.text(this.current_course);
         this.enable_list();
     };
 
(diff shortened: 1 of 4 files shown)


willingc:
@lgpage I'll let @jhamrick review this as she is better than me with frontend stuff. If we haven't said it enough recently, thank you so much for all of the contributions that you have made to nbgrader. :cake:

lgpage:
:smile:

jhamrick:
Thanks for figuring this out!

I guess it was the ability to switch courses that introduced this issue? It's weird that it really didn't seem to be a problem until recently. Maybe Travis has slowed down further or something since then.

jhamrick:
And yes, 👍  to what @willingc said 😄

lgpage:
I think is was the combination of the switch courses and the additional test for missing / renamed notebooks. It takes about 5-6s on my machine for the course dropdown to be clickable.

willingc:
Happy to merge. Thanks @jhamrick for the review.

jhamrick:
Ah, got it.

Yes, it is also slow for me. The reason is because it actually has to shell out to the command line apps rather than running them in the same python session, and the reason it can't run them in the same python session is because the `NotebookApp` and `NbGrader` both inherit from `JupyterApp` which is a singleton, so it raises an error if you try to instantiate `NbGrader` within the `NotebookApp`. Fixing this would require a fair bit of refactoring, but maybe it's worthwhile to open an issue about it...

jhamrick:
See #705

jhamrick:

PR #707, opened 2017-02-08
Force rerun nbgrader commands

jhamrick:
This should fix test failures like this, where it fails the first time and then the rerun fails because the nbgrader command has already been run: https://travis-ci.org/jupyter/nbgrader/jobs/199697338


jhamrick:

Issue #417, opened 2015-09-21
Preserve kernel information


jhamrick:
For example, if the notebook is expected to be python 3, but students turn it in as python 2, the kernel should be updated in the notebook metadata so that it is python 3 (and presumably also this notebook shouldn't validate properly unless students are using the right kernel).

cc @joschu

jhamrick:
I am actually not sure if there is a super good way to do this, because kernels can have different names. Python kernels, regardless of whether they are Python 2 or Python 3, have language "python" but have different display names. So you can't tell based on the language name alone, but I also don't want to enforce that the display name is the same because this might be different for different people (e.g. "IPython 3" vs. "Python 3").

For the time being, I am going to close this issue, but if a good solution to this comes up in the future then I may reopen it.

lgpage:
I would like to re-visit this issue. I have the following case that is annoying, grading with a kernel name `python3` but some student submissions are failing because their kernel name is `Python [root]`... stupid Anaconda

jhamrick:
Perhaps one way to do it would be to store the kernel information in the database, and then rewrite the kernel name during autograding (i.e. when the cell contents are replaced)?

lgpage:
That was my thinking.

lgpage:
xref:
- https://github.com/jupyter/nbformat/issues/81
- https://github.com/jupyter/nbconvert/issues/515

lgpage:
closed by #714

PR #714, opened 2017-02-18
Preserve kernelspec when autograding

lgpage:
@jhamrick what do you think about this for #417

### DIFF
--- a/nbgrader/api.py
+++ b/nbgrader/api.py
@@ -111,11 +111,8 @@ class Notebook(Base):
     #: Unique id of :attr:`~nbgrader.api.Notebook.assignment`
     assignment_id = Column(String(32), ForeignKey('assignment.id'))
 
-    #: The :class:`~nbgrader.api.Kernelspec` object for this notebook
-    kernelspec = None
-
-    #: Unique id of :attr:`~nbgrader.api.Notebook.kernelspec`
-    kernelspec_id = Column(String(32), ForeignKey('kernelspec.id'))
+    #: The json string representation of the kernelspec for this notebook
+    kernelspec = Column(String(128), nullable=True)
 
     #: A collection of grade cells contained within this notebook, represented
     #: by :class:`~nbgrader.api.GradeCell` objects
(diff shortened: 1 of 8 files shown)


jhamrick:
That seems reasonable. The one thing I'm wondering though is if we want to have an actual `Kernelspec` object in the database... because if the kernelspec format in the notebook changes, it means we have to change our objects in the database too, which I could see being a hassle. Maybe a simpler solution that doesn't require us to care about what the kernelspec actually is would be to just dump the kernelspec information to a JSON string and use that? 

One other thing, just to make sure we're updating the kernelspecs appropriately, would be when you add tests to include a test that runs the notebook json against the nbformat validator: https://github.com/jupyter/nbformat/blob/master/nbformat/validator.py

lgpage:
>The one thing I'm wondering though is if we want to have an actual Kernelspec object in the database... because if the kernelspec format in the notebook changes, it means we have to change our objects in the database too, which I could see being a hassle. Maybe a simpler solution that doesn't require us to care about what the kernelspec actually is would be to just dump the kernelspec information to a JSON string and use that?

I was debating the same thing, but opted for the database approach as a first stab. I have updated this to store just a json string for the notebook.

lgpage:
>One other thing, just to make sure we're updating the kernelspecs appropriately, would be when you add tests to include a test that runs the notebook json against the nbformat validator: https://github.com/jupyter/nbformat/blob/master/nbformat/validator.py

I think I used this correctly, if you wouldn't mind checking.

nbgrader/tests/preprocessors/test_overwritekernelspec.py
```diff
@@ -0,0 +1,58 @@
...
+        nb = new_notebook()
+        nb.metadata['kernelspec'] = kernelspec
+        nb, resources = preprocessors[0].preprocess(nb, resources)
+
+        nb.metadata['kernelspec'] = {}
+        nb, resources = preprocessors[1].preprocess(nb, resources)
+
+        validate(nb)
+        notebook = gradebook.find_notebook("test", "ps0")
+        assert json.loads(notebook.kernelspec) == kernelspec
```
jhamrick: I would probably add one more assert here, which is just that the kernelspec in the notebook document itself is correct:

```
assert nb.metadata['kernelspec'] == kernelspec
```

nbgrader/api.py
```diff
@@ -111,6 +111,9 @@ class Notebook(Base):
     #: Unique id of :attr:`~nbgrader.api.Notebook.assignment`
     assignment_id = Column(String(32), ForeignKey('assignment.id'))
 
+    #: The json string representation of the kernelspec for this notebook
+    kernelspec = Column(String(128), nullable=True)
```
jhamrick: I wonder if we potentially want to make this really long (e.g. 1024 characters) just in case the kernelspec is ever really long for some reason.

jhamrick:
> I think I used this correctly, if you wouldn't mind checking.

Yes, looks good to me.

jhamrick:
👍 good to merge once the tests pass!

PR #731, opened 2017-03-28
Reset late submission penalty before checking if submission is late or not

### LINKED ISSUE #723: Late submission penalty
There is a bug with not resetting the late submission penalty if the due-date or timestamp changed such that the submission is no longer late.

To re-create (using the "zero" late submission plugin):
- Set the due-date and timestamp such that the submission is late
- Autograde
- Change the due-date or timestamp such that the submission is no longer late
- Autograde again (--force)
- Feedback will still show a late submission penalty

I have a fix for this, but this issue serves as a reminder to push it through.

lgpage:
Closes #723


lgpage:

Issue #734, opened 2017-03-30
Adding DataTables to formgrader


lgpage:
I'm going to throw this idea out there to get your thoughts. This started with me looking for a way to sort the formgrader data according to column header and I came across [DataTables](https://datatables.net/).

## Out of the box

Adding DataTables to the existing tables gives this, for example:

![selection_037](https://cloud.githubusercontent.com/assets/1577821/24525741/c5e089b8-159b-11e7-956b-7e2e5df9361a.png)

## With very few config options

I'm not a big fan of paging, so 2 config options to disable it gives this, for example:

![selection_038](https://cloud.githubusercontent.com/assets/1577821/24525773/e951fbf2-159b-11e7-9547-c78b88f5fd58.png)

This works just about perfectly for sorting and searching data with just a few issues:
- Not all columns sort correctly - scores, for example, are interpreted as strings, so ``9 / 20`` is sorted after ``11 / 20``. This can be fixed by adding a ``data-order`` html attribute to each table cell.
- Delayed response - for larger sets of data there is a noticeable delay on loading, sorting, and searching. This could be improved by allowing DataTables to [source data through ajax](https://datatables.net/examples/ajax/orthogonal-data.html) and include the [scoller](https://datatables.net/extensions/scroller/) extension.

jhamrick:
That sounds like a great idea to me!

> Not all columns sort correctly - scores, for example, are interpreted as strings, so 9 / 20 is sorted after 11 / 20. This can be fixed by adding a data-order html attribute to each table cell.

That's definitely something I would expect to work correctly, but as long as we can fix it by doing the `data-order` attribute then that's fine with me.

> Delayed response - for larger sets of data there is a noticeable delay on loading, sorting, and searching. This could be improved by allowing DataTables to source data through ajax and include the scoller extension.

Sourcing the data through ajax is actually something I've had in the back of my mind as a refactor of how the gradebook interface works, so I would definitely be on board with refactoring it in that way to make the lag less noticeable (though I would suggest those be two separate PRs).

jhamrick:
Since this is technically addressed with #735 I am going to close this issue -- but we still have #737 open for sourcing the data from the tables via ajax.

PR #735, opened 2017-04-01
Add DataTables functionality to existing formgrade tables

lgpage:
Add DataTables functionality to existing formgrader tables as part of the discussion in #734


jhamrick:
Looks great! My only question would be, is it possible to install these javascript files using bower, as in the docs here? http://nbgrader.readthedocs.io/en/stable/contributor_guide/js_dependencies.html#adding-or-updating-javascript-libraries

lgpage:
Do I still commit the main component files?

jhamrick:
Yes, and ignore other files -- but looks like you figured that out. This looks great, thanks!

PR #737, opened 2017-04-08
[WIP] Source dataTables data via Ajax

lgpage:
Source dataTables data via Ajax as part of the discussion in #734


jhamrick:
Just as a heads up, I think as I am implementing the solution to #555, there are going to be very massive conflicts with this, as I am doing a redesign of the formgrader interface itself. Sorry about that... 😞

lgpage:
No worries. I will wait for #555 to be completed before I look at this again. Depending on what is easier I may just close this and start a new.

jhamrick:
Cool, thanks for being flexible!

On the bright side, the new implementation of the formgrader I think partially helps with this functionality (as it now sources the data through ajax), so hopefully it shouldn't be too complicated to get the streaming/scrolling functionality on top of that.

PR #746, opened 2017-05-05
Fix bug with --to in custom exporters

### LINKED ISSUE #745: self.to not set in custom exporter
I have a custom exporter class

```python
from nbgrader.plugins.export import ExportPlugin
 
class ToTester(ExportPlugin):
    def export(self, gradebook):
        self.log.info(f'Value of self.to is {self.to}')
```

Invoking it produces the output

```shell
$ nbgrader export --to foo.csv --exporter=myexporter.ToTester
[ExportApp | INFO] Using exporter: ToTester
[ExportApp | INFO] Value of self.to is
```

It is clear that the value of `self.to` is not being set for some reason. The strange part is `self.to` *is* being set for `CsvExportPlugin`

```shell
$ nbgrader export --to foo.csv                               
[ExportApp | INFO] Using exporter: CsvExportPlugin
[ExportApp | INFO] Exporting grades to foo.csv
```

Any idea why `self.to` is getting set in `CsvExportPlugin` and not `myexporter.ToTester`?

jhamrick:
Fixes #745

### DIFF
--- a/nbgrader/apps/exportapp.py
+++ b/nbgrader/apps/exportapp.py
@@ -54,7 +54,7 @@ class ExportApp(NbGrader):
     def _classes_default(self):
         classes = super(ExportApp, self)._classes_default()
         classes.append(ExportApp)
-        classes.append(self.plugin_class)
+        classes.append(ExportPlugin)
         return classes
 
     def start(self):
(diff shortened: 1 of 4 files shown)


lgpage:
@jhamrick is this change not also needed for the ``zip_collect`` app?

jhamrick:
Good point, it is probably also needed there, yeah. I'll double check.

PR #772, opened 2017-06-04
Standalone versions of nbgrader assign, autograde, and feedback

### LINKED ISSUE #763: Convert all nbgrader subcommands to standalone objects
If we want to run `nbgrader assign`, `nbgrader autograde`, and `nbgrader feedback` from the notebook web app (see #555), we will also have to create standalone objects that can perform these operations without inheriting from `JupyterApp` (as was done with `nbgrader release`, `fetch`, `collect`, etc.) or shell out to the command line to run these commands. The former approach is more work, but in the long run much better I think. The reason why shelling out would be necessary is that both the notebook and nbgrader inherit from `JupyterApp`, which is a singleton, and thus both can't be instantiated at the same time.

jhamrick:
Fixes #763

This creates standalone objects that can be used for assigning, autograding, and feedback. For example:

```python
from nbgrader.coursedir import CourseDirectory
from nbgrader.converters import Assign
from traitlets.config import Config

import logging
logging.basicConfig(level="INFO")

c = Config()
c.CourseDirectory.assignment_id = "ps1"
c.Assign.create_assignment = True
c.Assign.force = True

coursedir = CourseDirectory(config=c)
converter = Assign(coursedir=coursedir, config=c)
converter.start()
```

which will produce the expected `nbgrader assign` output:

```
WARNING:traitlets:Removing existing assignment: /private/tmp/course101/release/ps1
INFO:traitlets:Copying /private/tmp/course101/source/./ps1/jupyter.png -> /private/tmp/course101/release/./ps1/jupyter.png
INFO:traitlets:Updating/creating assignment 'ps1': {}
INFO:traitlets:Converting notebook /private/tmp/course101/source/./ps1/problem1.ipynb
INFO:traitlets:Writing 7167 bytes to /private/tmp/course101/release/./ps1/problem1.ipynb
INFO:traitlets:Converting notebook /private/tmp/course101/source/./ps1/problem2.ipynb
INFO:traitlets:Writing 1541 bytes to /private/tmp/course101/release/./ps1/problem2.ipynb
INFO:traitlets:Setting destination file permissions to 644
```

This is a first step in getting #555 working.


willingc:
Fantastic!

PR #779, opened 2017-06-29
Update the year!

jilljenn:


nbgrader/docs/source/conf.py
```diff
@@ -55,7 +55,7 @@
 
 # General information about the project.
 project = 'nbgrader'
-copyright = '2015, Jupyter Development Team'
+copyright = '2015–2017, Jupyter Development Team'
```
jhamrick: I think this is a n-dash, which is unicode, which causes the tests to fail on Python 2. Can you change it to be just a single hyphen?

Otherwise looks great, thanks for the contribution!

jhamrick:
Will go ahead and merge this and fix it in the upcoming release PR.

PR #781, opened 2017-07-05
Assign missing notebooks a score of zero and mark as not needing grading

jhamrick:
Currently, missing notebooks get an implicit score of zero (the default) and are marked as needing manual grading. However, it is impossible to manually grade a missing notebook. This PR detects notebooks that are missing and assigns them an explicit autograded score of 0 and also marks them as not needing manual grading.

### DIFF
--- a/nbgrader/converters/autograde.py
+++ b/nbgrader/converters/autograde.py
@@ -139,6 +139,24 @@ class Autograde(BaseConverter):
             self.log.error(msg)
             raise NbGraderException(msg)
 
+        # check for missing notebooks and give them a score of zero if they
+        # do not exist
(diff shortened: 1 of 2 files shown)


willingc:
Code looks good. Question: How would an instructor determine that the assignment was not turned in?

jhamrick:
In the student view of the formgrader, if you navigate to a particular student it will indicate which assignments are missing:

image

And in the new formgrader, if there are individual files missing, those will show up there, too:

image

It probably would be good to make it easier for instructors to query which students did not submit an assignment, though. I'll open an issue for that. Thanks!

willingc:
Sounds good.

PR #834, opened 2017-07-15
Add documentation for course_id and release

### LINKED ISSUE #812: Document that course_id needs to be set for release
The release button in the formgrader won't show up unless the course_id is set. This should be documented somewhere and perhaps even print out a warning on the command line.

Ruin0x11:
Fixes #812.

### DIFF
--- a/nbgrader/apps/fetchapp.py
+++ b/nbgrader/apps/fetchapp.py
@@ -73,6 +73,7 @@ class FetchApp(NbGrader):
     def start(self):
         super(FetchApp, self).start()
 
+        print("{}".format(self.extra_args))
         # set assignemnt and course
         if len(self.extra_args) == 1:
             self.coursedir.assignment_id = self.extra_args[0]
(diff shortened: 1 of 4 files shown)


nbgrader/apps/fetchapp.py
```diff
@@ -73,6 +73,7 @@ def _load_config(self, cfg, **kwargs):
     def start(self):
         super(FetchApp, self).start()
 
+        print("{}".format(self.extra_args))
```
jhamrick: I think this is an extra print statement that can now be removed.

nbgrader/docs/source/user_guide/managing_assignment_files.ipynb
```diff
@@ -1008,9 +1013,21 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python",
+   "display_name": "Python 2",
```
jhamrick: I like to avoid committing kernel information to the repository since it changes every time someone edits the notebook. To clear it, can you run:

```
invoke docs
```

from the root of the repository (this will take a little while) and then commit the results?

PR #838, opened 2017-07-15
Fetch multiple assignments in one command

### LINKED ISSUE #435: Add the ability to fetch multiple assignment with one command
As suggested by @mikebolt in #433, adding the ability to fetch multiple assignments with one command would be a nice future enhancement.

> It would also be nice to be able to fetch multiple assignments with one command:
> 
> nbgrader fetch --course data301a exampleA exampleB
> This isn't supported right now. Perhaps specifying --assignment exampleC would just add to the set of >assignments to fetch, and you could also use commas with the assignment flag like this:
> 
> nbgrader fetch --assignment=exampleA,exampleB,exampleC --course data301a

Ruin0x11:
Fixes #435.

### DIFF
--- a/nbgrader/apps/fetchapp.py
+++ b/nbgrader/apps/fetchapp.py
@@ -73,16 +73,19 @@ class FetchApp(NbGrader):
     def start(self):
         super(FetchApp, self).start()
 
-        # set assignemnt and course
-        if len(self.extra_args) == 1:
-            self.coursedir.assignment_id = self.extra_args[0]
-        elif len(self.extra_args) > 2:
-            self.fail("Too many arguments")
-        elif self.coursedir.assignment_id == "":
+        # set assignment and course
+        if len(self.extra_args) == 0:
             self.fail("Must provide assignment name:\nnbgrader <command> ASSIGNMENT [ --course COURSE ]")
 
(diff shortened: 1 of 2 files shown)


nbgrader/apps/fetchapp.py
```diff
@@ -73,16 +73,19 @@ def _load_config(self, cfg, **kwargs):
...
         super(FetchApp, self).start()
 
-        # set assignemnt and course
-        if len(self.extra_args) == 1:
-            self.coursedir.assignment_id = self.extra_args[0]
-        elif len(self.extra_args) > 2:
-            self.fail("Too many arguments")
-        elif self.coursedir.assignment_id == "":
+        # set assignment and course
+        if len(self.extra_args) == 0:
```
jhamrick: This is causing the tests to fail, because now if `--assignment` is passed (and no positional arguments), it won't work. So we need to check for `if len(self.extra_args) == 0 and self.coursedir.assignment_id == ""` and handle the case where the assignment id is set.

jhamrick:
I am not sure why that docs test is failing all of a sudden, but it seems to be unrelated. I'll merge this and make a new PR to fix that if necessary.

PR #871, opened 2017-08-20
Fix NbGraderAPI.timezone handling

thotypous:
**Before this patch**

Currently, the user may specify, e.g.

```
c = get_config()
c.NbGraderAPI.timezone = 'America/Sao_Paulo' 
```

Then it would correctly convert all timestamps to `'America/Sao_Paulo'`. However, it would also be displayed in the formgrader when the user sets the assignment's due date:

Timezone displayed in the formgrader before this patch

When the user sends this form, the timezone string is concatenated to the time and passed to `utils.parse_utc`, which cannot handle this string format.

**After this patch**

This patch converts `NbGraderAPI.timezone` to a format readable by `parse_utc` before displaying it to the user in the formgrader. This way the user does not need to change the default value presented in the Timezone field: it will just use the configured timezone out-of-the-box.

Timezone displayed in the formgrader after this patch

**Note**

The `dateutil.tz.gettz` function does not support reading a timezone in the numeric format (e.g. `-0300` or `UTC-3`), therefore it is not possible to use the numeric format to configure `NbGraderAPI.timezone`.

In other words, this patch is the only way to get the timezones working correctly in the Web UI. It also takes into account that the timezone offset might change during the course due to daylight savings.

### DIFF
--- a/nbgrader/apps/api.py
+++ b/nbgrader/apps/api.py
@@ -12,7 +12,7 @@ from ..coursedir import CourseDirectory
 from ..converters import Assign, Autograde
 from ..exchange import ExchangeList, ExchangeRelease, ExchangeCollect, ExchangeError
 from ..api import MissingEntry, Gradebook, Student, SubmittedAssignment
-from ..utils import parse_utc, temp_attrs, capture_log, as_timezone
+from ..utils import parse_utc, temp_attrs, capture_log, as_timezone, to_numeric_tz
 
 
 class NbGraderAPI(LoggingConfigurable):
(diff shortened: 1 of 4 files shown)


jhamrick:
Just noticed you updated the PR---for future reference (and for other projects), I don't get an email about it, so it is helpful to leave a comment or something so the maintainers know to review the PR again 😄 

In any case, looks good, thanks!

thotypous:
Thanks @jhamrick ! Sorry for not commenting here, I was still looking at writing the tests you recommended, but then I got a little overwhelmed with other stuff. If I get the chance to work on it I will open another PR. Thanks again for the review and for the merge :)

Issue #875, opened 2017-08-26
License


aliandra:
Rename COPYING.md to LICENSE and update license.

jhamrick:
Fixed by #876 and #1094.

PR #876, opened 2017-08-26
Revise and rename copying.md to license

aliandra:
Rename COPYING.md to LICENSE and update license. Closes issue #875


jzf2101:
@jhamrick can you confirm that you're ok with the license standardization from the sprint?

lgpage:
@aliandra I am not sure what you and @jhamrick discussed regarding the license during the sprint.
My suggestion is to standardize the license to the Jupyter governance license for projects (https://github.com/jupyter/governance/blob/master/projectlicense.md).

Pinging @ellisonbg for his thoughts.

jzf2101:
@lgpage I'm ok with having a license looser than BSD-3.  Based on my conversations with @ellisonbg on Jupyterlab, see https://github.com/jupyterlab/jupyterlab-demo/pull/31 , the version of the license proposed by @aliandra is our new standard.  We should update the governance repo to reflect those cahnges.

LICENSE
```diff
@@ -1,8 +1,6 @@
-nbgrader is licensed under the terms of the Modified BSD License
-(also known as New or Revised or 3-Clause BSD), as follows:
-
-Copyright (c) 2014, Jupyter Development Team
+BSD 3-Clause License
 
+Copyright (c) 2017, Project Jupyter Contributors
```
Carreau: Before merging, we also need to change date back to 2014. The date should be the year of first copyright, not last. 2014  is implicitly 2014-present.

lgpage:
Adding these discussions for reference:
- https://github.com/jupyter/governance/issues/37
- https://github.com/jupyterlab/jupyterlab/pull/543
- https://groups.google.com/forum/#!topic/jupyter/vZpWgw8zKdc

jasongrout:
> the version of the license proposed by @aliandra is our new standard

To be clear, the "new standard" is still under discussion: jupyter/governance#37

lgpage:
@jasongrout yup that was my understanding from the jupyter/governance#37 discussion.
I'll leave this PR open until @jhamrick is back and consensus is reached.

jhamrick:
[Catching up on issues and PRs as I've been away for the last two months]

@aliandra thank you for contributing! And thank you for being patient with us while we sort out some project-wide issues regarding licensing.

It sounds like based on jupyter/governance#37 that there are multiple opinions regarding how we should do the copyright year, so I think it is best to leave this open until jupyter/governance#37 has been resolved so we are consistent with the other Jupyter projects. Otherwise though I think this PR looks good!

jhamrick:
Thanks for the contribution @aliandra, and sorry this took so long to resolve! I went ahead and followed the convention used in some of the other Jupyter projects and made #1094 which is basically the same as this except that it reverts the year back to 2014.

PR #880, opened 2017-08-28
change directory when validating notebooks

### LINKED ISSUE #879: notebooks should be validated at their directories
It seems that notebooks are validated with `$HOME` being the working directory. Any tests which require reading a relative file, e.g., opening a dataset, will be failed by this behavior.

I have a simple notebook with one Autograder test

```python
import os
raise Exception(os.getcwd())
```
screen shot 2017-08-28 at 5 53 04 pm

It throws the current working directory which is `/home/randy3k/practice`

However, when I click the validate button, I got

screen shot 2017-08-28 at 5 54 11 pm

It indicates that the notebook is validated at the home directory but not in the directory where the notebook is located.

randy3k:
fix #879 

I would like to add some tests, however, it is totally out of my ability. Have no idea how `invoke` works.

### DIFF
--- a/nbgrader/server_extensions/validate_assignment/handlers.py
+++ b/nbgrader/server_extensions/validate_assignment/handlers.py
@@ -2,7 +2,6 @@
 
 import os
 import json
-import contextlib
 import traceback
 
 from tornado import web
@@ -19,14 +18,6 @@ from ... import __version__ as nbgrader_version
 
 
 static = os.path.join(os.path.dirname(__file__), 'static')
-
-
-@contextlib.contextmanager
-def chdir(dirname):
-    currdir = os.getcwd()
-    os.chdir(dirname)
-    yield
-    os.chdir(currdir)
 
 
 class ValidateAssignmentHandler(IPythonHandler):
--- a/nbgrader/utils.py
+++ b/nbgrader/utils.py
@@ -8,6 +8,7 @@ import shutil
 import stat
 import logging
 import traceback
+import contextlib
 
 from setuptools.archive_util import unpack_archive
 from setuptools.archive_util import unpack_tarfile
@@ -224,6 +225,17 @@ def full_split(path):
         return (rest,)
     else:
         return full_split(rest) + (last,)
+
+
+@contextlib.contextmanager
+def chdir(dirname):
+    currdir = os.getcwd()
+    if dirname:
+        os.chdir(dirname)
+    try:
+        yield
+    finally:
+        os.chdir(currdir)
 
 
 def rmtree(path):
--- a/nbgrader/tests/nbextensions/conftest.py
+++ b/nbgrader/tests/nbextensions/conftest.py
@@ -170,6 +170,9 @@ def _make_browser(tempdir):
     for filename in glob.glob(os.path.join(os.path.dirname(__file__), "files", "*.ipynb")):
         shutil.copy(filename, os.path.join(tempdir, os.path.basename(filename)))
 
+    for filename in glob.glob(os.path.join(os.path.dirname(__file__), "files", "*.txt")):
+        shutil.copy(filename, os.path.join(tempdir, os.path.basename(filename)))
+
     selenium_logger = logging.getLogger('selenium.webdriver.remote.remote_connection')
     selenium_logger.setLevel(logging.WARNING)
 
--- a/nbgrader/tests/nbextensions/files/data.txt
+++ b/nbgrader/tests/nbextensions/files/data.txt
@@ -1,0 +1,4 @@
+line 1
+line 2
+line 3
+line 4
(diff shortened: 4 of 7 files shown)


randy3k:
Not sure why it failed the tests.

lgpage:
A test for this can be added by:
- creating a new test notebook and dummy local file (say `data.txt`) in https://github.com/jupyter/nbgrader/tree/master/nbgrader/tests/preprocessors/files (the test notebook should load `data.txt` and pass tests when run)
- adding a test to https://github.com/jupyter/nbgrader/blob/master/nbgrader/tests/test_validator.py that uses this new notebook (which of course loads a local file) to test it succeeds.

randy3k:
I have updated the PR. Thanks for the direction of writing a test, I will take a look later today.

misolietavec:
This does not fix anything, in my case. I have several nbgrader/jupyterhub installations, never got "Validate" button in notebooks working. Teachers and users have notebooks not in home directory, but in $HOME/notebooks. 

For example, as a teacher I have this course root directory:
`/home/mike/notebooks/gvpt_martin`. 
and the assignment subdirectory `ps1` with notebook `problems1.ipynb` which I like to validate.
Trying "Validate" button (after @randy3k fix), this is what I get as part of traceback:
`File "/usr/local/lib/python3.5/dist-packages/nbgrader/utils.py", line 227, in chdir
    os.chdir(dirname)
FileNotFoundError: [Errno 2] No such file or directory: 'gvpt_martin/source/ps1'`
Of course, because this is relative path. Relative to what? I am sure, nbgrader developers
can fix that.  

Making no assumptions about the current working directory, I made for myself such replacement by absolute path in `validator.py` (about line 228):
`dirname =  os.path.join(os.environ['HOME'],'notebooks',os.path.dirname(filename))`
"Works for me" :) but in general, this is not the right path to go...
@randy3k : you not only moved the chdir function to nbgrader/utils.py, but you also changed it. I am not sure if that's for better :-)
@randy3k: I am also not convinced that (temporarily) changing directory is better solution than proper fixes to file paths.

randy3k:
@misolietavec We need to change directory anyway for reading relative files. Your issue is casued by a different bug. It was because `filename` is relative the the notebook directory but the working direction is different from `filename`. For instance `os.path.abspath(filename)` will not be able to find the files. We may need `NotebookApp.notebook_dir`.

nbgrader/server_extensions/validate_assignment/handlers.py
```diff
@@ -41,7 +41,8 @@ def validate_notebook(self, path):
         try:
             config = self.load_config()
             validator = Validator(config=config)
-            result = validator.validate(path)
+            notebook_dir = self.session_manager.parent.notebook_dir
```
randy3k: not sure what the best way to get `notebook_dir`.

randy3k:
have no idea why the last commit failed the tests...

lgpage:
@randy3k I fixed the kernelspec failures in #888

randy3k:
Thanks, I will rebase this. Btw, any suggestions on accessing `notebook_dir`? I am using session manager now but it seems there is better way.

lgpage:
@randy3k I'll have a closer look at this over the weekend and hopefully have a suggestion for you.

nbgrader/utils.py
```diff
@@ -226,6 +227,17 @@ def full_split(path):
         return full_split(rest) + (last,)
 
 
+@contextlib.contextmanager
+def chdir(dirname):
+    currdir = os.getcwd()
+    if dirname:
```
lgpage: Please can you remove this ``if`` statement. Let ``chdir`` handle `dirname` not being set properly.

lgpage:
@randy3k thanks. This looks good, just the one change and then I'll merge this.
@misolietavec I think your issue is unrelated to this, if you still have problems with the validation (after this PR is merged) please can you open a new issue so we can sort it out.

nbgrader/utils.py
randy3k: If we remove the if statement, we will need to do this later. Is that what you prefer?

```
        if dirname:
            with utils.chdir(dirname):
                nb = self._preprocess(basename)
        else:
            nb = self._preprocess(basename)
```

lgpage: Ah I see now why this is needed, never mind, stick to how it is now :)

misolietavec:
@lgpage I wrote above comment **after** making local changes from that PR. So, merging this PR does not resolve my issues. It can be seen as a proof of independency on that PR. I will open new issue, but before will try to think about the solution.

jhamrick:
[Catching up on issues and PRs as I've been away for the last two months]

Thanks very much for finding this bug and contributing the fix, @randy3k ! And thanks for reviewing @lgpage !

PR #892, opened 2017-09-17
Format code blocks in installation instructions

rwest:
Very small tweak. Two code blocks had single colon: not a double:: so were not rendering as a code blocks.


jhamrick:
Thanks!

PR #895, opened 2017-10-18
Add test and fix for parsing csv key names with spaces

### LINKED ISSUE #890: Not importing emails when Importing students from .csv
### Operating system
`Ubuntu 16.04`

### `nbgrader --version`

`nbgrader version 0.5.2`

### Expected behavior

I'm creating the students database from a `csv` file.  My csv files has:
`id,last_name,first_name,email` so I expect when I do `nbgrader db student list` to have all the fields
for each student.

### Actual behavior

For some reason the field corresponding to email it's not being import. For example in the csv file I have:
`id,last_name,first_name,email`
`stud1, smith, blabla, bla@mail.edu`

and when creating the database I got:

`Creating/updating student with ID 'stud1': {'last_name': 'smith', 'first_name': 'blabla'}`

when listing students I see: 

`stud1 (smith, blabla) -- None`

Any clue what is going on?

mpacer:
creates private method for preprocessing keys for any future transforms
that are needed

Closes #890.

### DIFF
--- a/nbgrader/apps/dbapp.py
+++ b/nbgrader/apps/dbapp.py
@@ -135,7 +135,11 @@ class DbGenericImportApp(NbGrader):
 
     @property
     def primary_key(self):
-        return [k.key for k in self.table_class.__mapper__.primary_key][0]
+        possible_keys = [k.key for k in self.table_class.__mapper__.primary_key]
+        if len(possible_keys)>1:
+            raise TypeError("%s has more than one primary_key" % self.table_class.__name__)
+        else:
+            return possible_keys[0]
 
     def start(self):
         super(DbGenericImportApp, self).start()
--- a/nbgrader/apps/dbapp.py
+++ b/nbgrader/apps/dbapp.py
@@ -165,8 +165,17 @@ class DbStudentImportApp(NbGrader):
         return Student.__table__.c.keys()
 
     def _preprocess_keys(self, keys):
-        new_keys = [key.strip() for key in keys]
-        return new_keys
+        """
+        Helper function for preprocessing keys
+        """
+        proposed_keys = [key.strip() for key in keys]
+        unknown_keys = [k for k in proposed_keys if k not in self.allowed_keys]
+        if unknown_keys:
+            self.log.info("Unknown keys in csv: '%s'",
+                          (', '.join(unknown_keys[:-1])
+                           + 'and '
+                           + unknown_keys[-1]))
+        return proposed_keys
 
 
 class DbStudentListApp(NbGrader):
--- a/nbgrader/apps/dbapp.py
+++ b/nbgrader/apps/dbapp.py
@@ -118,7 +118,13 @@ class DbStudentRemoveApp(NbGrader):
 
 class DbGenericImportApp(NbGrader):
 
-    def get_db_update_method(self, gb):
+    def db_update_method_name(self):
+        """
+        Name of the update method used on the Gradebook for this import app.
+        It is expected to have the signature:
+            * instance_id : string, identifies which instance you are updating based on self.primary_key
+            * instance : dictionary, contents for the update from the parsed csv rows; unpacked as kwargs
+        """
         raise NotImplementedError
 
     @property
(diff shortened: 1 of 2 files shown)


nbgrader/tests/apps/test_nbgrader_db.py
```diff
@@ -176,6 +176,61 @@ def test_student_import(self, db, temp_cwd):
...
+            student = gb.find_student("foo")
+            assert student.last_name == "xyz"
+            assert student.first_name == "abc"
+            assert student.email == "foo@bar.com"
+            student = gb.find_student("bar")
+            assert student.last_name is None
+            assert student.first_name is None
+            assert student.email is None
+
+        # check that it fails when no id column is given
```
jhamrick: Is anything after this line different from what's in `test_student_import`? If not I think you can delete it.

nbgrader/apps/dbapp.py
```diff
@@ -140,6 +140,7 @@ def start(self):
         with Gradebook(self.coursedir.db_url) as gb:
             with open(path, 'r') as fh:
                 reader = csv.DictReader(fh)
+                reader.fieldnames = self._preprocess_keys(reader.fieldnames)
```
jhamrick: I just realized we need to do this also in `DbAssignmentImportApp`, as it suffers from the same issue---that trailing whitespace will cause it to not recognize keys. Can you add this same functionality there too?

mpacer: done!

mpacer:
Also, my tests pass on both py2 and py3, not sure why they would be failing on Appveyor. The errors also seem unreleated to what I touched.

mpacer:
So much code sharing led me to think this should be genericised.

The method for defining the gradebook update method seemed a little hacky, but I couldn't think of any other way to do pass through the gradebook instance from within the context manager.

nbgrader/apps/dbapp.py
```diff
@@ -116,48 +116,106 @@ def start(self):
...
+    @property
+    def description(self):
+        raise NotImplementedError
+
+    @property
+    def table_class(self):
+        raise NotImplementedError
+
+    @property
+    def primary_key(self):
```
jhamrick: Since this is only really being used for the student table, it seems like it would be easier just to set `primary_key = 'id'` (just like in the assignment class) rather than detecting it automatically.

nbgrader/apps/dbapp.py
```diff
@@ -116,48 +116,106 @@ def start(self):
...
+                            instance[key] = val
+                    instance_primary_key = instance.pop(self.primary_key)
+
+
+                    self.log.info("Creating/updating %s with %s '%s': %s",
+                                  self.table_class.__name__,
+                                  self.primary_key,
+                                  instance_primary_key,
+                                  instance)
+                    self.get_db_update_method(gb)(instance_primary_key, **instance)
```
jhamrick: Rather than having this be a function you could use `getattr` and define the function name as a attribute of the class. For example:

```
class DbStudentImportApp(DbGenericImportApp):
    ...
    db_update_method = "update_or_create_student"
```

and then here you'd do `getattr(gb, self.db_update_method)(instance_primary_key, **instance)`

mpacer: good idea, done though I didn't do it in one line as that is more difficult to read and this is already decently complex.

nbgrader/apps/dbapp.py
mpacer: ✓

jhamrick:
Unfortunately it does look like replacing the traitlets with properties causes issues when building the docs 😞

mpacer:
I can use traitlet's dynamic defaults instead :).

mpacer:
I actually don't think that name was ever a traitlet, or we wouldn't be able to run into this problem.

mpacer:
also… looking at the doc generation code, it needs to be a class attribute, not an instance attribute, so traitlets couldn't work.

```python
        name = cls.name.replace(" ", "-")
```

PR #897, opened 2017-10-19
Give more details on how to use formgrader and jupyterhub

### LINKED ISSUE #881: Using nbgrader and jupyterhub for multiple classes
I am under the impression that I need to edit the jupyterhub_config.py file in order to use nbgrader in multiple classes at once. I can't figure out how to do this. I see wordage in the jupyterhub_config.py itself, like:

List of service specification dictionaries.
  
  A service
  
  For instance::
  
      services = [
          {
              'name': 'cull_idle',
              'command': ['/path/to/cull_idle_servers.py'],
          },
          {
              'name': 'formgrader',
              'url': 'http://127.0.0.1:1234',
              'token': 'super-secret',
              'environment': 
          }
      ]

but my attempts to use the formgrader example have not worked out. 

Does anyone have an example for how to set this up for multiple courses? I'll ping @minrk since from what I've seen you are the one who might know how to do this. Thanks very much.

jhamrick:
Fixes #881 

I try to go into much more detail here about different ways you could set up the formgrader and jupyterhub. Hopefully this clears up some of the confusion about how to get it to work. I would appreciate if you have any feedback @kthyng @yao2001626 @dsblank @reckoner

### DIFF
--- a/nbgrader/docs/source/configuration/jupyterhub_config.rst
+++ b/nbgrader/docs/source/configuration/jupyterhub_config.rst
@@ -66,8 +66,10 @@ Example Use Case: One Class, Multiple Graders
 If you have multiple graders, then you can set up a `shared notebook server
 <https://github.com/jupyterhub/jupyterhub/tree/master/examples/service-notebook>`_
 as a JupyterHub service. I recommend creating a separate grader account (such
-as ``grader-course101``) for this server to have access to. Your JupyterHub
-config would then look something like this:
+as ``grader-course101``) for this server to have access to. Then, install and
+enable the formgrader and Create Assignment extensions for this grader account
+(see :doc:`/user_guide/installation`). Your JupyterHub config would then look
+something like this:
 
 .. code:: python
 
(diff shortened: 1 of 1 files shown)


jhamrick:
I am going to merge this for now so it's in the official docs, but if anybody finds anything confusing or has any suggestions I'm happy to make more changes in a future PR!

kthyng:
Hi sorry to be late — I have a backlog of email and just saw this. I like the extra information you added. I haven't tried out the instructions for running two classes since I'm already running two classes in parallel with two different users for this semester. But, I think I'll be teaching the class again next semester and I'll try it out before starting. I haven't used the services before so I'll let you know if I can figure it out.

jhamrick:
No problem, I was also very late in updating these instructions (and I know how busy it gets in the middle of the semester) so no worries. When you get around to testing things out for next semester I should be more available to help debug, too, so hopefully we can make sure everything gets set up cleanly and the docs are clear 😄

kthyng:
Ok I am working on this now. It looks like the set up you've suggested is similar to what I ended up doing last semester -- using two accounts on my workstation, one for each of two classes -- except I didn't use Jupyterhub services. Is there a reason to use the service set up in particular?

kthyng:
I have just set things up like I did last semester, without Jupyterhub services. Seems to work fine again!

PR #905, opened 2017-10-28
Update changelog and rebuild docs from 0.5.4

jhamrick:


nbgrader/docs/source/changelog.rst
```diff
@@ -8,6 +8,27 @@ A summary of changes to nbgrader.
 0.5.x
 -----
 
+0.5.4
+~~~~~
+
+nbgrader version 0.5.3 is a bugfix release, with the following PRs merged:
```
willingc: I think you want to bump the version ;-)

jhamrick: Oops! Good catch. This is a typo in the 0.5.4. release too unfortunately, but is minor enough that I this it's probably fine to wait for 0.6.0 for the typo to be fixed in the docs.

Issue #902, opened 2017-10-23
The header file is not åäö-friendly


consideRatio:
### Operating system
Windows 10

### `nbgrader --version`
0.5.3

### `jupyter --version`
4.3.0

### Expected behavior
Generation of student notebooks should concatenate the content of the header file with the notebook without any issues with special letters such as å ä and ö.

### Actual behavior
The student versions's header gets its å ä and ö characters misrepresented.

### Steps to reproduce the behavior
Create a header file with the characters å ä and ö and release a student version, look at the student version and conclude that the header does not look like it should in the released student version.

jhamrick:
Hmm, I am unfortunately not able to reproduce it. Could you send me copies of notebooks that lead to this bug happening? Additionally, could you give me the output of:

```
jupyter nbconvert --version
```

and

```
jupyter notebook --version
```

Thanks!

jhamrick:
Also, if you could send me a copy of the notebook that isn't formatted correctly, that would be helpful, too!

consideRatio:
@jhamrick thanks for looking into the issue!

I tried to reproduce it myself on a Ubuntu computer without success, then I used a windows 10 computer with exactly the same versions of jupyter, jupyter notebook, nbgrader, nbconvert, python and conda, and was able to reproduce it!

https://github.com/consideRatio/nbgrader-issues

![image](https://user-images.githubusercontent.com/3837114/32135552-6a2844a0-bc01-11e7-91a2-125c46d1bb4f.png)

jhamrick:
Hmm, this sounds like it might be a windows issue, then. I unfortunately don't currently have access to a windows machine to test this (and won't for probably at least another month---I just moved to the UK and my windows machine is still in a shipping container somewhere)... @lgpage would you be able to see if you can reproduce the bug on windows?

[...]

PR #906, opened 2017-10-28
Resolves issues with UTF-8

consideRatio:
Issue #902 is resolved by this PR, but this PR isn't very thorough yet, the fix of clarifying what encoding to use when reading a file on windows (10) might need to be applied at more places in the nbgrader project.

@jhamrick correctly guessed that this change would resolve the issue, I just tried it.

### DIFF
--- a/nbgrader/apps/assignapp.py
+++ b/nbgrader/apps/assignapp.py
@@ -1,3 +1,5 @@
+# coding: utf-8
+
 import sys
 
 from traitlets import default
--- a/nbgrader/apps/autogradeapp.py
+++ b/nbgrader/apps/autogradeapp.py
@@ -1,3 +1,5 @@
+# coding: utf-8
+
 import sys
 
 from traitlets import default
--- a/nbgrader/apps/baseapp.py
+++ b/nbgrader/apps/baseapp.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-# encoding: utf-8
+# coding: utf-8
 
 from __future__ import print_function
 
--- a/nbgrader/apps/collectapp.py
+++ b/nbgrader/apps/collectapp.py
@@ -1,3 +1,5 @@
+# coding: utf-8
+
 from traitlets import default
 
 from .baseapp import NbGrader, nbgrader_aliases, nbgrader_flags
--- a/nbgrader/apps/dbapp.py
+++ b/nbgrader/apps/dbapp.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-# encoding: utf-8
+# coding: utf-8
 
 import csv
 import os
@@ -30,8 +30,8 @@ student_add_aliases.update({
 
 class DbStudentAddApp(NbGrader):
 
-    name = 'nbgrader-db-student-add'
-    description = 'Add a student to the nbgrader database'
+    name = u'nbgrader-db-student-add'
+    description = u'Add a student to the nbgrader database'
 
     aliases = student_add_aliases
     flags = flags
@@ -82,8 +82,8 @@ student_remove_flags.update({
 
 class DbStudentRemoveApp(NbGrader):
 
-    name = 'nbgrader-db-student-remove'
-    description = 'Remove a student from the nbgrader database'
+    name = u'nbgrader-db-student-remove'
+    description = u'Remove a student from the nbgrader database'
 
     aliases = aliases
     flags = student_remove_flags
@@ -142,8 +142,8 @@ class DbGenericImportApp(NbGrader):
         """
         raise NotImplementedError
 
-    name = ""
-    description = ""
+    name = u""
+    description = u""
 
 
     @default('examples')
@@ -238,8 +238,8 @@ class DbGenericImportApp(NbGrader):
 
 class DbStudentImportApp(DbGenericImportApp):
 
-    name = 'nbgrader-db-student-import'
-    description = 'Import students into the nbgrader database from a CSV file'
+    name = u'nbgrader-db-student-import'
+    description = u'Import students into the nbgrader database from a CSV file'
 
     @property
     def table_class(self):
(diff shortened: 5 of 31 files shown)


jhamrick:
Oh good, I'm really glad this resolved the issue!

> might need to be applied at more places in the nbgrader project

I just did a quick grep through the source code there aren't really any other places in the main code base, but there are a few places in the tests that could probably be fixed. Here is a patch that fixes all the places I think might be susceptible to this issue (it also modifies the header notebook used in the tests to include special characters, so hopefully we won't get a recurrence of this bug again):
 [changes.txt](https://github.com/jupyter/nbgrader/files/1424966/changes.txt)

You can apply it with `git apply changes.txt` and then commit the result.'

Thanks for working on this!

consideRatio:
I'll look into this test failure:

```python
    def test_collect_timestamp_skip_older(self, course_dir, archive_dir):
        extracted_dir = join(archive_dir, "..", "extracted")
        submitted_dir = join(course_dir, "submitted")
    
        # submissions are sorted so a before b
        os.makedirs(join(archive_dir, 'ps1_hacker_a_2017-01-30-15-30-10'))
        with io.open(join(archive_dir, 'ps1_hacker_a_2017-01-30-15-30-10', 'problem1.ipynb'), mode='w', encoding='utf-8') as fh:
```

\> fh.write('')
E TypeError: write() argument 1 must be unicode, not str

jhamrick:
Oops, sorry. I tested my patch on python 3 but not python 2, which is where it seems the test failures are coming from... I should have known better! Thanks for looking into it regardless.

consideRatio:
It seems Python 2's string literals are not Unicode literals, while Python 3's string literals are by default Unicode encoded with utf-8.

This test was actually passing on Windows 10. So, if nbgrader is run on Linux, with Python 27, then this is an issue it seems.

The fix is probably to include the following in one of the first two lines of some files, (perhaps only the files where string literals is written to files?). Perhaps someone with previous know-how of encoding issues can suggest a clear line of action?

    # -*- coding: utf-8 -*-

### Reference
Python 2: https://docs.python.org/2/howto/unicode.html#unicode-literals-in-python-source-code
Python 3: https://docs.python.org/3/howto/unicode.html#python-s-unicode-support

consideRatio:
## What I've learned

#### To analyze the error...

Use Linux and Python 2 and run the following test.

```py.test nbgrader/tests/apps/test_nbgrader_zip_collect.py```

---

#### Difference between Unicode and utf-8

> "Unicode only define code points, that is, a number which represents a character. How you store these code points in memory depends of the encoding that you are using. UTF-8 is one way of encoding Unicode characters, among many others." - https://stackoverflow.com/a/643719/2220152

---

#### What `# coding: utf-8` and `# -*- coding: utf-8 -*-` is about

> Ideally, you’d want to be able to write literals in your language’s natural encoding. You could then edit Python source code with your favorite editor which would display the accented characters naturally, and have the right characters used at runtime.
>
> Python supports writing Unicode literals in any encoding, but you have to declare the encoding being used. This is done by including a special comment as either the first or second line of the source file:
>
> ```# coding: utf-8```
>
> "The -*- symbols indicate to Emacs that the comment is special; they have no significance to Python but are a convention." - https://docs.python.org/2/howto/unicode.html#unicode-literals-in-python-source-code

#### nbconvert project

nbconvert seem to have run into similar issues. Look at [this io-helper](https://github.com/jupyter/nbconvert/blob/68a3925efef5c304c2c8682ad05cac9560c4173a/nbconvert/utils/io.py) they made.

## What needs to be done?

Without a full grasp of either nbgrader, nbconvert or unicode details in various environments, I'm reluctant to adjust the code-base in nbgrader further without some input from someone. Should we let pretty much all read and write operations be made through a nbconvert io-helper?

I noticed I could quickfix errors in the test files if ...

```python
# did this
io.write(u'')
# instead of
#io.write('')
```

But that didn't feel like the way to go.

jhamrick:
Ugh, yeah, dealing with unicode support between python 2 and 3 can be a real pain---thanks for looking into this so thoroughly!

I *think* using the `u` prefix like you've suggested is the way to go here. I found http://six.readthedocs.io/#six.u (six is a library for python 2/3 compatibility), but the documentation there says that function shouldn't be necessary as the `u` prefix exists in python 3.3+, which makes me think it is actually probably the correct solution. But it might not always be the solution if there are other issues with other tests.

I am currently working on finishing up another PR but I'll look into this some more when I'm done with that.

consideRatio:
@jhamrick okay! I went with it (using the `u` prefix) and saw where it got me.

* I searched for all unicode string literals (`u'...'`) and added `# coding: utf-8` too all those files-
* Where you updated `open` to `io.open(..., encoding="utf-8")` statements, I updated writing of normal literals to unicode literals.
* I found situations where unicode string literals had been used but not systematically, and made the use of them systematic.
* I ran all tests successfully on my linux with python2 (578 passed, 3 skipped, 0 failed)

Pweh! Closer... perhaps... at least :)

nbgrader/apps/assignapp.py
```diff
@@ -1,3 +1,5 @@
+# coding: utf-8
```
jhamrick: I think these lines aren't strictly necessary since this will only make a difference if the special characters are in the python source code itself, and I don't think this is the case anywhere in nbgrader. But I don't think it hurts to have these lines, so it's fine for them to be there.

jhamrick:
Tests are passing! Thanks for putting in so much work for this 🎉

nbgrader/apps/assignapp.py
consideRatio: Ah, I was uncertain about this myself, thanks for the feedback!

consideRatio:
@jhamrick Thanks for the encouragement and guidance!

I really appreciate Jupyter and the tools surrounding it such as the nbgrader project - so thanks for your work as well :D

jhamrick:
Thanks! 😄

PR #914, opened 2017-11-24
Archiving an assignment

kiranvasudev:
Hello,
  I have been working on a feature that would help share an assignment in the form of an archive. 
Once the student version of an assignment is generated, a button to download it is displayed. When the button is clicked, the user/professor/teacher can download the assignment in the form of a .zip file and this assignment can then be shared with the students.

The working is shown below:
* The list of assignments in the manage assignments page is shown.
Manage assignment page with the list of assignments
  
* Once the generate button is clicked, a button(floppy) is shown.
Download as an archive button

* When the floppy is clicked, all the files and folders that are part of the assignment are downloaded in the form of a .zip file. When extracted, the structure of the contents looks like:
assignment_name
    ├── folder1 in assignment
    │      └── folder1 files
    ├── folder2 in assignment
    │      └── subfolder in folder 2
    │              └── subfolder files
    ├── file1 in assignment
    └── file2 in assignment

I would greatly appreciate it if you kindly give me some feedback regarding this.
Thank you in advance.

### DIFF
--- a/nbgrader/converters/base.py
+++ b/nbgrader/converters/base.py
@@ -4,6 +4,7 @@ import re
 import shutil
 import sqlalchemy
 import traceback
+import zipfile
 
 from traitlets.config import LoggingConfigurable, Config
 from traitlets import Bool, List, Dict, Integer, Instance, Type
@@ -293,6 +294,9 @@ class BaseConverter(LoggingConfigurable):
                 for notebook_filename in self.notebooks:
                     self.convert_single_notebook(notebook_filename)
 
+                # create the archive
+                self.make_archive();
+
                 # set assignment permissions
                 self.set_permissions(gd['assignment_id'], gd['student_id'])
 
@@ -359,3 +363,28 @@ class BaseConverter(LoggingConfigurable):
 
             self.log.error(msg)
             raise NbGraderException(msg)
+
+    def recursively_archive(self, directory, archive_name):
+        all_files_and_folders = [ item for item in os.listdir(directory) if item != archive_name+".zip" ]
+
+        for item in all_files_and_folders:
+            full_path = os.path.join(directory, item)
+            if os.path.isfile(full_path):
+                self.archive.write(os.path.join(self.writer.build_directory, full_path.split(archive_name, 1)[-1][1:]), archive_name+"/"+full_path.split(archive_name, 1)[-1][1:])
+            elif os.path.isdir(full_path):
+                self.recursively_archive(full_path, archive_name)
+
+    def make_archive(self):
+        current_dir = self.writer.build_directory
+        archive_name = os.path.basename(self.writer.build_directory)
+
+        if archive_name+".zip" in os.listdir(current_dir):
+            os.remove(current_dir+"/"+archive_name+".zip")
+            self.log.info("Zip file already exists. File removed.")
+
+        self.archive = zipfile.ZipFile(current_dir+"/"+archive_name+".zip", 'w', compression=zipfile.ZIP_DEFLATED)
+
+        self.recursively_archive(current_dir, archive_name)
+
+        self.archive.close()
+        self.log.info("The archive has been created.")
--- a/nbgrader/server_extensions/formgrader/static/js/manage_assignments.js
+++ b/nbgrader/server_extensions/formgrader/static/js/manage_assignments.js
@@ -144,6 +144,19 @@ var AssignmentUI = Backbone.View.extend({
                 .append($("<span/>")
                     .addClass("glyphicon glyphicon-search")
                     .attr("aria-hidden", "true")));
+
+            this.$release.append($("<a/>")
+                    .attr({
+                        href: base_url + "/tree/" + url_prefix + "/" + release_path + "/" + this.$name.text() + ".zip",
+                        download: this.$name.text()
+                    })
+                    .append($("<span/>")
+                        .addClass("glyphicon glyphicon-floppy-disk")
+                        .attr("aria-hidden", "true")))
+                    .append($("<span/>")
+                        .addClass("glyphicon glyphicon-option-vertical")
+                        .attr("aria-hidden", "true")
+                        .attr("style", "margin: 0 5%"));
         }
 
         // release


nbgrader/converters/base.py
```diff
@@ -293,6 +294,9 @@ def _handle_failure(gd):
                 for notebook_filename in self.notebooks:
                     self.convert_single_notebook(notebook_filename)
 
+                # create the archive
+                self.make_archive();
```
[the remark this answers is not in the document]
jhamrick: Also, I think this isn't quite the right spot to have it actually create the archive, because this will cause nbgrader to create a zip archive for all steps of the process--not just `nbgrader assign`. I think what you want to do is probably to add a new "dummy" function to the base class (called something like `after_assignment_converted()`, or something alone those lines) that by default does nothing. Then, in the `Assign` class, override that function to call the archiving machinery (and then move `make_archive` and `recursively_archive` into the `Assign` class as well).

nbgrader/converters/base.py
```diff
@@ -359,3 +363,28 @@ def _handle_failure(gd):
 
             self.log.error(msg)
             raise NbGraderException(msg)
+
+    def recursively_archive(self, directory, archive_name):
+        all_files_and_folders = [ item for item in os.listdir(directory) if item != archive_name+".zip" ]
```
jhamrick: I know this is a little nitpicky, but I'd appreciate if you could match your code style to the same style as in the rest of the file, which largely follows PEP8---i.e. spaces around `+` signs, no spaces after `[` or before `]`, etc.

nbgrader/converters/base.py
```diff
@@ -359,3 +363,28 @@ def _handle_failure(gd):
...
+            if os.path.isfile(full_path):
+                self.archive.write(os.path.join(self.writer.build_directory, full_path.split(archive_name, 1)[-1][1:]), archive_name+"/"+full_path.split(archive_name, 1)[-1][1:])
+            elif os.path.isdir(full_path):
+                self.recursively_archive(full_path, archive_name)
+
+    def make_archive(self):
+        current_dir = self.writer.build_directory
+        archive_name = os.path.basename(self.writer.build_directory)
+
+        if archive_name+".zip" in os.listdir(current_dir):
```
jhamrick: I would only allow removing existing zip files if `force` is set to True, otherwise fail with an error.

nbgrader/converters/base.py
```diff
@@ -359,3 +363,28 @@ def _handle_failure(gd):
...
             self.log.error(msg)
             raise NbGraderException(msg)
+
+    def recursively_archive(self, directory, archive_name):
+        all_files_and_folders = [ item for item in os.listdir(directory) if item != archive_name+".zip" ]
+
+        for item in all_files_and_folders:
+            full_path = os.path.join(directory, item)
+            if os.path.isfile(full_path):
+                self.archive.write(os.path.join(self.writer.build_directory, full_path.split(archive_name, 1)[-1][1:]), archive_name+"/"+full_path.split(archive_name, 1)[-1][1:])
```
jhamrick: It might be a bit cleaner to use `os.path.relpath` here rather than splitting on `archive_name`, i.e. `os.path.relpath(fullpath, archive_name)`

jhamrick: I would also not rely on `self.writer.build_directory` here, but pass in the root directory that you care about as an argument to the function and then use that.

nbgrader/converters/base.py
```diff
@@ -359,3 +363,28 @@ def _handle_failure(gd):
...
+            elif os.path.isdir(full_path):
+                self.recursively_archive(full_path, archive_name)
+
+    def make_archive(self):
+        current_dir = self.writer.build_directory
+        archive_name = os.path.basename(self.writer.build_directory)
```
jhamrick: `make_archive` should probably take as an argument the assignment id, which you can then use as the archive name rather than getting it from the writer.

nbgrader/converters/base.py
```diff
@@ -359,3 +363,28 @@ def _handle_failure(gd):
...
+
+        if archive_name+".zip" in os.listdir(current_dir):
+            os.remove(current_dir+"/"+archive_name+".zip")
+            self.log.info("Zip file already exists. File removed.")
+
+        self.archive = zipfile.ZipFile(current_dir+"/"+archive_name+".zip", 'w', compression=zipfile.ZIP_DEFLATED)
```
jhamrick: The only thing that should have access to the archive is `recursively_archive`, so I would pass the handle to the archive into that function as an argument rather than setting it as an attribute of the class. You should also probably wrap the archiving funcationality with a try/except/finally so you can still close the archive even if there is an error.

nbgrader/converters/base.py
```diff
@@ -359,3 +363,28 @@ def _handle_failure(gd):
...
+            elif os.path.isdir(full_path):
+                self.recursively_archive(full_path, archive_name)
+
+    def make_archive(self):
+        current_dir = self.writer.build_directory
+        archive_name = os.path.basename(self.writer.build_directory)
+
+        if archive_name+".zip" in os.listdir(current_dir):
+            os.remove(current_dir+"/"+archive_name+".zip")
+            self.log.info("Zip file already exists. File removed.")
```
jhamrick: Specify the path to the zip file here.

nbgrader/converters/base.py
```diff
@@ -359,3 +363,28 @@ def _handle_failure(gd):
...
+
+        self.archive = zipfile.ZipFile(current_dir+"/"+archive_name+".zip", 'w', compression=zipfile.ZIP_DEFLATED)
+
+        self.recursively_archive(current_dir, archive_name)
+
+        self.archive.close()
+        self.log.info("The archive has been created.")
```
jhamrick: Specify the path to the zip file that has been created.

nbgrader/converters/base.py
```diff
@@ -359,3 +363,28 @@ def _handle_failure(gd):
 
             self.log.error(msg)
             raise NbGraderException(msg)
+
+    def recursively_archive(self, directory, archive_name):
+        all_files_and_folders = [ item for item in os.listdir(directory) if item != archive_name+".zip" ]
+
+        for item in all_files_and_folders:
```
jhamrick: I would use `os.walk` here rather than first coming up with a list of all the files and folders. You can then use `continue` to skip if the file matches the archive name.

kiranvasudev:
Thank you so much for your feedback. I will work on the requested changes and keep you updated on the progress.

nbgrader/converters/base.py
```diff
@@ -359,3 +363,28 @@ def _handle_failure(gd):
...
+                self.archive.write(os.path.join(self.writer.build_directory, full_path.split(archive_name, 1)[-1][1:]), archive_name+"/"+full_path.split(archive_name, 1)[-1][1:])
+            elif os.path.isdir(full_path):
+                self.recursively_archive(full_path, archive_name)
+
+    def make_archive(self):
+        current_dir = self.writer.build_directory
+        archive_name = os.path.basename(self.writer.build_directory)
+
+        if archive_name+".zip" in os.listdir(current_dir):
+            os.remove(current_dir+"/"+archive_name+".zip")
```
cdvv7788: When working with paths, it is better to use https://docs.python.org/2/library/os.path.html#os.path.join
I am not sure that this concatenation works, for example, in windows (not sure if the package is intended to be run in windows machines, but that's beside my point).
Also, concatenations, in other contexts, are better handled by .format

jhamrick: 👍

jhamrick:
Hi @kiranvasudev, I just wanted to check if you think you are planning to finish this PR?

kiranvasudev:
Hi @jhamrick. I am sorry, but at the moment, I will not be working on this.

jhamrick:
Ok, thanks for the update. In that case I am going to go ahead and close this PR, but if you (or anybody else) wants this functionality in the future please do feel free to resubmit a new one!

Issue #929, opened 2018-02-16
Slow _filter_existing_notebooks impacts each submission


ryanlovett:
On a deployment with ~600 students, _filter_existing_notebooks takes about 30s. This hits us when manual grading. A single submission is loaded (/formgrader/submissions/:submission_id) and that invokes api.get_notebook_submission_indices which calls the filter which walks the filesystem and filters out non-existing files. So when the grader presses Next, the next submission is loaded which causes another 30s filesystem walk.

### `nbgrader --version`
0.6.0.dev
### `jupyterhub --version` (if used with JupyterHub)
0.8.1
### `jupyter notebook --version` 
5.4.0
### Expected behavior
Indexes are cached or are built asynchronously?
### Actual behavior
Indexes are gathered for each submission load.
### Steps to reproduce the behavior
Manual grading in big course, visit assignment. Then visit first submission.

ryanlovett:
Re the filter:
> This functionality is necessary for cases where student delete or rename on or more notebooks in their assignment, but still submit the assignment.

Should this invalidate their submission when submitting?

ryanlovett:
I see that one can set ExchangeSubmit.strict=True. So I think _filter_existing_notebooks should be a no-op when this is set?

jhamrick:
Wow, that seems really slow. There probably should be a way to either disable this call if `ExchangeSubmit.strict == True` but I am also curious why it is so slow for you in the first place. I just did a quick test with 600 students and for me it takes less than a second to run `get_notebook_submission_indices` (which is what calls `_filter_existing_notebooks`):

>>>
In [8]: len(api.get_notebook_submission_indices("ps1", "problem1"))
Out[8]: 601

In [9]: %timeit api.get_notebook_submission_indices("ps1", "problem1")
853 ms ± 76 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Then again... this is on my laptop which has a SSD, so I suppose this could be substantially slower on a regular hard drive. What sort of setup do you have for running nbgrader?

**Comment by ryanlovett** [comment]
This is happening from within a Kubernetes pod which attaches volumes via NFS. Our values:
```
len(api.get_notebook_submissions('lab01', 'lab01'))
599

%timeit api.get_notebook_submissions('lab01', 'lab01')
33.7 s ± 989 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

jhamrick:
[...]

PR #934, opened 2018-02-18
Skip filtering notebooks when ExchangeSubmit.strict == True

jhamrick:
Should help with #929

@ryanlovett Is it possible for you to test this to see if it helps?

### DIFF
--- a/nbgrader/apps/api.py
+++ b/nbgrader/apps/api.py
@@ -10,7 +10,7 @@ from traitlets import Instance, Enum, Unicode, observe
 
 from ..coursedir import CourseDirectory
 from ..converters import Assign, Autograde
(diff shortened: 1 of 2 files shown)


ryanlovett:
Yes, will do so very soon. Thanks!

ryanlovett:
With the patch it was still slow for us so I added some debugging just to verify that the "strict" configuration was getting picked up. I also tweaked the logic slightly so that it would also skip walking through each notebook. It performed a lot better the second go around so now I should probably retain the logging but revert the extra logic and retry.

https://github.com/berkeley-dsep-infra/nbgrader/commit/7aba39af22dec16d836f28767cfc338694a08ddd

jhamrick:
Cool, thanks---I actually like your logic better so I'm going to update my PR to use that!

PR #936, opened 2018-02-18
Switch from PhamtomJS to Firefox

### LINKED ISSUE #935: Figure out what to do about running nbextension tests
I am now seeing a warning printed out when running the nbextension tests:

```
/Users/jhamrick/miniconda3/envs/nbgrader/lib/python3.5/site-packages/selenium/webdriver/phantomjs/webdriver.py:49: UserWarning: Selenium support for PhantomJS has been deprecated, please use headless versions of Chrome or Firefox instead
```

We should investigate replacing Selenium with headless Chrome or Firefox instead, as suggested by the warning, though I am not sure how well it will work to do this on Travis?

jhamrick:
Fixes #935

Note that unfortunately, the Firefox webdriver does not support capturing logs from javascript. This is annoying, but not a dealbreaker.

In making this change I uncovered a bunch of race conditions in the javascript tests, which is good I guess!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,8 @@ python:
   - 3.5
   - 3.6
 env:
+  global:
+    - MOZ_HEADLESS=1
   matrix:
     - GROUP=docs
     - GROUP=nbextensions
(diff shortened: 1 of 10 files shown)


jhamrick:
So this has turned out to be way more annoying of a change than I anticipated. In particular, it seems the most annoying thing is that Firefox appears to leak memory as more tests are run. For `test_formgrade.py`, this is particularly bad since there are a lot of tests and before it even gets halfway through all the tests it is eating up 1GB of memory. I modified the browser fixture so that it closes and reopens Firefox after every test which forces the memory to be freed, but then this makes the tests take 2x as long to run 😞 

There are also still some tests that are flaky, like `test_load_live_notebook` and `test_next_prev_assignments` in `test_formgrade.py`. I can't seem to reproduce the failure of `test_load_live_notebook` locally, and while I can sometimes reproduce `test_next_prev_assignments`, I cannot determine why it is failing. I suspect there is a problem with the keypresses being sent before the javascript is fully loaded, but Firefox can't capture javascript logs so I cannot isolate the problem 😠

Issue #483, opened 2016-04-06
Improve security by running autograder in isolation


jhamrick:
Perhaps this could be done in Docker containers? Or maybe just a chroot would work?

huwf:
I looked into running `nbgrader autograde` in a Docker container.  This isolated it from other files on the system, but I still had to mount `gradebook.db`, which the assessment could presumably write to since was run by the instructor which we don't really want.  A different kind of security, but we don't want students to be able to change their grade! :)

Which part runs the student's code,  I wasn't able to figure that out?  If I can separate that, then I could run it in a container and return the result to then update the `gradebook.db` on the host.  I'll see how I can integrate this better into the nbgrader code.  I'm running as a cron job at the moment

A bit convoluted probably and not properly tested, but here's my code.  And TBH I can hardly even remember what it does.  Too much work!:

    p = subprocess.Popen([ 'docker', 'run', '-dt', '--rm'
       '-v', '%s:/home/instructor/user-submission/' % (submitted_path),
       # I need to be able to autograde without having the whole autograde directory,
       # but only want to mount on one directory
       # This is a temporary directory.  In autograde.sh we will run:
       # cp -R autograded/username/assignment_name/* autograded-docker
       '-v', '%s:/home/instructor/autograded-docker' % (autograde_path),
       '-v', '%s/gradebook.db:/home/instructor/gradebook.db' % dirpath,
       '-v', '%s/nbgrader_config.py:/home/instructor/nbgrader_config.py' % dirpath,
       '-v', '%s/students.txt:/home/instructor/students.txt' % dirpath,
       'nbgrader-container-admin',
       '/autograde', username, assignment_name, str(instructor_id)])

Dockerfile

	FROM python:3.5.2-slim

	RUN python3 -m pip install nbgrader
	RUN python3 -m pip install pymongo
	RUN python3 -m pip install nose

	RUN useradd -ms /bin/bash instructor

	RUN nbgrader extension install --symlink --debug
	RUN nbgrader extension activate

	WORKDIR /home/instructor

	RUN mkdir -p /home/instructor/submitted
	RUN mkdir /home/instructor/autograded
        # Maybe we could auto generate and build Dockerfile rather than hardcoding
	RUN mkdir -p /home/instructor/source/assignment_name

	# Folders for mounting...
	RUN mkdir -p /home/instructor/user-submission
	RUN mkdir -p /home/instructor/autograded-docker

	RUN touch .nbgrader.log
[...]

PR #970, opened 2018-05-21
[WIP] Splitting of autograde in view of security issue

gertingold:
This PR is a very first attempt to split the autograde process in view of the discussion in #483. It implements the three options `--only-sanitize`, `--only-execute` (which should eventually be run in isolation), and `--only-record`. The last two steps need to be run together with the `--force` option. Running
- `nbgrader autograde --only-sanitize ps1`
- `nbgrader autograde --only-execute --force ps1`
- `nbgrader autograde --only-record --force ps1`
for the example problem set `ps1` correctly recorded the grades in the database. However, I am worried that the use of `--force` could have side effects which I have not detected so far.

Note: This PR is only meant for discussion purposes and to decide whether the solution suggested here is worth being pursued or whether the use of subcommands is a better choice. Thus, this PR is not intended to be merged.

### DIFF
--- a/nbgrader/apps/autogradeapp.py
+++ b/nbgrader/apps/autogradeapp.py
@@ -34,6 +34,39 @@ flags.update({
         {'BaseConverter': {'force': True}},
         "Overwrite an assignment/submission if it already exists."
     ),
+    'only-sanitize': (
+        {
+            'Execute': {'enabled': False},
+            'LimitOutput': {'enabled': False},
+            'SaveAutoGrades': {'enabled': False},
+            'AssignLatePenalties': {'enabled': False},
+        },
+        "Only apply sanitizing preprocessors."
+    ),
+    'only-execute': (
+        {
+            'ClearOutput': {'enabled': False},
+            'DeduplicateIds': {'enabled': False},
+            'OverwriteKernelspec': {'enabled': False},
+            'OverwriteCells': {'enabled': False},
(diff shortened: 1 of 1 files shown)


jhamrick:
Hmm, yeah, so the way you've implemented it so far does have the issue I mentioned in #483 :

> So if you use --force with --only-record, it may delete those files and re-copy them, which would be a problem then because then it will be determining grades based on notebooks that haven't been executed yet.

which is why you have to run the last two steps together. I think what you'll need to do is a bit more restructuring in `converters/autograde.py` so that if you're running with `--only-record` it doesn't copy the files from the source directory in `init_assignment()`, and so that it doesn't do the "sanitizing" step at all.

I think this might be a bit of a more robust approach in the long run too than individually disabling each of the preprocessors, because it might be easy to in the future add a new preprocessor but forget to disable it in the config options.

gertingold:
Thanks for your comment. I will further look into the issue as soon as possible.

gertingold:
Unfortunately, due to lack of time, I was unable to continue work on this PR. According to the discussion on #483, the preferred solution seems to be using a docker image or something similar instead of modifying ``nbgrader`` as attempted in this PR. Since I was able to successfully do grading in ``nbgrader`` within a docker container with the help of ``envkernel``, it looks like there is a good solution to #483. Therefore, I close this PR, at least for the time being. Of course, the PR could be reopened if new interest in the approach chosen here should arise.

PR #972, opened 2018-06-01
Use mathjax macro for formgrader

JanBobolz:
Currently, LaTeX oversets are rendered weird in the [formgrader](https://homepages.upb.de/jbobolz/nbgraderRenderExample/badRender.html) (i.e. while grading a submission), to the point where some submissions become unreadable. 
In the notebook and the [feedback](https://homepages.upb.de/jbobolz/nbgraderRenderExample/feedback.html), everything is rendered fine.
An example notebook to reproduce the problem can be found [here](https://homepages.upb.de/jbobolz/nbgraderRenderExample/formattest.ipynb)

This pull request changes the way mathjax is included on formgrader pages to be consistent with the include on the feedback page (where everything works fine). It also makes sense to me that the formatting should be the same for the formgrader and the feedback.

I have this change running productively for some time now and it seems to work fine. However, I am not familiar with nbgrader's development, so no guarantees that this is not a stupid fix.

edit: as it turns out this was not the fix for my rendering problem. The pull request now provides a minor code quality improvement, I guess :)

### DIFF
--- a/nbgrader/server_extensions/formgrader/templates/formgrade.tpl
+++ b/nbgrader/server_extensions/formgrader/templates/formgrade.tpl
@@ -15,7 +15,7 @@
 {% endfor %}
 
 <!-- Loading mathjax macro -->
-{{ mathjax() }}
+{{ mathjax( resources.base_url + '/' + resources.mathjax_url + '?config=TeX-AMS-MML_HTMLorMML-full') }}
 
 <link rel="stylesheet" href="{{ resources.base_url }}/formgrader/static/css/formgrade.css" />
 
--- a/nbgrader/server_extensions/formgrader/templates/formgrade.tpl
+++ b/nbgrader/server_extensions/formgrader/templates/formgrade.tpl
@@ -1,5 +1,6 @@
 {%- extends 'basic.tpl' -%}
 {% from 'formgrade_macros.tpl' import nav, header %}
+{% from 'mathjax.tpl' import mathjax %}
 
 {%- block header -%}
 <!DOCTYPE html>
@@ -13,26 +14,8 @@
     </style>
 {% endfor %}
 
-<!-- MathJax -->
-<script type="text/javascript">
-window.MathJax = {
-    tex2jax: {
-        inlineMath: [ ['$','$'], ["\\(","\\)"] ],
-        displayMath: [ ['$$','$$'], ["\\[","\\]"] ],
-        processEscapes: true,
-        processEnvironments: true
-    },
-    // Center justify equations in code and markdown cells. Elsewhere
-    // we use CSS to left justify single line equations in code cells.
-    displayAlign: 'center',
-    "HTML-CSS": {
-        styles: {'.MathJax_Display': {"margin": 0}},
-        linebreaks: { automatic: true }
-    }
-};
-</script>
-
-<script type="text/javascript" src="{{ resources.base_url }}/{{ resources.mathjax_url }}?config=TeX-AMS-MML_HTMLorMML-full"></script>
+<!-- Loading mathjax macro -->
+{{ mathjax() }}
 
 <link rel="stylesheet" href="{{ resources.base_url }}/formgrader/static/css/formgrade.css" />
 


nbgrader/server_extensions/formgrader/templates/formgrade.tpl
```diff
@@ -13,26 +14,8 @@
...
-    "HTML-CSS": {
-        styles: {'.MathJax_Display': {"margin": 0}},
-        linebreaks: { automatic: true }
-    }
-};
-</script>
-
-<script type="text/javascript" src="{{ resources.base_url }}/{{ resources.mathjax_url }}?config=TeX-AMS-MML_HTMLorMML-full"></script>
+<!-- Loading mathjax macro -->
+{{ mathjax() }}
```
jhamrick: This removes the ability to specify a custom mathjax url, so maybe change this to:

```
{{ mathjax('{{ resources.base_url }}/{{ resources.mathjax_url }}?config=TeX-AMS-MML_HTMLorMML-full') }}
```

I am not 100% sure that's the right syntax but I think something along those lines should work looking at the macro arguments here: https://github.com/jupyter/nbconvert/blob/master/nbconvert/templates/html/mathjax.tpl

JanBobolz:
After changing the line as requested, it seems like the "fix" to my initial problem was based on using the cdn mathjax (default parameter) instead of the "local" jupyter mathjax. 
I'll investigate a bit and may open another merge request or issue at some point.

This particular pull request can be merged (as far as I'm concerned).

nbgrader/server_extensions/formgrader/templates/formgrade.tpl
JanBobolz: You're right. I've changed it. This destroys my "make formgrader and feedback use the exact same mathjax urls" plans, but after digging around a bit, I see that this idea never made much sense :)

jhamrick:
Hmm, interesting. If you change just the `config` parameter in the URL to the one that's used in the default mathjax macro? I'm wondering if it's just based on that config value or if it really is a difference between the local version and the CDN version.

In any case, the tests are failing for unrelated reasons which I will try to fix and then merge this. Thanks for the help!

JanBobolz:
> If you change just the config parameter in the URL to the one that's used in the default mathjax macro? 

`{{ mathjax( resources.base_url + '/' + resources.mathjax_url + '?config=TeX-AMS_HTML') }}` does not render any LaTeX at all. Firefox says `Loading failed for the <script> with source “https://localhost:8000/services/shared-formgrader/static/components/MathJax/config/TeX-AMS_HTML.js?rev=2.6.0”`.

I've also tried `{{ mathjax( resources.base_url + '/' + resources.mathjax_url + '?config=TeX-AMS-MML_HTMLorMML-full,Safe') }}`, which is (except for missing `delayStartupUntil=configured`) what jupyter notebooks use.
This exhibits the rendering problems described in the initial post with the oversets. It also triggers the error `Loading failed for the <script> with source “https://localhost:8000/services/shared-formgrader/static/components/MathJax/jax/output/HTML-CSS/fonts/TeX/fontdata.js?rev=2.6.0”.`

jhamrick:
Hmm, that's interesting, using `TeX-AMS_HTML` with the local installation doesn't work at all but it does when loading it from the CDN (and also then fixes the problems with the oversets)? It could be that the local Jupyter version doesn't contain the full installation of MathJax; I'm not sure...

I think @mpacer has done some work with MathJax in the notebook; maybe they're more familiar with this issue?

jhamrick:
Also the unrelated test failures should be fixed now. Can you do a rebase?

JanBobolz:
> using TeX-AMS_HTML with the local installation doesn't work at all but it does when loading it from the CDN (and also then fixes the problems with the oversets)? 

Correct. It's also noteworthy that the jupyter *notebook* renders everything perfectly without errors with its local mathjax (`?config=TeX-AMS-MML_HTMLorMML-full,Safe`). However, in the formgrader with the same `config`, we get a script loading error and LaTeX is rendered with weird oversets.

> Also the unrelated test failures should be fixed now. Can you do a rebase?

done.

jhamrick:
Thanks!

Issue #977, opened 2018-06-09
Global tasks


danielmaitre:
Hi all,

I've used nbgrader last year and I plan to use it more in the future. One functionality I am missing is what I will call "tasks", they would be exercises/objectives that span multiple cells. For example the task "Illustrate this aspect of the course with a plot!" would probably require students to create a plot in a code cell and comment on it in a markdown cell. Other tasks might involve multiple code/markdown cells. I think it would be beneficial because students could use more of the notebook features to organise an answer rather than being limited to filling specific parts of single cells. Another possibility is to have global grades for an entire notebook, for example for using good coding practices or awarding points for well structured narrative in the notebook. 

I think it is hard to achieve with what is currently available in nbgrader (if I'm wrong that would be great!) and I'm willing to have a go at implementing it, but before I do I might need some guidance. 

I'm thinking of implementing a new type of cells, task_cell that would be manually graded and could be commented on during the manual marking. It looks like there would be quite a lot of changes to do and before I try to implement it I wanted to make sure that it is a feature that others would find useful and that could be incorporated in nbgrader.

jhamrick:
> One functionality I am missing is what I will call "tasks", they would be exercises/objectives that span multiple cells.

There's nothing technically preventing this from being done in nbgrader---students can certainly provide code/comments in other cells beyond the cell which has the grade in it. I suppose there is a drawback here which is that then you can't provide comments during the feedback phase on other cells, though. But you could get around that too by having those cells be marked as graded but be worth zero points, something like this:

![image](https://user-images.githubusercontent.com/83444/41191427-3ebe0a70-6be7-11e8-87a3-8aa34d2aab38.png)

> have global grades for an entire notebook, for example for using good coding practices or awarding points for well structured narrative in the notebook.

I think that's a reasonable suggestion, though it would require a lot of nontrivial UI changes 🙁

> I'm thinking of implementing a new type of cells, task_cell that would be manually graded and could be commented on during the manual marking.

Can you elaborate a little further about how task_cell would work beyond regular manually graded cells?

danielmaitre:
Thank you for the comment!  I realise that this would introduce quite a bit of changes, that's why I'm asking if it is of interest before trying to implement it.

Thank you for the suggestion with the manually graded cell with 0 points suggestion, I might use it in the interim. The drawback is that I would need to decide on how many cells the students would need while I'd rather them develop their own notebook style. 

[...]

PR #984, opened 2018-06-30
Adding global graded tasks  

danielmaitre:
Hi, 
this is an attempt at adding a new type of cells: task cells. I described what I wanted to achieve in #977 . They are tasks one can set to the students that are not tied to a single cell. I think they are useful when entire sections of notebooks are considered and it allows to ask students to develop parts of a notebook and get grades for it. Task cell can be used for task where we want students to use more than one cell to submit an answer and in this way exercises using the notebook as a platform rather than a single cell.

I'm sorry it ended up being quite a large set of changes. Let me know what you think!

Daniel


nbgrader/alembic/versions/167914646830_added_task_cells.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision = '167914646830'
+down_revision = '724cde206c17'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
```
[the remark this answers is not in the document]
jhamrick: Ah, I see. I think in that case you might not have to use alembic at all if the table is automatically created. But we might want to check this with a different flavor of sql database like Postgres just to be sure that's not sqlite behavior.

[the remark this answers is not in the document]
jhamrick: (I should say, the point of alembic is for database migrations, as you've inferred, so it can automatically modify an old database to be compatible with a new spec. Usually what happens if, say, you add a new column, is that when nbgrader would try to access the database, it would throw an error, hence the need for alembic. But in the context of nbgrader the migrations should always happen through `nbgrader upgrade` rather directly through alembic itself.)

danielmaitre:
Hi, thank you for the thorough review! I'm working on all items above. Concerning the high level comments above:
1. Yes I think the task cell should be markdown cells. It is enforced in the validation step. My concern with the option being only accessible is that it would be less visible as new cells are code cells by default. I considered making selecting "Task cell" convert the cell to markdown cell automatically but finally settled with only enforcing it in the assign step. I have no strong opinion about it, if you think it is better to only have the task cell option visible for markdown cells I can do it.

2. I'll have a go at updating the documentation

3. I first tried to implement the task cells as Gradecell but got into issues when I wanted it to have comments. The comments explicitly refer to a solution cell through the foreign key and it could not be a grade cell. I hope that the solution proposed will also help if we want to develop new types of cells and we can add the ability to be graded or commented simply through the mixin classes. (I'm toying with the idea of adding multiple choice question cells)

danielmaitre:
Ok, I think I've addressed the issues raised in the review, please let me know what you think!

danielmaitre:
Hi, I'm sorry I'm not quite sure what I should do now with this PR?

jhamrick:
Thanks so much for all the additional work on this PR, and sorry I've been slow to look at it. I'll take a closer look in the next few days!

danielmaitre:
HI, I've made the changes as requested, sorry for the delay. As a side note I've used nbgrader with task cells for exercises for a class with 180 students and it worked very well!

nbgrader/nbextensions/create_assignment/main.js
```diff
@@ -386,28 +424,43 @@ define([
                     set_solution(cell, true);
                     set_grade(cell, true);
                     set_locked(cell, false);
+                    set_task(cell, false);
+                } else if (val === "task") {
+                    set_schema_version(cell);
+                    set_solution(cell, false);
+                    set_grade(cell, false);
+                    set_locked(cell, true);
+                    set_task(cell,true);
```
jhamrick:
```suggestion
                    set_task(cell, true);
```

nbgrader/nbextensions/create_assignment/main.js
```diff
@@ -386,28 +424,43 @@ define([
...
+                    if (cell.get_text() === ''){
+                      cell.set_text('Describe the task here!')
+                    }
                 } else if (val === "solution") {
                     set_schema_version(cell);
                     set_solution(cell, true);
                     set_grade(cell, false);
                     set_locked(cell, false);
+                    set_task(cell,false);
```
jhamrick:
```suggestion
                    set_task(cell, false);
```

nbgrader/nbextensions/create_assignment/main.js
```diff
@@ -386,28 +424,43 @@ define([
...
                 } else if (val === "tests") {
                     set_schema_version(cell);
                     set_solution(cell, false);
                     set_grade(cell, true);
                     set_locked(cell, true);
+                    set_task(cell,false);
```
jhamrick:
```suggestion
                    set_task(cell, false);
```

nbgrader/nbextensions/create_assignment/main.js
```diff
@@ -386,28 +424,43 @@ define([
...
                 } else if (val === "readonly") {
                     set_schema_version(cell);
                     set_solution(cell, false);
                     set_grade(cell, false);
                     set_locked(cell, true);
+                    set_task(cell,false);
```
jhamrick:
```suggestion
                    set_task(cell, false);
```

nbgrader/nbextensions/create_assignment/main.js
```diff
@@ -386,28 +424,43 @@ define([
...
                 } else {
                     throw new Error("invalid nbgrader cell type: " + val);
                 }
             };
 
             var getter = function (cell) {
-                if (is_solution(cell) && is_grade(cell)) {
+                if (is_task(cell) ) {
```
jhamrick:
```suggestion
                if (is_task(cell)) {
```

jhamrick:
Oops, tests look like they are passing but actually they are not, this is due to a bug introduced I think by #1053. Will hold off on merging until that's fixed, then we'll need to do another rebase/merge (sorry about that!)

jhamrick:
See #1054

nbgrader/api.py
```diff
@@ -234,34 +290,58 @@ def __repr__(self):
...
+        """
+        return {
+            "id": self.id,
+            "name": self.name,
+            "notebook": self.notebook.name,
+            "assignment": self.assignment.name
+        }
+
+    def __repr__(self):
+        return "{}/{}".format(self.notebook, self.name)
```
jhamrick:
```suggestion
        return "SolutionCell<{}/{}/{}>".format(
            self.assignment.name, self.notebook.name, self.name)
```

williamstein:
> Regarding the drop down menu, yes, I think it would be better to only have the "task cell" option visible for markdown cells.

[The documentation](https://nbgrader.readthedocs.io/en/stable/user_guide/creating_and_grading_assignments.html#manually-graded-task-cells) asserts this too "If you select the “Manually graded task” option (available for markdown cells), the nbgrader extension will ..."  However, the "Manually graded task" option seems to be available for code cells as well:

image

I don't really care either way.  It's just that I'm implementing a hopefully 100% compatible version of nbgrader from scratch for https://cocalc.com, and I need to know what to implement...

Issue #862, opened 2017-08-11
Update developer installation instructions


jhamrick:
So that when installing the extensions, all the commands are in the same code block.

jzf2101:
@fenwickipedia also might be something to consider

fenwickipedia:
Okay this looks good, I'm working on it. Thanks!

jhamrick:
Reopening this issue because #911 didn't quite address what I had in mind, though this is totally my bad for (1) not explaining well enough in the issue description and (2) not being around to review. More specifically, in the developer installation instructions [here](https://github.com/jupyter/nbgrader/blob/master/nbgrader/docs/source/contributor_guide/installation_developer.rst#installing-notebook-extensions) there are two separate code blocks containing required pieces of the installation process, one that looks like this:

```
# The nbextensions are JavaScript/HTML/CSS so they require
# separate installation and enabling.
jupyter nbextension install --symlink --sys-prefix --py nbgrader
jupyter nbextension enable --sys-prefix --py nbgrader
```

and one which looks like this:

```
# The serverextension is a Python module inside nbgrader, so only an
# enable step is needed.
jupyter serverextension enable --sys-prefix --py nbgrader
```

These should ideally be merged so they are in the same code block (as in the user guide installation instructions) to make it easier for people to just copy and paste everything in one go.

jhamrick:
Fixed by #989

PR #989, opened 2018-07-04
Update developer installation instructions

cdvv7788:
Make changes as required at https://github.com/jupyter/nbgrader/issues/862

### DIFF
--- a/nbgrader/docs/source/contributor_guide/installation_developer.rst
+++ b/nbgrader/docs/source/contributor_guide/installation_developer.rst
@@ -27,19 +27,18 @@ Previously this was done using the ``nbgrader extension install`` command.
 However, moving forward this is done using the ``jupyter nbextension`` and
 ``jupyter serverextension`` commands.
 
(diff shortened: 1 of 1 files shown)


cdvv7788:
@jhamrick Is this what you had in mind for https://github.com/jupyter/nbgrader/issues/862 ?

jhamrick:
Thanks!

I think it would be a bit clearer for new developers that all three commands should be run if the they are one right after another with no linebreaks/comments, e.g.:

```
jupyter nbextension install --symlink --sys-prefix --py nbgrader
jupyter nbextension enable --sys-prefix --py nbgrader
jupyter serverextension enable --sys-prefix --py nbgrader
```

The other documentation (e.g. about what the symlink is doing) could go before or after (but ideally not as comments in the code block itself).

cdvv7788:
@jhamrick Better?

jhamrick:
Looks great now, thanks!

PR #1000, opened 2018-08-02
Allow instructors to share files via shared group id

rkdarst:
Hi,

I have been working on a setup where I wanted instructors to be able to share releasing files, collecting, and so on.nbgrader says that it doesn't support this, but I tried anyway.  This PR contains the changes needed to make this work, which have been mostly minimal.  Things are not perfect, and some of these changes may be not needed.  I'm still testing locally and this is mainly to get some more expert opinions.

This adds a new option `groupshared` which enables the behavior.  By default, no behavior should change.

These changes assume that the following are set by the environment (admins).  They are satisfied in my setup because we run via kubespawner and I carefully control these:
- instructors have a unique-per-course primary gid which students do not have.  Alternatively, the setgid bit could be set (in practice I do both).  If using setgid, you have to set that up yourself.
- umasks are 0007
- shared course and exchange directories, like normal.
- Instructors won't do something wrong with permissions which breaks the security.

We expect these behaviors to happen when `groupshared = True`:
- All info in the course directory is `g+rw`.  Instructors can all edit files.
- Multiple instructors can release and collect files from the exchange.  Basically, when releasing, group permissions mirror user, not other.  Because the submitted folder is set g+rwx, other instructors can also collect.
- Don't insist on non-group writability in some places.

It actually seems to mostly work (though I am still doing internal tests) Remaining problems (I will keep updating this as I find more):
- [x] The `groupshared` set in two places.  Set it only once.  (I didn't see a common base class...)
- [x] when the `release/` and `release/{assignment_id}` directories are created, it is `g-w`.  Haven't been able to figure out why yet - it is not the umask, and I can't figure out where in the code the dirs are created. (help requested!)
- [x] sqlite3 hardcodes a create mode of 644.  I hacked around this by touching `gradebook.db` and setting a proper mode first.
- [x] tests
- [x] Study 664 vs 444 as default permissions for autograde. (rkdarst)
- [x] Study what the problem with ensure_root was (rkdarst)

For now I am trying to just make *something* that will work, then I can work on improving it more.  What do you think?  Is this something that I should work on to merge upstream?

### DIFF
--- a/nbgrader/converters/base.py
+++ b/nbgrader/converters/base.py
@@ -264,7 +264,8 @@ class BaseConverter(LoggingConfigurable):
             if self.groupshared and os.stat(dirname).st_uid == os.getuid():
                 #for subdirname in subdirnames:
                 #    subdirname = os.path.join(dirname, subdirname)
-                os.chmod(dirname, (os.stat(dirname).st_mode|0o2770) & 0o2777)
+                try:   os.chmod(dirname, (os.stat(dirname).st_mode|0o2770) & 0o2777)
+                except PermissionError: pass
         # If groupshared, set write permissions on directories.  Directories
         # are created within ipython_genutils.path.ensure_dir_exists via
         # nbconvert.writer, (unless there are supplementary files) with a
@@ -273,9 +274,10 @@ class BaseConverter(LoggingConfigurable):
         if self.groupshared:
             # Root may be created in this step, and is not included above.
             rootdir = self.coursedir.format_path(self._output_directory, '.', '.')
-            if os.stat(rootdir).st_uid == os.getuid():
-                # Add 2770 to existing dir permissions (don't unconditionally override)
-                os.chmod(rootdir, (os.stat(rootdir).st_mode|0o2770) & 0o2777)
+            # Add 2770 to existing dir permissions (don't unconditionally override)
+            try:   os.chmod(rootdir, (os.stat(rootdir).st_mode|0o2770) & 0o2777)
+            except PermissionError: pass
+
 
     def convert_single_notebook(self, notebook_filename):
         """Convert a single notebook.
--- a/nbgrader/converters/base.py
+++ b/nbgrader/converters/base.py
@@ -264,8 +264,10 @@ class BaseConverter(LoggingConfigurable):
             if self.groupshared and os.stat(dirname).st_uid == os.getuid():
                 #for subdirname in subdirnames:
                 #    subdirname = os.path.join(dirname, subdirname)
-                try:   os.chmod(dirname, (os.stat(dirname).st_mode|0o2770) & 0o2777)
-                except PermissionError: pass
+                try:
+                    os.chmod(dirname, (os.stat(dirname).st_mode|0o2770) & 0o2777)
+                except PermissionError:
+                    pass
         # If groupshared, set write permissions on directories.  Directories
         # are created within ipython_genutils.path.ensure_dir_exists via
         # nbconvert.writer, (unless there are supplementary files) with a
@@ -275,8 +277,10 @@ class BaseConverter(LoggingConfigurable):
             # Root may be created in this step, and is not included above.
             rootdir = self.coursedir.format_path(self._output_directory, '.', '.')
             # Add 2770 to existing dir permissions (don't unconditionally override)
-            try:   os.chmod(rootdir, (os.stat(rootdir).st_mode|0o2770) & 0o2777)
-            except PermissionError: pass
+            try:
+                os.chmod(rootdir, (os.stat(rootdir).st_mode|0o2770) & 0o2777)
+            except PermissionError:
+                pass
 
 
     def convert_single_notebook(self, notebook_filename):
--- a/nbgrader/converters/base.py
+++ b/nbgrader/converters/base.py
@@ -44,20 +44,9 @@ class BaseConverter(LoggingConfigurable):
         )
     ).tag(config=True)
 
-    groupshared = Bool(
-        False,
-        help=dedent(
-            """
-            Be less strict about user permissions (instructor files are by
-            default group writeable.  Requires that admins ensure that primary
-            groups are correct!
-            """
-        )
-    ).tag(config=True)
-
     @default("permissions")
     def _permissions_default(self):
-        return 664 if self.groupshared else 444
+        return 664 if self.coursedir.groupshared else 444
 
     coursedir = Instance(CourseDirectory, allow_none=True)
 
@@ -261,7 +250,7 @@ class BaseConverter(LoggingConfigurable):
             for filename in filenames:
                 os.chmod(os.path.join(dirname, filename), permissions)
             # If groupshared, set dir permissions - see comment below.
-            if self.groupshared and os.stat(dirname).st_uid == os.getuid():
+            if self.coursedir.groupshared and os.stat(dirname).st_uid == os.getuid():
                 #for subdirname in subdirnames:
                 #    subdirname = os.path.join(dirname, subdirname)
                 try:
@@ -273,7 +262,7 @@ class BaseConverter(LoggingConfigurable):
         # nbconvert.writer, (unless there are supplementary files) with a
         # default mode of 755 and there is no way to pass the mode arguments
         # all the way to there!  So we have to walk and fix.
-        if self.groupshared:
+        if self.coursedir.groupshared:
             # Root may be created in this step, and is not included above.
             rootdir = self.coursedir.format_path(self._output_directory, '.', '.')
             # Add 2770 to existing dir permissions (don't unconditionally override)
--- a/nbgrader/converters/base.py
+++ b/nbgrader/converters/base.py
@@ -44,6 +44,17 @@ class BaseConverter(LoggingConfigurable):
         )
     ).tag(config=True)
 
+    groupshared = Bool(
+        False,
+        help=dedent(
+            """
+            Be less strict about user permissions (instructor files are by
+            default group writeable.  Requires that admins ensure that primary
+            groups are correct!
+            """
+        )
+    ).tag(config=True)
+
     @default("permissions")
     def _permissions_default(self):
         return 444
--- a/nbgrader/converters/base.py
+++ b/nbgrader/converters/base.py
@@ -264,7 +264,7 @@ class BaseConverter(LoggingConfigurable):
             if self.groupshared and os.stat(dirname).st_uid == os.getuid():
                 #for subdirname in subdirnames:
                 #    subdirname = os.path.join(dirname, subdirname)
-                os.chmod(dirname, (os.stat(dirname).st_mode|0o770) & 0o777)
+                os.chmod(dirname, (os.stat(dirname).st_mode|0o2770) & 0o2777)
         # If groupshared, set write permissions on directories.  Directories
         # are created within ipython_genutils.path.ensure_dir_exists via
         # nbconvert.writer, (unless there are supplementary files) with a
@@ -274,8 +274,8 @@ class BaseConverter(LoggingConfigurable):
             # Root may be created in this step, and is not included above.
             rootdir = self.coursedir.format_path(self._output_directory, '.', '.')
             if os.stat(rootdir).st_uid == os.getuid():
-                # Add 770 to existing dir permissions (don't unconditionally override)
-                os.chmod(rootdir, (os.stat(rootdir).st_mode|0o770) & 0o777)
+                # Add 2770 to existing dir permissions (don't unconditionally override)
+                os.chmod(rootdir, (os.stat(rootdir).st_mode|0o2770) & 0o2777)
 
     def convert_single_notebook(self, notebook_filename):
         """Convert a single notebook.
--- a/nbgrader/converters/base.py
+++ b/nbgrader/converters/base.py
@@ -250,11 +250,12 @@ class BaseConverter(LoggingConfigurable):
             for filename in filenames:
                 os.chmod(os.path.join(dirname, filename), permissions)
             # If groupshared, set dir permissions - see comment below.
-            if self.coursedir.groupshared and os.stat(dirname).st_uid == os.getuid():
+            st_mode = os.stat(dirname).st_mode
+            if self.coursedir.groupshared and st_mode & 0o2770 != 0o2770:
                 try:
-                    os.chmod(dirname, (os.stat(dirname).st_mode|0o2770) & 0o2777)
+                    os.chmod(dirname, (st_mode|0o2770) & 0o2777)
                 except PermissionError:
-                    pass
+                    self.log.warning("Could not update permissions of %s to make it groupshared", dirname)
         # If groupshared, set write permissions on directories.  Directories
         # are created within ipython_genutils.path.ensure_dir_exists via
         # nbconvert.writer, (unless there are supplementary files) with a
@@ -264,10 +265,12 @@ class BaseConverter(LoggingConfigurable):
             # Root may be created in this step, and is not included above.
             rootdir = self.coursedir.format_path(self._output_directory, '.', '.')
             # Add 2770 to existing dir permissions (don't unconditionally override)
-            try:
-                os.chmod(rootdir, (os.stat(rootdir).st_mode|0o2770) & 0o2777)
-            except PermissionError:
-                pass
+            st_mode = os.stat(rootdir).st_mode
+            if st_mode & 0o2770 != 0o2770:
+                try:
+                    os.chmod(rootdir, (st_mode|0o2770) & 0o2777)
+                except PermissionError:
+                    self.log.warning("Could not update permissions of %s to make it groupshared", rootdir)
 
 
     def convert_single_notebook(self, notebook_filename):
--- a/nbgrader/converters/base.py
+++ b/nbgrader/converters/base.py
@@ -57,7 +57,7 @@ class BaseConverter(LoggingConfigurable):
 
     @default("permissions")
     def _permissions_default(self):
-        return 444
+        return 664 if self.groupshared else 444
 
     coursedir = Instance(CourseDirectory, allow_none=True)
 
--- a/nbgrader/converters/base.py
+++ b/nbgrader/converters/base.py
@@ -251,8 +251,6 @@ class BaseConverter(LoggingConfigurable):
                 os.chmod(os.path.join(dirname, filename), permissions)
             # If groupshared, set dir permissions - see comment below.
             if self.coursedir.groupshared and os.stat(dirname).st_uid == os.getuid():
-                #for subdirname in subdirnames:
-                #    subdirname = os.path.join(dirname, subdirname)
                 try:
                     os.chmod(dirname, (os.stat(dirname).st_mode|0o2770) & 0o2777)
                 except PermissionError:
--- a/nbgrader/converters/base.py
+++ b/nbgrader/converters/base.py
@@ -260,6 +260,22 @@ class BaseConverter(LoggingConfigurable):
         for dirname, _, filenames in os.walk(dest):
             for filename in filenames:
                 os.chmod(os.path.join(dirname, filename), permissions)
+            # If groupshared, set dir permissions - see comment below.
+            if self.groupshared and os.stat(dirname).st_uid == os.getuid():
+                #for subdirname in subdirnames:
+                #    subdirname = os.path.join(dirname, subdirname)
+                os.chmod(dirname, (os.stat(dirname).st_mode|0o770) & 0o777)
+        # If groupshared, set write permissions on directories.  Directories
+        # are created within ipython_genutils.path.ensure_dir_exists via
+        # nbconvert.writer, (unless there are supplementary files) with a
+        # default mode of 755 and there is no way to pass the mode arguments
+        # all the way to there!  So we have to walk and fix.
+        if self.groupshared:
+            # Root may be created in this step, and is not included above.
+            rootdir = self.coursedir.format_path(self._output_directory, '.', '.')
+            if os.stat(rootdir).st_uid == os.getuid():
+                # Add 770 to existing dir permissions (don't unconditionally override)
+                os.chmod(rootdir, (os.stat(rootdir).st_mode|0o770) & 0o777)
 
     def convert_single_notebook(self, notebook_filename):
         """Convert a single notebook.
--- a/nbgrader/exchange/exchange.py
+++ b/nbgrader/exchange/exchange.py
@@ -142,12 +142,12 @@ class Exchange(LoggingConfigurable):
         if self.groupshared:
             for dirname, _, filenames in os.walk(dest):
                 # dirs become ug+rwx
-                if os.stat(dirname).st_uid == os.getuid():
-                    os.chmod(dirname, (os.stat(dirname).st_mode|0o2770) & 0o2777)
+                try:   os.chmod(dirname, (os.stat(dirname).st_mode|0o2770) & 0o2777)
+                except PermissionError: pass
                 for filename in filenames:
                     filename = os.path.join(dirname, filename)
-                    if os.stat(filename).st_uid == os.getuid():
-                        os.chmod(filename, (os.stat(filename).st_mode|0o660) & 0o777)
+                    try:    os.chmod(filename, (os.stat(filename).st_mode|0o660) & 0o777)
+                    except PermissionError: pass
 
     def start(self):
         if sys.platform == 'win32':
--- a/nbgrader/exchange/exchange.py
+++ b/nbgrader/exchange/exchange.py
@@ -68,17 +68,6 @@ class Exchange(LoggingConfigurable):
         )
     ).tag(config=True)
 
-    groupshared = Bool(
-        False,
-        help=dedent(
-            """
-            Be less strict about user permissions (instructor files are by
-            default group writeable.  Requires that admins ensure that primary
-            groups are correct!
-            """
-        )
-    ).tag(config=True)
-
     coursedir = Instance(CourseDirectory, allow_none=True)
     authenticator = Instance(Authenticator, allow_none=True)
 
@@ -139,7 +128,7 @@ class Exchange(LoggingConfigurable):
                                                log=self.log))
         # copytree copies access mode too - so we must add go+rw back to it if
         # we are in groupshared.
-        if self.groupshared:
+        if self.coursedir.groupshared:
             for dirname, _, filenames in os.walk(dest):
                 # dirs become ug+rwx
                 try:
@@ -157,7 +146,7 @@ class Exchange(LoggingConfigurable):
         if sys.platform == 'win32':
             self.fail("Sorry, the exchange is not available on Windows.")
 
-        if not self.groupshared:
+        if not self.coursedir.groupshared:
             # This just makes sure that directory is o+rwx.  In group shared
             # case, it is up to admins to ensure that instructors can write
             # there.
@@ -191,5 +180,5 @@ class Exchange(LoggingConfigurable):
             # so we have to create and then chmod.
             os.chmod(path, mode)
         else:
-            if not self.groupshared and not self_owned(path):
+            if not self.coursedir.groupshared and not self_owned(path):
                 self.fail("You don't own the directory: {}".format(path))
--- a/nbgrader/exchange/exchange.py
+++ b/nbgrader/exchange/exchange.py
@@ -137,6 +137,17 @@ class Exchange(LoggingConfigurable):
                                                include=self.coursedir.include,
                                                max_file_size=self.coursedir.max_file_size,
                                                log=self.log))
+        # copytree copies access mode too - so we must add go+rw back to it if
+        # we are in groupshared.
+        if self.groupshared:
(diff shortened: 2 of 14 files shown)


rkdarst:
I think this is coming close to completion (I'm finding less things that are missing).  I'd like feedback on what you think.  If it's worth taking it forward, I can ask about some questions relating to internal improvement it needs, and then start fixing docs and tests.

nbgrader/converters/base.py
```diff
@@ -43,9 +43,21 @@ class BaseConverter(LoggingConfigurable):
...
+            default group writeable.  Requires that admins ensure that primary
+            groups are correct!
+            """
+        )
+    ).tag(config=True)
+
     @default("permissions")
     def _permissions_default(self):
-        return 444
+        return 664 if self.groupshared else 644
```
jhamrick: So I think that the permissions should still be 444 when running the autograder, which is what this line does (it looks like for `assign` and `feedback`, it gets overwritten to 644). The reasoning behind that is to prevent instructors from modifying the autograded notebooks (mostly to prevent confusion, so students don't get back a notebook that is different from the one they submitted). So I'm fine changing the default here but we should also change it to 444 in the autograder then.

nbgrader/converters/base.py
```diff
@@ -43,9 +43,21 @@ class BaseConverter(LoggingConfigurable):
         )
     ).tag(config=True)
 
+    groupshared = Bool(
```
jhamrick: I think this should probably be an attribute of CourseDirectory, and then accessed via `self.coursedir.groupshared`, so that the config value only needs to be set once.

nbgrader/exchange/exchange.py
```diff
@@ -76,6 +76,18 @@ def _cache_default(self):
 
     coursedir = Instance(CourseDirectory, allow_none=True)
 
+    groupshared = Bool(
```
jhamrick: I think this should probably be an attribute of CourseDirectory, and then accessed via `self.coursedir.groupshared`, so that the config value only needs to be set once.

nbgrader/converters/base.py
```diff
@@ -43,9 +43,21 @@ class BaseConverter(LoggingConfigurable):
...
+        False,
+        help=dedent(
+            """
+            Be less strict about user permissions (instructor files are by
+            default group writeable.  Requires that admins ensure that primary
```
jhamrick: Missing a paren

nbgrader/converters/base.py
```diff
@@ -248,6 +260,24 @@ def set_permissions(self, assignment_id, student_id):
         for dirname, _, filenames in os.walk(dest):
             for filename in filenames:
                 os.chmod(os.path.join(dirname, filename), permissions)
+            # If groupshared, set dir permissions - see comment below.
+            if self.groupshared and os.stat(dirname).st_uid == os.getuid():
+                #for subdirname in subdirnames:
```
jhamrick: Remove commented code

rkdarst:
I said I would comment on how my PRs went...

This has been working so far, it seems, because there haven't been any complaints about wrong permissions lately and files seem OK.  Can't promise I got everything, though.

Things that are still needed:
- [ ] documentation
- [ ] make only one `groupshared` option, right now you have to set it twice on two different casses - I'm not sure how to do this.
- [ ] tests?  It will require a new type of test I guess.
- [ ] Check if any of the logic can be simplified, especially the inline ternary operators

Currently, I will wait for a detailed review before starting work.  But I can't promise to get back to it soon, so if someone else take it and fixes stuff, please go ahead.  However, some of these things I could not do quickly so I encourage help!

nbgrader/converters/base.py
```diff
@@ -248,6 +260,24 @@ def set_permissions(self, assignment_id, student_id):
...
+                #    subdirname = os.path.join(dirname, subdirname)
+                try:   os.chmod(dirname, (os.stat(dirname).st_mode|0o2770) & 0o2777)
```
jhamrick: This is a style nitpick, but would you mind formatting the try/except statements so they are across 4+ lines, like:

```
try:
    ...
except PermissionError:
    ...
```

so that it matches the style of the rest of the code?

nbgrader/converters/base.py
```diff
@@ -248,6 +260,24 @@ def set_permissions(self, assignment_id, student_id):
...
+        # If groupshared, set write permissions on directories.  Directories
+        # are created within ipython_genutils.path.ensure_dir_exists via
+        # nbconvert.writer, (unless there are supplementary files) with a
+        # default mode of 755 and there is no way to pass the mode arguments
+        # all the way to there!  So we have to walk and fix.
+        if self.groupshared:
+            # Root may be created in this step, and is not included above.
+            rootdir = self.coursedir.format_path(self._output_directory, '.', '.')
+            # Add 2770 to existing dir permissions (don't unconditionally override)
+            try:   os.chmod(rootdir, (os.stat(rootdir).st_mode|0o2770) & 0o2777)
```
jhamrick: Please reformat try/except

nbgrader/exchange/exchange.py
```diff
@@ -121,12 +133,28 @@ def copy_files(self):
...
+        # copytree copies access mode too - so we must add go+rw back to it if
+        # we are in groupshared.
+        if self.groupshared:
+            for dirname, _, filenames in os.walk(dest):
+                # dirs become ug+rwx
+                try:   os.chmod(dirname, (os.stat(dirname).st_mode|0o2770) & 0o2777)
+                except PermissionError: pass
+                for filename in filenames:
+                    filename = os.path.join(dirname, filename)
+                    try:    os.chmod(filename, (os.stat(filename).st_mode|0o660) & 0o777)
```
jhamrick: Please reformat try/except

nbgrader/exchange/exchange.py
```diff
@@ -121,12 +133,28 @@ def copy_files(self):
...
+                    except PermissionError: pass
+
 
     def start(self):
         if sys.platform == 'win32':
             self.fail("Sorry, the exchange is not available on Windows.")
 
-        self.ensure_root()
+        if not self.groupshared:
```
jhamrick: Hmm, I think this check is mainly to ensure that the current user can access the exchange, so I think it is probably still worthwhile to keep this check in even when self.groupshared is set. Unless there's some reason why this check doesn't work that I'm missing?

jhamrick:
Oh, hah, we both commented at the same time :)

> This has been working so far, it seems, because there haven't been any complaints about wrong permissions lately and files seem OK. Can't promise I got everything, though.

Awesome---having use cases in an actual course is my favorite way to have nbgrader be tested, so if it seems to be working that's great.

> documentation

I think as a start you could just write up the details of your particular use case and how you're using the option. If you don't have time to write official documentation right away what would be helpful would be to open a new issue with detailed information and then I or you, or someone else can later add it to the official docs.

>  make only one groupshared option, right now you have to set it twice on two different casses - I'm not sure how to do this.

See my review comments.

>  tests? It will require a new type of test I guess.

Yeah... the current tests don't really check permissions either because the require having different user accounts which I couldn't figure out how to do on Travis. Even though I'd prefer to have tests I'm not sure what the best way to test it is. Maybe at the very least just copy the existing tests and run them with the groupshared option just to make sure stuff doesn't break.

>  Check if any of the logic can be simplified, especially the inline ternary operators

I'll have another look with respect to this.

nbgrader/converters/base.py
```diff
@@ -248,6 +260,24 @@ def set_permissions(self, assignment_id, student_id):
         for dirname, _, filenames in os.walk(dest):
             for filename in filenames:
                 os.chmod(os.path.join(dirname, filename), permissions)
+            # If groupshared, set dir permissions - see comment below.
+            if self.groupshared and os.stat(dirname).st_uid == os.getuid():
+                #for subdirname in subdirnames:
+                #    subdirname = os.path.join(dirname, subdirname)
+                try:   os.chmod(dirname, (os.stat(dirname).st_mode|0o2770) & 0o2777)
+                except PermissionError: pass
```
jhamrick: Should these `pass` statements on the PermissionError raise an error? Presumably they  happen if the user doesn't actually  have permission to set this, which means they shouldn't be using nbgrader either maybe?

rkdarst:
Big picture comment:

> Overall I think this looks good! I just want to try to understand the use case a bit better, though---is the idea that you have multiple instructors each with their own course directory, or who share a course directory? I guess if it's the latter, then you are not using the formgrader web interface since that requires that the user's notebook can access the files, which it wouldn't if those files are in another user's directory?

I guess this is the hardest thing to convey, since it's quite different from what other poeple do.  I control the whole system: user accounts, filesystem mounts, users, and groups.  An artifical example (in reality done with kubernetes mounts and uid/gid setting):
```
addgroup instructors
adduser inst1 instructors
adduser inst2 instructors
mkdir /coursedir/
chgrp instructors /coursedir/
chmod 2770 /coursedir/
```

So, there is one course directory shared between two instructors (the second option).  Because it is mode `g+rwX`, both people can access.  Formgrader Just Worked(tm) with this setup.  Because it is `g+s`, files inside automatically get `g+w` so both instructors can do course management.

Some of the changes to g+w were so that if one instructor does something, the other can come back and (auto)overwrite/delete it to run the same task again.  There were other small implicit assumptions which needed updating.

jhamrick:
Ok, got it. And when you say the formgrader just worked, were you running a separate notebook service with the formgrader from the course directory, that then both instructors had access to? Or did each instructor separately run a formgrader instance?

rkdarst:
On Sun, Oct 07, 2018 at 08:30:59AM -0700, Jessica B. Hamrick wrote:
> Ok, got it. And when you say the formgrader just worked, were you running a separate notebook service with the formgrader from the course directory, that then both instructors had access to? Or did each instructor separately run a formgrader instance?

We are using JupyterHub with no "shared services" like I recently saw
mentioned.  I think this counts as different formgrader instances for
each instructor, though with the same configuration.  I never realized
that formgrader might be a problem. (only types of problems were
e.g. gradebook.db created without g+w)

nbgrader/converters/base.py
rkdarst: I think one consideration was: if instructor1 autogrades, then it creates some files which instructor2 can't remove.  So no one else can autograde an updated assignment.  (This was a problem somewhere, but I may be remembering something wrong).

jhamrick: Oh, hmm, I see. I think though that if the file has the right group ownership, then other people in the group should be able to remove the file, even if write permissions are disabled. At least, you do see a similar issue with single instructors---they can't directly remove the file but because they own it they can change the permissions and then remove (and I believe this happens in nbgrader somewhere under the hood if you use --force). So I would expect you could do something similar with group-level permissions, though I haven't tested it.

jhamrick:
Huh, I am surprised that works given #1012. I'm curious what the nbgrader_config.py looks like for your instructor accounts?

nbgrader/exchange/exchange.py
rkdarst: My commit message was `groupshared: don't force exchange root +w for everyone`.  I can't seem to remember the exact reason, but... maybe it's because now that instructors have a special group, we don't need to be g+w for everyone.  But this doesn't necessarily seem specific to groupshared.

One other thing I remembered was tring to reset permissions or force-create the directory.  But, users can't chmod dirs owned by others, so it fails.  This may be realated to me not setting others+write on the top-level exchange directory.

I should think about this more and see if it's really needed.  I probably have to remove it and see what breaks!

rkdarst:
The `nbgrader_config.py` files are fun:

```
c = get_config()
c.CourseDirectory.root = "/course"
c.Exchange.course_id = "testcourse"
c.Exchange.multiuser = True
c.Exchange.groupshared = True
c.BaseConverter.groupshared = True
c.AssignmentList.assignment_dir = "/notebooks/"
c.Exchange.assignment_dir = "/notebooks/"                    # I think this is duplicated for backwards compatibilty with a previous version.  Or maybe not...
c.ExecutePreprocessor.timeout = 120
c.NbGrader.logfile = "/course/.nbgraber.log"
```

I give students an instructors access to the whole filesystem: `notebook_dir="/".  Per-user notebooks are mounted in `/notebooks`, and course dir is at `/course`.   `default_url` is set so that you start in `notebooks` if not an insturctor, `/course` if an instructor.  There is also an optional `/coursedata` directory, which people should be able to access.

I couldn't mount the `course` and `coursedata` directories as subpaths of `/notebooks` because of the way NFS works.  I have `root_squash` on, so I couldn't get kubernetes to mkdir/mount within the user-only `/notebooks` directory.  (Student directories have their real university UID, because they can also mount them via SMB.  So I can't make student directories world-read/write).

jhamrick:
Ahh, ok, I think I understand now. Thanks for all the clarifications!

I think the reason why it works for you is because you set notebook_dir to be the root of the filesystem---that way all the course directories are still under the root of the notebook server. I believe your setup wouldn't work if that weren't the case (as in #1012 ). I think it's generally recommended to not run the notebook server from the root of the filesystem, for security reasons, but since you're working with kubernetes I guess it's not so much a problem there.

Anyway, I undestand better now what your usecase is and I think that's definitely fine. We will just need to make sure we document this usecase so people understand what the different options are for. This could potentially go somewhere around here: https://nbgrader.readthedocs.io/en/master/configuration/jupyterhub_config.html#example-use-case-one-class-multiple-graders as an alternate way to setup a class with multiple graders.

nbgrader/converters/base.py
rkdarst: I'll have to try without to see what the problem was.  It could relate to subdirs (if you can write dir, you can remove file, but if you can't write subdir, you can't delete stuff within to get an empty dir to remove)...

I'll note to come back to this.

rkdarst:
The `nbgrader_config.py` files are fun:

```
c = get_config()
c.CourseDirectory.root = "/course"
c.Exchange.course_id = "testcourse"
c.Exchange.multiuser = True
c.Exchange.groupshared = True
c.BaseConverter.groupshared = True
c.Exchange.assignment_dir = "/notebooks/"
c.AssignmentList.assignment_dir = "/notebooks/"
c.ExecutePreprocessor.timeout = 120
c.NbGrader.logfile = "/course/.nbgraber.log"
```

I give students an instructors access to the whole filesystem: `notebook_dir="/".  Per-user notebooks are mounted in `/notebooks`, and course dir is at `/course`.   `default_url` is set so that you start in `notebooks` if not an insturctor, `/course` if an instructor.  There is also an optional `/coursedata` directory, which people should be able to access.

I couldn't mount the `course` and `coursedata` directories as subpaths of `/notebooks` because of the way NFS works.  I have `root_squash` on, so I couldn't get kubernetes to mkdir/mount within the user-only `/notebooks` directory.  (Student directories have their real university UID, because they can also mount them via SMB.  So I can't make student directories world-read/write).

rkdarst:
I've just tried to correct the easy problems, but I am sure there is more.  I am about to run out of time now, but once you give me a go-ahead I will rebase to new master with working tests.

One could ask "why am I going through all this?"  I wanted to start with the vision of a shared unix system and make Jupyter fit to that, rather than the idea of a typical limited Jupyter setup.  And for this, nbgrader is a side-feature.  I think the result has been good.

jhamrick:
> once you give me a go-ahead I will rebase to new master with working tests.

👍 

(though FYI tests are still a bit flaky on master, sigh...)

bbhopesh:
HI @rkdarst. I have exactly same use case where I have one gid for instructors group. Thanks a lot for doing this. 
I was trying to work around this problem by setting configuration `c.BaseConverter.permissions=660`. This was not perfect though as this only affects `.ipynb` files and not intermediate directories like `release/{assignment_id}`. Your code will be released with `0.6.0`. In the meanwhile, what do you think will be best workaround for me to set correct permissions on `release/{assignment_id}`

bbhopesh:
HI @jhamrick . I ran into the same problem as #1012 . My work around was bit different from that of @rkdarst. I(the admin) creates a symlink to the shared nbgrader root directory in user's home and add this line to `nbgrader_config.py` `c.CourseDirectory.root = os.path.expanduser('~/symlink_to_ngroot')`

nbgrader/converters/base.py
rkdarst: done, thanks.

nbgrader/exchange/exchange.py
rkdarst: done, thanks

jhamrick:
Hi @rkdarst , I just wanted to check in on this one too 🙂and to discuss a little bit about the way forward here. I would still be open to having this functionality---allowing the files to be shared via group id---but I do want to point out we are (almost) finished with much better support for multiple graders and classes via a different model (which is a set of shared JupyterHub services running the formgrader). So my questions are:

1. Do you think you'll be able to finish up this PR by rebasing it to master and ensuring the tests pass (and maybe ideally adding a bit of documentation too)?
2. Do you think the other model that we've been working with (via shared JupyterHub services) would work for you? I realize it would probably require a lot of work for you to change to that since you have everything set up quite differently, but I'm curious if there are fundamental things that might still be missing that wouldn't work for your use case.

rkdarst:
I was also thinking about this over the last few days.  I went from still thinking it was necessary, to maybe not, to yes.  I don't know what the real story is.  My current thinking (I'm sorry if this is a bit rambling):

This does two things
(1) Don't unconditionally set permissions for course directory (could be replaced with respecting umask)
(2) Make exchange directories `g+rwx` in addition to `u+rwx`

It seems the only supported way of having multiple graders is a JH-level type service.  This is basically like another user account, and to me adding a user account to share files doesn't seem very elegant.  If multiple people need access to something, I'd rather just share it directly.  For cases where all single-user servers run as the same UID, one can basically just mount the coursedir inside wherever it is needed (this requires custom pre_spawn_hook editing though).  In practice, the fact that one notebook-server = 1 formgrader-course limits things and thus the services are a nice way of handling things.

In my case, I'm starting to see a value in jupyterhub services.  It solves the problem of having to respawn the notebook-server if you change course, which would be a big plus for me, and I need to look into that.  But then I realized that (1) is still needed: All of my users can access data outside the hub.  Services would run under some fake-UID and write files that real-UIDs would need to access by some other method (NFS).  The way to do this is to make the files `g+w`.   (2) may not be needed if only fake-UID interacts with the exchange.

`g+w` would even be useful in the case that you do use a `formgrader-$courseid` group to specify who has access to which courses' formgraders, if these are real users with real access to the course files. (example: students/instructors also have ssh access to the server hosting the data, and the `formgrader-*` group means something).

So in short, all this does is make sure that all files are made `g+w` where they would otherwise be only `u+w`.  My current thought is it could be a useful compliment to the recently added support.

(and now that I think about it, all instructor files are made `g+r` by default anyway... since these would normally be considered confidential, there's a practical assumption that anyone who can access the files is an instructor anyway!  Making them `g+w` maybe isn't that big of a stretch...)

It's late here so maybe I'll be more coherent tomorrow...

rkdarst:
It looks like tests pass (should always be the case if default is no difference).  Code coverage fails.

I guess writing a few basic tests to exercise the code shouldn't be too hard.  Do you have suggestions on the best place to put them is?

It should affect:
- nbgrader assign
- nbgrader release
- anything else that creates new files in instructor directory.
There are obvious places in `tests/apps` for the first two... maybe that's enough?

jhamrick:
> It seems the only supported way of having multiple graders is a JH-level type service. This is basically like another user account, and to me adding a user account to share files doesn't seem very elegant.

Yes, currently that's correct. I agree it's not very elegant in the context of a system where you can have a shared regular filesystem, but I think it makes sense in cases where for example you can't easily run a NFS server. In these cases, it would simplify things a lot to just allow files to be managed through JupyterHub rather than through the filesystem. (I think I would personally find it would be easier even if you can run without NFS, even if it's in theory possible, as it's one more thing to setup and configure). But in another sense having shared files across user accounts does make things more flexible, allowing you to do things like more easily share large data stores, etc. So there are definitely pros and cons.

(I realize of course that the shared filesystem is still needed for nbgrader because of the way the exchange directory works, but in the future we are going to abstract that away too so you won't need a shared filesystem at all, unless you need to do fancier mounting of external data, etc.)

> It solves the problem of having to respawn the notebook-server if you change course

Ah yeah, it didn't quite occur to me that you're having to do that. The services model was definitely intended to be a way to deal with that sort of thing.

> g+w would even be useful in the case that you do use a formgrader-$courseid group to specify who has access to which courses' formgraders, if these are real users with real access to the course files. (example: students/instructors also have ssh access to the server hosting the data, and the formgrader-* group means something).

Hmm, I hadn't thought of that. Ok, thanks, this helps clarify things for me and I agree this would be nice to have.

> My current thought is it could be a useful compliment to the recently added support.

Yes, I agree and it's not a hugely major change so I am happy to accept it 🙂

nbgrader/exchange/exchange.py
```diff
@@ -129,12 +129,31 @@ def do_copy(self, src, dest, log=None):
...
     def start(self):
         if sys.platform == 'win32':
             self.fail("Sorry, the exchange is not available on Windows.")
 
-        self.ensure_root()
+        if not self.coursedir.groupshared:
+            # This just makes sure that directory is o+rwx.  In group shared
+            # case, it is up to admins to ensure that instructors can write
+            # there.
+            self.ensure_root()
```
jhamrick: Maybe it would be nice in the groupshared case if the permissions are not set up properly to at least print a warning/error rather than failing with a traceback later (which is what I assume happens if the permissions aren't correct).

nbgrader/converters/base.py
jhamrick: Reminder to remove this commented code 🙂

jhamrick:
Also just as a heads up, #1120 introduces some new commands that may also need to have some changes made to them for the groupshared functionality. Would you mind if I wait until that one is merged, and then you could make the last few changes for that too before I merge this one?

rkdarst:
Thanks for the reviews and comments!  This helps for sure.

> Also just as a heads up, #1120 introduces some new commands that may also need to have some changes made to them for the groupshared functionality. Would you mind if I wait until that one is merged, and then you could make the last few changes for that too before I merge this one?

That's a good idea... I'm in no rush here.

I'll look for all uses of set_perms... is there anything else that
should be grepped and handled?

nbgrader/converters/base.py
```diff
@@ -248,6 +248,28 @@ def set_permissions(self, assignment_id, student_id):
...
+        # default mode of 755 and there is no way to pass the mode arguments
+        # all the way to there!  So we have to walk and fix.
+        if self.coursedir.groupshared:
+            # Root may be created in this step, and is not included above.
+            rootdir = self.coursedir.format_path(self._output_directory, '.', '.')
+            # Add 2770 to existing dir permissions (don't unconditionally override)
+            try:
+                os.chmod(rootdir, (os.stat(rootdir).st_mode|0o2770) & 0o2777)
+            except PermissionError:
+                pass
```
jhamrick: If the permissions can't be set, wouldn't it be worthwhile to throw an error?

nbgrader/exchange/exchange.py
```diff
@@ -129,12 +129,31 @@ def do_copy(self, src, dest, log=None):
...
                                                log=self.log))
+        # copytree copies access mode too - so we must add go+rw back to it if
+        # we are in groupshared.
+        if self.coursedir.groupshared:
+            for dirname, _, filenames in os.walk(dest):
+                # dirs become ug+rwx
+                try:
+                    os.chmod(dirname, (os.stat(dirname).st_mode|0o2770) & 0o2777)
+                except PermissionError:
+                    pass
```
jhamrick: Same here

nbgrader/exchange/exchange.py
```diff
@@ -129,12 +129,31 @@ def do_copy(self, src, dest, log=None):
...
+                for filename in filenames:
+                    filename = os.path.join(dirname, filename)
+                    try:
+                        os.chmod(filename, (os.stat(filename).st_mode|0o660) & 0o777)
+                    except PermissionError:
+                        pass
```
jhamrick: Same here

jhamrick:
> I'll look for all uses of set_perms... is there anything else that should be grepped and handled?

I think that should get you all the cases with `set_perms` and `ensure_directory` in the exchange apps, and then I would say for the converters (assign, autograde, etc.) just check those permissions are correct after the fact too. I think those two things should give you much better test coverage!

nbgrader/converters/base.py
rkdarst: For this (and all other permission errors / `ensure_root` comments), what I was thinking was that only the owner can change permissions - so any `chmod` call will always fail even if it is OK.  Since we are expecting this to be run by multiple users, failures will be the rule rather than the exception.

My thought for a proper solution now: add a test if permissions are OK and only try to `chmod` if needed.  If chmod is not possible, log a warning.

Another option is to add in some code comments / docs in the option explaining this and ignoring the problem for now.

jhamrick: Oops, sorry I forgot to reply to this earlier. I see, that reasoning makes sense.

> My thought for a proper solution now: add a test if permissions are OK and only try to chmod if needed. If chmod is not possible, log a warning.

I think I would prefer that solution, I think silently hiding errors is always a bit dangerous 🙂

jhamrick:
Ok, now that #1120 is merged, I think you can do a rebase and then add the final set of changes 😄 There is now a "release_feedback" command which I think you'll need to update with the groupshared options as well.

nbgrader/converters/base.py
rkdarst: I wasn't sure what to do, because it would fail in the common case when owner != current user.  In the end, I stat the file first, see if the mode needs to be changed, then try to change it - so errors will only come when it can't be changed _and_ it is actually a problem.  We had to stat everything anyway, so no major loss.

So I think these should be fixed now.

nbgrader/exchange/exchange.py
rkdarst: This checks if the exchange directory is writeable (even when run as a student).  But students only need access to write to {root}/inbound.  So the real issue isn't related to groupshared, it's just that groupshared causes me be a bit more precise, so this got thrown in here.

What do you think we should do?  I can think of at least: a) leave it as is for now, b) make check_exchange_root a configuration option that can be disabled, c) nothing, I make it a separate local change I maintain separately, or d) I allow the root to be writeable by students.

nbgrader/converters/base.py
rkdarst: Solved/commented on down below.  Now is more clever about changing permissions and logs a message.

nbgrader/exchange/exchange.py
rkdarst: OK, this is the same comment as below (https://github.com/jupyter/nbgrader/pull/1000#discussion_r305368967).  Discussion about it is written there.  I'll leave this sub-conversation open but do discussion down there.

rkdarst:
OK, rebased to current master and checked all the pieces.  But, still no automated tests and no manual tests from anyone other than me.

Most issues are resolved, with the exception of one related issue above.  I'll probably move it to a different branch, but will wait for your opinion anyway.

rkdarst:
I've now added tests for:
* generate assignment
* release assignment
* collect
* autograde
* generate feedback
* release feedback

Push incoming once they pass locally.  Then... I think almost everything has been resolved for this.

nbgrader/exchange/exchange.py
rkdarst: I propose (a), it's least invasive and has least risk, and if someone is doing groupshared stuff, they are probably able to debug permission errors pretty easily.

rkdarst:
Took some time to debug windows builds, but it's done now.

Looking at the CI output, the failure is codecov - but basically every line is hit except exception handling, which requires there to be a wrong UID owning a file so that `os.chmod` fails.  This is such a corner case and difficult to test (would require OS switching users or emulating that) that it can probably be left out.  Am I interpreting this correctly?

Then the remaining changes requested are discussed here: https://github.com/jupyter/nbgrader/pull/1000#pullrequestreview-244618610 .  What do you think should be done?

nbgrader/exchange/exchange.py
jhamrick: Oh oops, I missed this comment before I merged. But yes, I agree with your assessment, I think it's fine to leave it as it is for now.

rkdarst:
Thanks!  This was quite a huge process, thanks for the review and hints.

Next up is documentation.  Would you like this option to be somewhat hidden or given extensive documentation?  I can think to add it to one of:
- only under "configuration options" with an more extensive help string that serves as the full documentation of how to use it. This would make it only appear under "Configuration options" (relatively hidden).
- under the "Advanced topic" section (promoted).

Of course whatever the case, it will be clearly indicated what the conditions and risks are.

jhamrick:
> Thanks! This was quite a huge process, thanks for the review and hints.

Thank you for all the effort you put into getting this in! (And sorry it took so long on my end)

> Next up is documentation. Would you like this option to be somewhat hidden or given extensive documentation?

I think it is probably fine to put it under the "Advanced" section so we can give a bit more extensive documentation, but I would maybe explicitly add a warning using one of the `.. warning` sphinx directives saying this feature is more experimental and that we are less likely to be able to provide help on debugging if someone can't get it to work.

PR #1001, opened 2018-08-05
Allow setting a different assignment dir for students than the root notebook directory

rkdarst:
Currently, nbgrader assumes that fetching assignments via the web UI will always put them in `notebook_dir` - so that `notebook_dir`/`default_url` can't be used.  This PR seem to minimally fix that - at least I could do one manual test and it worked.  It will need serious examination from others, in addition to tests and documentation.

Please let me know what you think: is this on the right track?  What should be done to make it useful?

---
I'm not sure if the `.tag(config=True)` is needed.

The current config needs `c.Exchange.assignment_dir = "/notebooks/` in `nbgrader_config.py` and `c.AssignmentList.assignment_dir` in `jupyter_notebook_config.py`.

### DIFF
--- a/nbgrader/exchange/exchange.py
+++ b/nbgrader/exchange/exchange.py
@@ -31,6 +31,17 @@ class Exchange(LoggingConfigurable):
         )
     ).tag(config=True)
 
+    assignment_dir = Unicode(
+        ".",
+        help=dedent(
+            """
+            Local path for storing student assignments.  Defaults to '.'
+            which is normally Jupyter's notebook_dir.
+            """
+        )
+    ).tag(config=True)
+
+
     @validate('course_id')
     def _validate_course_id(self, proposal):
         if proposal['value'].strip() != proposal['value']:
--- a/nbgrader/exchange/fetch.py
+++ b/nbgrader/exchange/fetch.py
@@ -30,7 +30,7 @@ class ExchangeFetch(Exchange):
             root = os.path.join(self.course_id, self.coursedir.assignment_id)
         else:
             root = self.coursedir.assignment_id
-        self.dest_path = os.path.abspath(os.path.join('.', root))
+        self.dest_path = os.path.abspath(os.path.join(self.assignment_dir, root))
         if os.path.isdir(self.dest_path) and not self.replace_missing_files:
             self.fail("You already have a copy of the assignment in this directory: {}".format(root))
 
--- a/nbgrader/server_extensions/assignment_list/handlers.py
+++ b/nbgrader/server_extensions/assignment_list/handlers.py
@@ -74,9 +74,9 @@ class AssignmentList(LoggingConfigurable):
             else:
                 for assignment in assignments:
                     if assignment['status'] == 'fetched':
-                        assignment['path'] = os.path.relpath(assignment['path'], self.assignment_dir)
+                        assignment['path'] = os.path.relpath(assignment['path'], self.parent.notebook_dir)
                         for notebook in assignment['notebooks']:
-                            notebook['path'] = os.path.relpath(notebook['path'], self.assignment_dir)
+                            notebook['path'] = os.path.relpath(notebook['path'], self.parent.notebook_dir)
                 retvalue = {
                     "success": True,
                     "value": sorted(assignments, key=lambda x: (x['course_id'], x['assignment_id']))
--- a/nbgrader/server_extensions/assignment_list/handlers.py
+++ b/nbgrader/server_extensions/assignment_list/handlers.py
@@ -53,7 +53,9 @@ class AssignmentList(LoggingConfigurable):
         # first get the exchange assignment directory
         with chdir(self.parent.notebook_dir):
             config = self.load_config()
-            assignment_dir = config.Exchange.assignment_dir
+
+        lister = ExchangeList(config=config)
+        assignment_dir = lister.assignment_dir
 
         # now cd to the full assignment directory and load the config again
         with chdir(assignment_dir):
--- a/nbgrader/server_extensions/assignment_list/handlers.py
+++ b/nbgrader/server_extensions/assignment_list/handlers.py
@@ -33,7 +33,7 @@ def chdir(dirname):
 
 class AssignmentList(LoggingConfigurable):
 
-    assignment_dir = Unicode('', help='Directory where the nbgrader commands should be run')
+    assignment_dir = Unicode('', help='Directory where the nbgrader commands should be run').tag(config=True)
     @default("assignment_dir")
     def _assignment_dir_default(self):
         return self.parent.notebook_dir
--- a/nbgrader/server_extensions/assignment_list/handlers.py
+++ b/nbgrader/server_extensions/assignment_list/handlers.py
@@ -33,11 +33,6 @@ def chdir(dirname):
 
 class AssignmentList(LoggingConfigurable):
 
-    assignment_dir = Unicode('', help='Directory where the nbgrader commands should be run').tag(config=True)
-    @default("assignment_dir")
-    def _assignment_dir_default(self):
-        return self.parent.notebook_dir
-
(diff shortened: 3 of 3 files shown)


rkdarst:
Checking this again... it seems to have worked in practice.  Remaining considerations:

- [ ] Undocumented
- [ ] I can't promise that it covers all of the different places it needs.  Someone with more inside knowledge should probably look and check everything...
- [ ] untested - and I guess that making tests will be a bit delicate, which I probably can't do in a reasonable time.

rkdarst:
For others reading, this explains what I was trying to accomplish:  https://github.com/jupyter/nbgrader/pull/1000#issuecomment-427664752

@jhamrick, thanks for the hints.  I'll try to take a look.

I think I modified `handlers.py` because the AssignmentList add-on was trying to open the notebooks in the wrong place.  As you can guess, this was quite exploratory development.

jhamrick:
Hi @rkdarst , just wanted to check in to see if you'd had more time to work on this? If not, I might be able to implement something similar this weekend.

rkdarst:
I did work on it a bit, using your suggestions (I think that's the new commits since then).  As far as I know it works and I've been using it successfully since then.  I don't know of any problems with it, but I also don't know enough to know what other problems there could be.  It's also has no tests or docs.

I'll rebase it to current master, that will be the next push.  Then you can look and see what you think.  If it's something you would like, I can see about any finalizations... do you know of how to reasonably make tests for this?

jhamrick:
Ah, great! I'll have a look then and test it out. I think this would be a useful thing to have, so I am definitely happy to accept it.

I'll give another comment in a bit once I've had to look through more closely and think about the tests.

jhamrick:
So my first main comment is that I'd prefer to not have two separate configurables for `assignment_dir` as I think that will make the configuration a bit confusing. But here is a patch that I think should allow you to just use the configurable from the exchange. It changes the behavior to (1) first look for an nbgrader_config.py in the directory where the notebook is running, from which it loads the `assignment_dir` path. Then, it switches to that directory and reloads `nbgrader_config.py`, and uses that directory for the actual listing/fetching/submitting.

I have only tested this minimally so far, but will test more later. In terms of tests, it is indeed tricky and not very straightforward as it would require re-launching the notebook server with a different configuration. As much as it pains me to say so, for this it might be easier not to include tests for the time being (there are also currently no tests for `path_includes_course` either) and open an issue to figure out a better way to run tests for all of these various options going forward.

```
From 145f332d92a0a011465c1577e61eb438c4ce502b Mon Sep 17 00:00:00 2001
From: "Jessica B. Hamrick" <jessica.b.hamrick@gmail.com>
Date: Sat, 1 Jun 2019 16:37:03 +0100
Subject: [PATCH] Simplify logic to have only one assignment_dir configurable

---
 .../assignment_list/handlers.py               | 30 +++++++++++--------
 1 file changed, 17 insertions(+), 13 deletions(-)

diff --git a/nbgrader/server_extensions/assignment_list/handlers.py b/nbgrader/server_extensions/assignment_list/handlers.py
index 34f6da95..22e0c908 100644
--- a/nbgrader/server_extensions/assignment_list/handlers.py
+++ b/nbgrader/server_extensions/assignment_list/handlers.py
@@ -33,11 +33,6 @@ def chdir(dirname):
 
 class AssignmentList(LoggingConfigurable):
 
-    assignment_dir = Unicode('', help='Directory where the nbgrader commands should be run').tag(config=True)
-    @default("assignment_dir")
-    def _assignment_dir_default(self):
-        return self.parent.notebook_dir
-
     def load_config(self):
         paths = jupyter_config_path()
         paths.insert(0, os.getcwd())
@@ -53,10 +48,22 @@ class AssignmentList(LoggingConfigurable):
 
         return full_config
 
+    @contextlib.contextmanager
+    def get_assignment_dir_config(self):
+        # first get the exchange assignment directory
+        with chdir(self.parent.notebook_dir):
+            config = self.load_config()
+            assignment_dir = config.Exchange.assignment_dir
+
+        # now cd to the full assignment directory and load the config again
+        with chdir(assignment_dir):
+            for new_config in NbGrader._load_config_files("nbgrader_config", path=[os.getcwd()], log=self.log):
+                config.merge(new_config)
+            yield config
+
     def list_released_assignments(self, course_id=None):
-        with chdir(self.assignment_dir):
+        with self.get_assignment_dir_config() as config:
             try:
-                config = self.load_config()
                 if course_id:
                     config.Exchange.course_id = course_id
 
@@ -85,9 +92,8 @@ class AssignmentList(LoggingConfigurable):
         return retvalue
 
     def list_submitted_assignments(self, course_id=None):
-        with chdir(self.assignment_dir):
+        with self.get_assignment_dir_config() as config:
             try:
-                config = self.load_config()
                 config.ExchangeList.cached = True
                 if course_id:
                     config.Exchange.course_id = course_id
@@ -140,9 +146,8 @@ class AssignmentList(LoggingConfigurable):
         return retvalue
 
     def fetch_assignment(self, course_id, assignment_id):
-        with chdir(self.assignment_dir):
+        with self.get_assignment_dir_config() as config:
             try:
-                config = self.load_config()
                 config.Exchange.course_id = course_id
                 config.CourseDirectory.assignment_id = assignment_id
 
@@ -165,9 +170,8 @@ class AssignmentList(LoggingConfigurable):
         return retvalue
 
     def submit_assignment(self, course_id, assignment_id):
-        with chdir(self.assignment_dir):
+        with self.get_assignment_dir_config() as config:
             try:
-                config = self.load_config()
                 config.Exchange.course_id = course_id
                 config.CourseDirectory.assignment_id = assignment_id
 
-- 
2.17.2 (Apple Git-113)
```

jhamrick:
I've tested a bit more extensively and it seems to work ok for me, including with the `path_includes_course` setting.

rkdarst:
Thanks for that patch - the multiple `assignment_dir` configs was quite annoying, so I'm glad you could fix it.  I don't fully understand some parts of this, but will it still work if assignment_dir is set in `/etc/jupyter/nbgrader_config.py` ?

Will you apply that patch or me?  If you already made it on your branch, feel free to push it here (assuming it's set so you can, I can't see how to check it)

jhamrick:
> will it still work if assignment_dir is set in /etc/jupyter/nbgrader_config.py

Yes, it should, since that config file can be found from anywhere nbgrader is run!

I can push a commit to your branch.

jhamrick:
Done---let's see how the tests do but if they pass then I'm happy to merge!

jhamrick:
Oops, my patch broke things, sorry about that. Fix incoming...

jhamrick:
Ok, pushed another commit that should fix the tests.

> I don't fully understand some parts of this

So to clarify, what the patch is doing is essentially loading the config twice: once to get the path to the assignment dir, and a second time just in case the assignment dir also includes an nbgrader_config.py file. Then it returns the config it has loaded the second time and uses that in each of the handlers. Does that make it clearer?

rkdarst:
It looks like docs are auto-generated, and this is kind of a special-use case so only needs the auto-generated docs... do you agree?

And now I think I understand what you did.  The confusing part was understanding the chdirs.  Since  I don't have an assignment_dir/notebook_dir configuration file, this wasn't needed.  But it is in the general case, since config files are probably based on assignment_dir.  And all this is needed on the server -side.  The important part is the stuff besides `get_assignment_dir_config` which gets assignments "the right way"...

Thanks for the updates!  Let me know if there is more I should do...

jhamrick:
> It looks like docs are auto-generated, and this is kind of a special-use case so only needs the auto-generated docs... do you agree?

Yes, I think it's fine to just have the autogenerated docs in this case.

> Let me know if there is more I should do

Nope, I think this is ready to merge! Thanks for the PR 😄

rkdarst:
woo, thanks for the help and advice!

PR #1002, opened 2018-08-10
Change to notebook directory when validating (repeat of #880)

rkdarst:
- This was fixed in #880, but a refactoring to split reading from
  preprocessing has removed the wrapper around the preprocess step.

There is in fact a test for this, but perhaps pytest chdirs to the ipynb's directory so that the test can't fail.

I'm still making my final tests but this seems to work.

### DIFF
--- a/nbgrader/tests/apps/files/data.txt
+++ b/nbgrader/tests/apps/files/data.txt
@@ -1,0 +1,4 @@
+line 1
+line 2
+line 3
+line 4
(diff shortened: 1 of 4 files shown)


jhamrick:
Looks good, thanks! Should I merge this or did you still want to add a test?

rkdarst:
It's easy to merge now, but I guess then tests will never come.  I probably won't get around to tests soon, so you can decide what you'd like to do.

I seem to remember checking, and the problem was that pytest directly executes notebooks.  And chdirs to the notebook directory before executing it.  So there would have to be another layer that pytest runs, which runs another the actual test on a notebook in another dir.  (This memory could be wrong).

jhamrick:
I've added a test which is fixed by this change, and will merge once it passes. Thanks!

rkdarst:
Thanks a lot!  Let's hope it works.

Issue #978, opened 2018-06-11
Students able to access others' submissions?


liffiton:
### Operating system
Linux (CentOS 6)

### `nbgrader --version`
`nbgrader version 0.5.4`

### `jupyterhub --version` (if used with JupyterHub)
`0.8.1`

### `jupyter notebook --version` 
`5.5.0`

### Issue

I'm just starting to explore nbgrader, so I may be missing an important detail, but in my testing it looks like a student could access another's submissions.  Submissions appear to be created in `coursedir/inbound` in a directory with a standard name: `USER+ASSIGNMENT+TIMESTAMP`.  The permissions on `inbound` are set such that users other than the creator (whoever runs formgrader, I believe?) cannot list its contents.  However, the permissions of the submission directories are world-readable and -writable.  Given the standard format for submission directory names, couldn't anyone write a simple script or program to search for valid directory names and access whatever submission directories they find?

Now, this isn't a huge issue, because the inclusion of microseconds makes for a fairly large search space.  But on my system, a simple Python program can search a range of one second (one million different directory names) in about six seconds.  If one student knows another has just submitted something (if they work in a shared lab, say), then it wouldn't be too hard to search through a range of a few minutes to find it.  Then they could read and/or alter the other student's work.

(Then there's the fact that if a student figures all of this out and implements it well, they're probably doing fine in the course anyway.  I'm mostly just opening the issue to check my understanding.)

If this is an accurate description of the issue, then I think a simple fix would be to add a random string to each submission directory name.

jhamrick:
Yes, I just tested it and this is in fact correct. This problem will be alleviated when we switch to HubShare (see #659 ) but I think for now it's a reasonable workaround to add a random string to the submission names.

rgerkin:
@jhamrick To have even this level of security, the inbound directory should have 773 permissions (no regular users listing files; submissions that need to be shown in the assignments tab can be read from the database) and the outbound should have 775 (regular users shouldn't be writing random new assignments).  Is that right?

rkdarst:
Exactly.  It should done automatically if nbgrader creates the exchange dirs.  (But it also makes it 733 and 755 by default I think, assuming groups are not for instructors).

jhamrick:
Yes, exactly--and this is indeed something that's set by nbgrader automatically.

PR #1005, opened 2018-08-22
Add random string to submission filenames for better hiding

rkdarst:
- Previously, the exchange was writeable by all users, readable by all
  users, but only listable by the instructors.  Non-guessability of
  the filename is the method of security.  However, filenames are a
  function of (username, assignment_id, and microsecond timestamp).  A
  microsecond timestamp alone is not immediately guessable, but not
  perfect either.
- This patch introduces a random string to the submission filename.
- This is backwards compatible, except if there is a "+" in a username
  and the instructor uses new nbgrader and students submit with old
  nbgrader.  This is because "+" is used to split the path components,
  but is also a valid character for usernames.  It was split with
  .rsplit("+", 2) before, now it is .rsplit("+", 3).
- Related: #978
- Closes: #978

---

The non-extension tests work locally, let's see what travis says for the rest...

### DIFF
--- a/nbgrader/exchange/collect.py
+++ b/nbgrader/exchange/collect.py
@@ -26,8 +26,8 @@ class ExchangeCollect(Exchange):
     def _path_to_record(self, path):
         filename = os.path.split(path)[1]
         # Only split twice on +, giving three components. This allows usernames with +.
-        filename_list = filename.rsplit('+', 2)
-        if len(filename_list) != 3:
+        filename_list = filename.rsplit('+', 3)
+        if len(filename_list) < 3:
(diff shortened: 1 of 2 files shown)


jhamrick:
Awesome, thanks!

rkdarst:
Thanks!

Gourd you check that this filename isn't parsed anywhere else, like
nbgrader list?  Not currently at a computer, but I can't be sure that I got
everything!  We aren't using all possible features, but is seems to have
worked for two rounds so far...

jhamrick:
Oh, you're right, it actually is being used by nbgrader list. I'm going to revert this for now so we can add the relevant changes for that, too.

Issue #1015, opened 2018-09-25
Update documentation links.


Carreau:
Many links in the documentation are 404 (for example) continuum is now anaconda, and make th test fails. 

It should be easy to either update the links, or remove rephrase the corresponding sections.

amrs-tech:
I'll try to sort out this issue. Can you please assign it to me?

amrs-tech:
Can you please mention the file(s) which is to be updated ?

Carreau:
If you look at some of the latest [tests of the Documentation Build](https://travis-ci.org/jupyter/nbgrader/jobs/433069483), you will see near to the end that many links are red with reason as to why. 

I see that someone submitted #1016, so check the latest tests

mathagician:
I would like to pick some. Could please help and suggest how and which one to take.

mddarmawan:
@Carreau me too. Would you please mention me the files which is to be updated?

stvmachine:
@Carreau hi, i would love to help too

jhamrick:
This has been fixed by #1016 and #1017

PR #1016, opened 2018-10-01
Fix anaconda link

swarnava:
Fix of #1015 

I did a quick look and found only anaconda links are broken, Let me know if there are any additional links which are also broken


Carreau:
that's good already.

If you look at the [tests on 2.7](https://travis-ci.org/jupyter/nbgrader/jobs/435629636) you'll see that there are a couple of permanent redirect as well which could also be taken care off.

The test are failing  on other version of Python for unrelated reasons.

Many thanks.

PR #1017, opened 2018-10-01
Fix all redirection

swarnava:
This should cover all #1015 testcases

### DIFF
--- a/nbgrader/docs/source/index.rst
+++ b/nbgrader/docs/source/index.rst
@@ -56,5 +56,5 @@ For an overview and demonstration of nbgrader's core functionality, check out th
     :caption: Resources
 
     Source repository <https://github.com/jupyter/nbgrader>
-    Jupyter education mailing list <http://tinyurl.com/jupyter-education>
+    Jupyter education mailing list <https://groups.google.com/forum/#!forum/jupyter-education>
     changelog
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ Coverage: [![codecov.io](http://codecov.io/github/jupyter/nbgrader/coverage.svg?
 
 A system for assigning and grading Jupyter notebooks.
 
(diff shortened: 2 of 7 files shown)


swarnava:
This should cover all #1015 testcases

nbgrader/docs/source/index.rst
```diff
@@ -56,5 +56,5 @@ For an overview and demonstration of nbgrader's core functionality, check out th
     :caption: Resources
 
     Source repository <https://github.com/jupyter/nbgrader>
-    Jupyter education mailing list <http://tinyurl.com/jupyter-education>
```
npenzin: Do you think it's possible to add some exclusion marks and keep the `tinyurl` version? I guess it is there for a reason like to see the number of people going to the Google Groups from this exact link?

swarnava: Noob question 😅 How do I exclude it? Do you mean should I keep tinyurl link? or is there a way to do that at travis config?

swarnava: @npenzin ?

npenzin: @swarnava Soorry for late reply. I'm afraid I can't help you on that but I guess TravisCI docs might have an answer on that, check it out. Best! :)

jhamrick: I believe the reason this uses tinyurl is because the `#` in the google groups URL was causing problems with the linkchecker. This might not be a problem anymore, in which case it would be fine to put the full URL.

swarnava: Does this mean we are good to merge this PR now ? 😄

jhamrick: Yep I just tested it locally and it seems to be ok. The tests are failing for other reasons which I'll try to investigate separately.

jhamrick:
Thanks!

PR #1019, opened 2018-10-04
Make nbgrader quickstart work with existing directories

### LINKED ISSUE #818: Allow nbgrader quickstart to work on an existing directory
Rather than refusing to run if the directory exists, quickstart should check if it is going to overwrite any files and if it is only then throw an error.

cc @ncclementi

dechristo:
@jhamrick Please let me know if I need to change something.
Thanks! =)

Fixes #818


jhamrick:
Looks great, thanks a lot!

PR #1020, opened 2018-10-06
pull01

vidurathegeek:
To check out the code and help


jhamrick:
This looks like it's a PR from master to the 0.2.x branch, but we shouldn't be adding any more commits to 0.2.x (that's a very old release).

PR #1021, opened 2018-10-06
Fix tests, all of which are failing :(

jhamrick:
- [x] Fix docs tests
- [x] Fix nbextension tests
- [x] Fix python tests


jhamrick:
Meh, the tests passed on this PR but now they're still failing on master. Aargh 😠

PR #1032, opened 2018-10-18
fix: return info of reper function is wrong in api.py

00Kai0:


jhamrick:
Thanks! I think the test failures are unrelated, but I will check more closely tonight just to be sure.

PR #1038, opened 2018-10-30
More language support

sigurdurb:
# Adds octave support, refactors to less code, introduces cutoff

* Introduces utils `is_validate_error(cell, lang = None, cutoff = 100)`. This function refactors the code that was in validator._extract_error and utils.determine_grade, so now they both call this function instead of having the same code in both those places. The cutoff parameter could probably be reduced to 10.

* The optional `cutoff` parameter to is_validate_error. This should not affect grading because the determine_grade function calls is_validate_error with 1 as cutoff to give a zero or max_grade; that is the functionality of the autograder test cells, its also usually enough because the kernels usually stop giving out error messages at the first assertion error. But for example octave will give out message for every error because we are only regex matching the error text so the kernel is not stopping at an error like other kernels would.

* Using `lang = nb.metadata.get('kernelspec',{}).get("language","python")` to get the name of the kernel. Even though it says python it gets the name of the kernel. So 'octave' is returned for the octave kernel for example. This will hopefully introduce more language support. I also think it is safer to get the language then for example having the code that finds octave errors being checked for every language.

* octave kernel error matching now works, I tried some other things that might give the same error like calling display('error ') but it does not conflict. If somebody has some other ideas I could try that, otherwise I think this is a go. Edit: Okei, display('error: ') is a problem but only if they are in a autograder test cell, which are read only for the student, maybe that is safe enough.

### DIFF
--- a/nbgrader/alembic/versions/282d30a0218e_adds_lms_user_id_column_to_student_table.py
+++ b/nbgrader/alembic/versions/282d30a0218e_adds_lms_user_id_column_to_student_table.py
@@ -1,0 +1,24 @@
+"""adds lms_user_id column to student table
+
+Revision ID: 282d30a0218e
+Revises: 724cde206c17
+Create Date: 2018-10-25 20:14:56.887686
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = '282d30a0218e'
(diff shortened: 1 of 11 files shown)


sigurdurb:
Closing because I accidentally used this on top of other code that was not related to this enhancement

PR #1050, opened 2018-12-15
Create 0.5.5 release

jhamrick:
This is a minor release for publishing the JOSE paper on nbgrader: https://github.com/openjournals/jose-reviews/issues/32

cc @jedbrown @willingc @dsblank


dsblank:
Great! Thank you all!

setup.py
```diff
@@ -103,7 +103,8 @@
         "requests",
         "jsonschema",
         "alembic",
-        "ipython<=6.2.1"
+        "ipython<=6.2.1",
```
willingc: FYI, In a requirements.txt file (not sure about in setup.py) you can specify an environment_marker for python version:
```
ipython==5.4.1 ; python_version < "3.0"
ipython>=6.0 ; python_version > "3.3"
```

Even if it doesn't work in setup.py, it may be worth using in requirements.txt while you are debugging the tests

jhamrick: Ah cool, thanks, I didn't realize! Hopefully I've got it sorted now, but I'll try that out if it still doesn't work...

damianavila: @jhamrick can I ask you why you are constraining the ipython version to <=6.2.1? Did you find any issues with ipython 7?

ocefpaf: I'm also curious about this. The pin here is quite restrictive and is preventing some deployments of nbgrader with modern IPython.

jhamrick: I can't remember the reason why I pinned it unfortunately but I think it may have been causing some issues with the tests. (See also #1149)

I'll try to unpin this and create an official 0.5.6 release soon. Thanks @ocefpaf for creating the conda release that does this already!

PR #1053, opened 2018-12-15
Remove changes to sitecustomize.py and dependency on invoke

### LINKED ISSUE #1046: Public side-effects of running tests
Running the test suite currently tries to create/modify a `sitecustomize.py` in a system path. This file is often not owned by the user and in any case, has public side-effects if other python programs run while the tests run. Instead of these manipulations in `tasks.py`, I think it would be preferable to just have a static file in the nbgrader source and add its directory to PYTHONPATH. Do you foresee any problems with this?

Cc: https://github.com/openjournals/jose-reviews/issues/32

jhamrick:
Fixes #1046

cc @jedbrown @willingc

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -32,7 +32,7 @@ def _check_if_directory_in_path(pth, target):
 def docs(args):
     del args  # unused
     if not WINDOWS:
-        run('py.test --nbval-lax --current-env nbgrader/docs/source/user_guide/*.ipynb')
+        run('pytest --nbval-lax --current-env nbgrader/docs/source/user_guide/*.ipynb')
     run('python nbgrader/docs/source/build_docs.py')
     run('python nbgrader/docs/source/clear_docs.py')
     run('make -C nbgrader/docs html')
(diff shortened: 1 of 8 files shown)


jhamrick:
I am not 100% sure the coverage stuff will still work with this, but we'll see once the tests on travis run :)

tasks.py
```diff
@@ -31,54 +29,23 @@ def _check_if_directory_in_path(pth, target):
...
-                """
-                ### begin nbgrader changes
-                import coverage; coverage.process_startup()
-                ### end nbgrader changes
-                """
-            ).lstrip())
-
+def _run_tests(mark, skip, junitxml):
     cmd = []
-    if not WINDOWS:
```
willingc: Not directly related to this PR, but pytest now prefers `pytest` for command execution. Since we have a dependency on a recent version of pytest, we should update: `cmd.append('py.test')`

jhamrick: Ah, I didn't realize that, thanks! I went ahead and made those changes in this PR as well.

PR #1055, opened 2018-12-17
Fix alembic upgrade chain

jhamrick:
This fixes an issue that was introduced by #984 with alembic migrations, and which I didn't think to check for before merging.


jhamrick:
Whoops, turns out I'm mistaken, nevermind.

PR #1061, opened 2019-01-11
Set env var NBGRADER_VALIDATING when validating

rkdarst:
- This allows a notebook to know when it is validating and when it is
  not.
- This was needed for a course with using major computational
  resources in each notebook.  The notebook is tell it is validating
  and go into a validation mode, using only cached data.

====

This came up as a need recently.  Actual use case is training models in the notebook and saving the models to a file, the validation can't re-do that because it takes so long.  For autograding, we can use hidden tests to change behavior, but I didn't find a way to do this when validating.

- Is there a proper way to do this, so that this change isn't needed?
- Untested.  If this is considered useful, I might be able to figure out tests myself.
- making a new context managed seems heavy-handed compared to the minimum change needed.  But I realized I needed to handle errors and slightly preferred a more proper solution.

What do you think?

### DIFF
--- a/nbgrader/utils.py
+++ b/nbgrader/utils.py
@@ -266,6 +266,20 @@ def chdir(dirname):
         os.chdir(currdir)
 
 
+@contextlib.contextmanager
+def setenv(**kwargs):
+    previous_env = { }
+    for key, value in kwargs.items():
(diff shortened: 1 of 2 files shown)


jhamrick:
Sorry to be so behind on PRs---this looks great, thanks!

PR #1064, opened 2019-01-28
Validate all cells

rkdarst:
An instructor asked why the "validate" interface didn't present errors in all cells (instead of just those that were graded).  It seems it was designed this way.

We took a look and figured out how to make it validate everything: we added a `validate_all` option, which makes the validation interface show all errors.  There's no difference in what's actually computed, it just doesn't hide the errors from non-graded cells.

This is currently without tests, though we've been using it for some time.  I can look at tests if the reception is positive.


jhamrick:
I think it is probably just a spurious test failure.

jhamrick:
Thanks @rkdarst !

PR #1071, opened 2019-03-06
student and assignment selection in exportapp implemented

### LINKED ISSUE #799: Allow the exporting of grades per assignment id
The main issue I have with the exporting of grades (csv export) is that it is extremely sluggish exporting the entire gradebook (5min+), especially in cases where there may only be one or two grade changes for a given assignment.

My suggestion is to allow the exporting of grades to individual files per assignment id, if an assignment id is specified, otherwise export the entire gradebook.

A custom plugin can be created to achieve this, but I think it may be better to support this.
@jhamrick what do you think?

This is also something that is specific to the use case of a large number of students.

ddland:
fixes #799 
The student and assignment option can both be a set of values. Making it possible to export almost all students and / or assignments.

### DIFF
--- a/nbgrader/apps/exportapp.py
+++ b/nbgrader/apps/exportapp.py
@@ -38,7 +38,7 @@ class ExportApp(NbGrader):
         You can export the grades for a single (or limited set) of students 
         or assignments with the --assignment and/or --student flag:
 
-            nbgrader export --asignment assignmentID --student [studentID1,studentID2]
+            nbgrader export --assignment assignmentID --student [studentID1,studentID2]
 
         Where the studentIDs are a list of IDs and the assignmentID a string with
         IDs separated by semicolumns.
--- a/nbgrader/apps/exportapp.py
+++ b/nbgrader/apps/exportapp.py
@@ -38,10 +38,12 @@ class ExportApp(NbGrader):
         You can export the grades for a single (or limited set) of students 
         or assignments with the --assignment and/or --student flag:
 
-            nbgrader export --assignment assignmentID --student [studentID1,studentID2]
+            nbgrader export --assignment [assignmentID] 
+                            --student [studentID1,studentID2]
 
-        Where the studentIDs are a list of IDs and the assignmentID a string with
-        IDs separated by semicolumns.
+        Where the studentIDs and assignmentIDs are a list of IDs and 
+        assignments. The assignments or studentIDs need to quoted if they 
+        contain not only numbers. The square brackets are obligatory.
 
         To change the export type, you will need a class that inherits from
         nbgrader.plugins.ExportPlugin. If your exporter is named
--- a/nbgrader/apps/exportapp.py
+++ b/nbgrader/apps/exportapp.py
@@ -9,7 +9,10 @@ aliases = {
     'log-level' : 'Application.log_level',
     'db': 'CourseDirectory.db_url',
     'to' : 'ExportPlugin.to',
-    'exporter': 'ExportApp.plugin_class'
+    'exporter': 'ExportApp.plugin_class',
+    'assignment' : 'ExportPlugin.assignment',
+    'student': 'ExportPlugin.student',
+
 }
 flags = {}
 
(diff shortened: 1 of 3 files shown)


nbgrader/apps/exportapp.py
```diff
@@ -32,6 +35,14 @@ class ExportApp(NbGrader):
 
             nbgrader export --to mygrades.csv
 
+        You can export the grades for a single (or limited set) of students 
+        or assignments with the --assignment and/or --student flag:
+
+            nbgrader export --asignment assignmentID --student [studentID1,studentID2]
```
jhamrick:
```suggestion
            nbgrader export --assignment assignmentID --student [studentID1,studentID2]
```

nbgrader/plugins/export.py
```diff
@@ -9,6 +9,12 @@ class ExportPlugin(BasePlugin):
 
     to = Unicode("", help="destination to export to").tag(config=True)
 
+    student = List([], 
+        help="list of students to export").tag(config=True)
+    
+    assignment = Unicode("",
```
jhamrick: Why have two different ways of passing in lists here---a `List` in the case of students and then a `Unicode` string in the case of assignments? I think it would be better to have whichever method of passing in a sequence here be consistent.

nbgrader/plugins/export.py
```diff
@@ -56,46 +79,50 @@ def export(self, gradebook):
...
-
-                for key in score:
-                    if score[key] is None:
-                        score[key] = ''
-                    if not isinstance(score[key], str):
-                        score[key] = str(score[key])
-
-                fh.write(fmt.format(**score))
+            # test if assignment is requested
+            if not allassignments or assignment.name in allassignments:
```
jhamrick: So that four levels of indentation aren't required, I'd suggest the following instead:

```
if allassignments and assignment.name not in allassignments:
    continue

for student in gradebook.students:
    if allstudents and student.id not in allstudents:
        continue

    ...
```

nbgrader/plugins/export.py
ddland: The mix in lists was because of the presence of number-only names (studentID's) and alpha-numeric (assignments). Now implemented a list only version which does require alpha-numeric names quoted (bash shell) and the quotes need to be escaped. 
Changed to List for both student and assignment in new commit.

nbgrader/plugins/export.py
```diff
@@ -35,7 +41,24 @@ def export(self, gradebook):
...
+
+        if len(self.assignment) == 0:
+            allassignments = False
+        else:
+            # make sure assignment(s) are a list of strings
+            allassignments = [str(item) for item in self.assignment]
+
         self.log.info("Exporting grades to %s", dest)
+        if allassignments:
+            self.log.info("Exporting only assignments: %s", self.assignment)
```
jhamrick:
```suggestion
            self.log.info("Exporting only assignments: %s", allassignments)
```

nbgrader/plugins/export.py
```diff
@@ -35,7 +41,24 @@ def export(self, gradebook):
...
+
+        if allstudents:
+            self.log.info("Exporting only students: %s", self.student)
```
jhamrick:
```suggestion
            self.log.info("Exporting only students: %s", allstudents)
```

nbgrader/plugins/export.py
```diff
@@ -35,7 +41,24 @@ def export(self, gradebook):
         else:
             dest = self.to
 
+        if len(self.student) == 0:
+            allstudents = False
```
jhamrick:
```suggestion
            allstudents = []
```

`[]` evaluates to False in Python, and I think it's probably better to keep the datatypes consistent between different branches of the conditional.

nbgrader/plugins/export.py
```diff
@@ -35,7 +41,24 @@ def export(self, gradebook):
...
+        else:
+            # make sure studentID(s) are a list of strings
+            allstudents = [str(item) for item in self.student]
+
+        if len(self.assignment) == 0:
+            allassignments = False
```
jhamrick:
```suggestion
            allassignments = []
```

ddland:
The tests are implemented (first time using the framework, but they do pass!). Not sure how to add the new parameters to the docs, or is that done automatically?

jhamrick:
Awesome, thanks a lot! The new parameters will be automatically added to the docs, so no need to worry about that.

PR #1075, opened 2019-03-13
Give ExecutePreprocessor the Traitlets config during validation

rkdarst:
Root problem: it seems that validation timeout could not be changed with `--ExecutePreprocessor.timeout`, and in fact nothing we did would adjust the validation timeout.

=====
Commit message:

- We found that preprossor config was not followed - in particular,
  the ExecutePreprocessor.timeout wasn't noticed when it was increased
  to 120.
- A instructor did some digging and suggested that this change would
  make the configuration work.
- I don't know what the root problem could be, and I don't know if
  this is even needed.  Is this patch useful, or is there a "right"
  way of making traitlets work?

=====

So I don't know if this patch is needed.  I don't know if it's the right way.  From what I can tell about Traitlets, this shouldn't be needed.  Also documentation says that `c.ExecutePreprocessor.timeout` should work even without this change.

I've wondered if I could be putting the option in the wrong config file, but I have tried several and it even didn't work from the command line.

This patch was invented by an instructor using nbgrader.

I hope that giving a possible solution can help someone explain what we are doing wrong, but otherwise, here is a patch.

### DIFF
--- a/nbgrader/validator.py
+++ b/nbgrader/validator.py
@@ -272,6 +272,10 @@ class Validator(LoggingConfigurable):
         resources = {}
         with utils.setenv(NBGRADER_VALIDATING='1'):
             for preprocessor in self.preprocessors:
+                # https://github.com/jupyter/nbgrader/pull/1075
+                # It seemes that without the self.config passed below,
+                # --ExecutePreprocessor.timeout doesn't work.  Better solution
+                # requested, unknown why this is needed.
                 pp = preprocessor(**self.config[preprocessor.__name__])
                 nb, resources = pp.preprocess(nb, resources)
         return nb
--- a/nbgrader/validator.py
+++ b/nbgrader/validator.py
@@ -272,7 +272,7 @@ class Validator(LoggingConfigurable):
         resources = {}
         with utils.setenv(NBGRADER_VALIDATING='1'):
             for preprocessor in self.preprocessors:
-                pp = preprocessor()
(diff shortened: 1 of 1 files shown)


jhamrick:
Thanks for this! This is indeed strange and I would think shouldn't be required either, but I am happy to merge this as a workaround for now. However, the PR unfortunately needs a rebase. (Though the tests are currently broken at head, so you may want to hold off on that until they're fixed---I'll comment here again when that's done)

rkdarst:
Rebased, because I happened to be doing nbgrader stuff recently.  I should add I don't remember how much I tested this (I did another horrible hack as a workaround before this solution).

jhamrick:
Ok, seems to be fine with the preprocessor tests at least. Before I merge though, would you mind actually just adding a comment linking to this issue explaining why the config is being passed in the way it is?

rkdarst:
Done.  I really wish I could re-test to re-confirm this works...

FWIW this is my previous hack, as a container startup hook: `sed -i -E 's#(timeout = Integer\()(30)(,)#\1240\3#' /opt/conda/lib/python3.6/site-packages/nbconvert/preprocessors/execute.py`.

jhamrick:
Thanks!

Just as soon as #1084 passes and is merged then you can rebase again and test if you'd like. I can also try to test it manually (though probably not tonight, I have to go to bed soon :)

Issue #1073, opened 2019-03-11
nbgrader does not work with SQLAlchemy 1.3


soldis:
### Operating system
4.15.0-29-generic #31-Ubuntu SMP Tue Jul 17 15:39:52 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

### `nbgrader --version`
Python version 3.6.7 (default, Oct 22 2018, 11:32:17)
[GCC 8.2.0]
nbgrader version 0.5.5

### `jupyterhub --version` (if used with JupyterHub)
0.9.4

### `jupyter notebook --version` 
5.7.6

### Expected behavior
autograding, manual grading and watching submissions should work

### Actual behavior
they don't work

### Steps to reproduce the behavior

In formgrader when I go to "Manage Assignments" -> Submissions list (and in some other places in GUI - but the error is the same) I get "Loading, please wait..." and nothing happens. In log file:

Traceback (most recent call last):
      File "/usr/local/lib/python3.6/dist-packages/tornado/web.py", line 1697, in _execute
        result = method(*self.path_args, **self.path_kwargs)
      File "/usr/local/lib/python3.6/dist-packages/tornado/web.py", line 3174, in wrapper
        return method(self, *args, **kwargs)
      File "/usr/local/lib/python3.6/dist-packages/nbgrader/server_extensions/formgrader/base.py", line 103, in wrapper
        return f(self, *args, **kwargs)
      File "/usr/local/lib/python3.6/dist-packages/nbgrader/server_extensions/formgrader/apihandlers.py", line 146, in get
        submissions = self.api.get_submissions(assignment_id)
      File "/usr/local/lib/python3.6/dist-packages/nbgrader/apps/api.py", line 560, in get_submissions
        db_submissions = gb.submission_dicts(assignment_id)
      File "/usr/local/lib/python3.6/dist-packages/nbgrader/api.py", line 2372, in submission_dicts
        ).join(SubmittedNotebook, Notebook, Assignment, Student, Grade, GradeCell)\ 
      File "/usr/local/lib/python3.6/dist-packages/sqlalchemy/orm/query.py", line 2235, in join
        from_joinpoint=from_joinpoint,
      File "<string>", line 2, in _join
      File "/usr/local/lib/python3.6/dist-packages/sqlalchemy/orm/base.py", line 220, in generate
        fn(self, *args[1:], **kw)
[...]

PR #1080, opened 2019-05-10
compatibility with SQLAlchemy 1.3+

gertingold:
The present version of `nbgrader` is incompatible with `SQLAlchemy 1.3+` as pointed out in #1073. The reason is that version 1.3+ requires to explicitly specify the left side of a join (https://docs.sqlalchemy.org/en/13/changelog/migration_13.html#query-join-handles-ambiguity-in-deciding-the-left-side-more-explicitly). This PR proposes a solution by introducing the appropriate `select_from` method before a `join`.

### DIFF
--- a/nbgrader/api.py
+++ b/nbgrader/api.py
@@ -2920,6 +2920,7 @@ class Gradebook(object):
             SubmittedAssignment.id.label("id"),
             func.sum(Grade.score).label("code_score"),
             func.sum(GradeCell.max_score).label("max_code_score"),
+        ).select_from(SubmittedAssignment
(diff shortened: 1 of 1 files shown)


jhamrick:
Awesome, thanks a lot for investigating and fixing this. I think the test failures are unrelated---but let me just double check before merging.

gertingold:
Thank you for double checking. So far, I only had time to take a brief look at the errors and they seemed to be unrelated to this PR, at least at least at first sight.

jhamrick:
Yes, after checking I think they are unrelated (a few things have changed upstream that are breaking the tests). I'm going to merge this and then submit a new PR to fix the tests. Thanks again!!

PR #1093, opened 2019-05-29
Add authentication plugin support

jhamrick:
This PR moves the JupyterHub auth support for determining whether a student has access to a particular class into a special plugin class.

1. Creates a new `Authenticator` object which has methods for determining which classes a student has access to, as well as for adding and removing students from access groups.
2. Adds a plugin object to the `Authenticator` object, which is by default to have no authentication but there is also the option to use JupyterHub groups authentication, which uses the logic from #1040
3. The `Authenticator` object is passed around much like the `CourseDirectory` is so that the same object can be shared by multiple high-level nbgrader components.

cc @sigurdurb @damianavila @psychemedia @BertR

I will add docs and tests tomorrow. I also need to check that I haven't broken the JupyterHub auth.


jhamrick:
Ok, I have added some test cases but still not tested with an actual JupyterHub instance but will try to do that next. FYI, to use the JupyterHub auth, you would set the following in your config file:

```
from nbgrader.auth import JupyterHubAuthPlugin
c = get_config()
c.Authenticator.plugin_class = JupyterHubAuthPlugin
```

jhamrick:
Confirmed that this works with jupyterhub assuming you provide the correct config. Here's what I used:

```
c = get_config()

# Our user list
c.Authenticator.whitelist = [
    'instructor1',
    'instructor2',
    'student1',
]

c.Authenticator.admin_users = {
    'instructor1',
    'instructor2'
}

# instructor1 and instructor2 have access to different shared servers:
c.JupyterHub.load_groups = {
    'formgrade-course101': [
        'instructor1',
        'grader-course101',
    ],
    'formgrade-course123': [
        'instructor2',
        'grader-course123'
    ],
    'nbgrader-course101': [],
    'nbgrader-course123': []
}

# Start the notebook server as a service. The port can be whatever you want
# and the group has to match the name of the group defined above.
c.JupyterHub.services = [
    {
        'name': 'course101',
        'url': 'http://127.0.0.1:9999',
        'command': [
            'jupyterhub-singleuser',
            '--group=formgrade-course101',
            '--debug',
        ],
        'user': 'grader-course101',
        'cwd': '/home/grader-course101',
        'api_token': ''  # include api token from admin user
    },
    {
        'name': 'course123',
        'url': 'http://127.0.0.1:9998',
        'command': [
            'jupyterhub-singleuser',
            '--group=formgrade-course123',
            '--debug',
        ],
        'user': 'grader-course123',
        'cwd': '/home/grader-course123',
        'api_token': ''  # include api token from admin user
    },
]
```

PR #1098, opened 2019-05-30
Customizable student ID in `nbgrader submit`

nthiery:
This implements `nbgrader submit ps1 --student a4.jane.doo`

Some use cases:
- handling groups of students: helping the instructors access just the students of their group by adding a common prefix
- submissions directly from a jupyterhub, where all students share the same jovian login; in general infrastructures where all students share the same login
- a posteriori submissions by the instructor (e.g. because they are collected by some other mean)


nthiery:
Thanks!

PR #1099, opened 2019-05-30
Add course table, add course_id column to assignment

### LINKED ISSUE #960: Extend nbgrader.db to cope with multiple courses (and think beyond that)
The current `nbgrader.db` database has no concept of a course: each course has to have it's own database (configured in the config file)

In our environment, we have

- multiple instructors on a course, and 
- one instructor doing multiple courses

We also want to consider how to move to an external database (postgresql/mysql/etc).... but we are also very conscious that the notebook-servers themselves should *not* have access to create databases in our server (that could be horribly abused :chuckle: )

Our initial thought is to add a `courses` table, and link assignments to that

Keen to hear who else has been down this route, and what solutions have been tried.

BertR:
This will add a course table and a foreign key from the assignment table into the course table.

Closes #960

### DIFF
--- a/nbgrader/alembic/versions/e43177bfe90b_added_course_table_and_relationships.py
+++ b/nbgrader/alembic/versions/e43177bfe90b_added_course_table_and_relationships.py
@@ -16,11 +16,25 @@ branch_labels = None
 depends_on = None
 
 
+def _get_or_create_table(*args):
+    try:
+        table = op.create_table(*args)
+    except sa.exc.OperationalError:
+        table = sa.sql.table(*args)
+    return table
+
+
 def upgrade():
     """
     This migrations adds a course column to the assignment table
     and a matching foreign key into the course table
     """
+
+    new_course_table = _get_or_create_table(
+        'course',
+        sa.Column("id", sa.String(128), unique=True,
+                  primary_key=True, nullable=False),
+    )
 
     # If the course table is empty, create a default course
     connection = op.get_bind()
@@ -45,3 +59,5 @@ def downgrade():
 
         batch_op.drop_constraint('fk_course_assignment', type_='foreignkey')
         batch_op.drop_column('assignment', 'course_id')
+
+    op.drop_table('course')
--- a/nbgrader/alembic/versions/e43177bfe90b_added_course_table_and_relationships.py
+++ b/nbgrader/alembic/versions/e43177bfe90b_added_course_table_and_relationships.py
@@ -30,7 +30,7 @@ def upgrade():
 
     if len(results) == 0:
         connection.execute(
-            f"INSERT INTO course (id) VALUES ('{default_course}')")
+            "INSERT INTO course (id) VALUES ('{}')".format(default_course))
 
     with op.batch_alter_table("assignment") as batch_op:
 
--- a/nbgrader/alembic/versions/e43177bfe90b_added_course_table_and_relationships.py
+++ b/nbgrader/alembic/versions/e43177bfe90b_added_course_table_and_relationships.py
@@ -1,0 +1,47 @@
+"""Added course table and relationships
+
+Revision ID: e43177bfe90b
+Revises: 167914646830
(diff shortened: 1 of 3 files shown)


jhamrick:
Hey @BertR , just wanted to see how this was coming along? This is the last PR needed for the multiple classes support 🙂 

Just as a heads up I'm going to be away beginning on Sunday for two weeks and will have minimal time to look at stuff then, in case you want to get any feedback from me on it before then!

BertR:
HI @jhamrick, this now passes the test and should work as a first stepping stone towards
having configurable courses.

nbgrader/api.py
```diff
@@ -1375,34 +1384,36 @@ def set_course_id(self, course_id, **kwargs):
...
-    def course_id(self):
-        try:
-            course_id_record = self.db.query(Course)\
-                .one()
-            return course_id_record.id
-        except NoResultFound:
-            raise MissingEntry("No course id")
+        return course
+
+    # @property
```
jhamrick: Could you remove this commented code?

nbgrader/alembic/versions/e43177bfe90b_added_course_table_and_relationships.py
```diff
@@ -0,0 +1,47 @@
...
+
+    if len(results) == 0:
+        connection.execute(
+            "INSERT INTO course (id) VALUES ('{}')".format(default_course))
+
+    with op.batch_alter_table("assignment") as batch_op:
+
+        batch_op.add_column(sa.Column(
+            'course_id', sa.VARCHAR(128), nullable=False, default=default_course))
+        batch_op.create_foreign_key(
```
jhamrick: Do you know if `fk_course_assignment` is the same name as the foreign key created when the database is created from scratch? If not I'm just wondering if that could potentially cause problems later down the line...

jhamrick:
Thinking about it a bit more actually, I guess it won't work properly with multiple courses yet since it would actually require changing all the database queries to specify the course id. I think that's ok for now: we can make the other changes in another PR. For now, if you just address the more minor comments that I left, I'm happy to merge this 🙂

nbgrader/api.py
jhamrick: Looks like the commented code is still there, unfortunately...

nbgrader/tests/api/test_gradebook.py
```diff
@@ -433,6 +433,33 @@ def test_remove_notebook(assignment):
             assignment.find_notebook(nb.name, 'foo')
 
 
+def test_course_id_constructor():
```
jhamrick: Thanks for adding these tests!

nbgrader/alembic/versions/e43177bfe90b_added_course_table_and_relationships.py
BertR: In Sqlite, the foreign key constraint is not named. The migration seems to work in both directions, up and down, but I don't know whether I've considered all cases.

jhamrick: Ok, sounds good! Thanks for clarifying. I think it's fine if the constraint normally doesn't have an name.

nbgrader/alembic/versions/e43177bfe90b_added_course_table_and_relationships.py
```diff
@@ -0,0 +1,47 @@
...
+
+def upgrade():
+    """
+    This migrations adds a course column to the assignment table
+    and a matching foreign key into the course table
+    """
+
+    # If the course table is empty, create a default course
+    connection = op.get_bind()
+    res = connection.execute("select id from course")
```
jhamrick: This line will fail if the `course` table does not yet exist.

jhamrick:
Ok, #1147 is now merged so if you could rebase (or merge) that would be great!

PR #1102, opened 2019-05-30
Only build docs with one version of python

jhamrick:
cc @BertR


BertR:
Looks good, should shave 10 to 20 minutes :leopard: !

PR #1106, opened 2019-05-30
Remove testing of python 3.4

### LINKED ISSUE #1085: Deprecate support for python 3.4
Python 3.4 has also reached end-of-life, so this is a reminder to remove it from the CI tests and in setup.py.

jhamrick:
Fixes #1085

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 language: python
 python:
   - 2.7
-  - 3.4
   - 3.5
   - 3.6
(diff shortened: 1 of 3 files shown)


jhamrick:
Update: originally I had wanted to add tests for python 3.7 as well, but it seems that there are some dependencies which fail (not sure if this is because of 3.7 itself or because of the travis dependency on xenial). For now I will exclude them.

PR #1109, opened 2019-05-31
naming the temporary directories in tests

danielmaitre:
This made a big difference when debugging tests. Perhaps it would help others?

### DIFF
--- a/nbgrader/tests/apps/conftest.py
+++ b/nbgrader/tests/apps/conftest.py
@@ -12,7 +12,7 @@ from ...utils import rmtree
 
 @pytest.fixture
 def db(request):
-    path = tempfile.mkdtemp()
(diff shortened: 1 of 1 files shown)


jhamrick:
Oh, great idea! Thanks a lot! 🎉

nbgrader/tests/apps/conftest.py
```diff
@@ -36,7 +36,7 @@ def fin():
 @pytest.fixture
 def temp_cwd(request, course_dir):
     orig_dir = os.getcwd()
-    path = tempfile.mkdtemp()
+    path = tempfile.mkdtemp(prefix='tmp-cdw-')
```
slel: Did you want `cwd` instead of `cdw` here?

slel: See #1115 for a fix.

PR #1110, opened 2019-05-31
Systematic use of utils.get_username instead of $USER

nthiery:
By the way: adds a consistency test for get_username.


jhamrick:
Ah, I think you will need to mark `test_get_username` as a non-windows test. There is a decorator used elsewhere in the tests to mark non-windows tests that you can look for (I think it might be `@notwindows`, but you should double check)

PR #1111, opened 2019-05-31
Optional consistency check between owner and student_id upon collect

nthiery:
Mitigates #1097, #1098.

### DIFF
--- a/nbgrader/exchange/collect.py
+++ b/nbgrader/exchange/collect.py
@@ -1,7 +1,9 @@
 import os
 import glob
+import pwd
 import shutil
 from collections import defaultdict
+from textwrap import dedent
 
 from traitlets import Bool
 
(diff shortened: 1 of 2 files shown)


nthiery:
Hmm, this may require a  nowindows decorator. Let's see what the tests say.

nbgrader/exchange/collect.py
```diff
@@ -1,7 +1,9 @@
 import os
 import glob
+import pwd
```
jhamrick: Ah yeah, I think you will need to wrap this in a try/except so that the file can at least be imported on windows.

jhamrick: I think you still need to remove this line as it's still trying to import `pwd` on windows :(

nthiery: Hem, hem, hem :-) Sounds like I was not quite awake. Thank you for the cleanup!

nthiery:
Thank you Jessica for the review!

nbgrader/exchange/collect.py
jhamrick: It happens 😄No problem!

Issue #1030, opened 2018-10-10
Multiple instructors, multiple courses, and multiple solutions


perllaghu:
# This is a discussion-point, to try and bring together the various threads

We have a number of similar threads on multiple courses, multiple instructors, and the possibility of an external exchange service.

We all want to use NBGrader, as it's an excellent tool for creating tests & grading the responses.

It would be good to get a common solution, or at least several non-exclusive solutions, for these issues.

Please could people describe what they are trying to achieve, and the environment in wich they are running nbgrader?

jhamrick:
Great idea, thanks for creating this thread. Maybe let's explicitly ping some of the other people who have been thinking about this lately or might be interested in this discussion: @FranLucchini @rkdarst @sgvasquez @fishfree @zonca @sigurdurb @yuvipanda @dsblank 

* Related issues: #998 #961 #960 #544 #659
* Related PRs: #893 #1000

(Please feel free to reference any additional issues/PRs if I missed anything)

perllaghu:
# Environment

* Running in a Kubenetes cluster, with everything as Docker Images
* Pros: 
   * One central service for the whole organisation
   * Platform can be scaled by adding/removing worker nodes as needed (scale to suit load)
   * External service can deploy from a selection of Notebooks (different courses need different libraries)
   * Can use external data for users persistent storage (Network Storage services can be cheap)
   * Integrates with LTI to provide `course_id`, `user_id`, and `role` values (Devolves authentication & authorisation)
* Cons:
  * One central service for the whole organisation
  * All notebooks run as the same user (all external data is owned by the same user)
  * Persistent storage needs to be volume-mounted into the docker image (so it not dynamically changeable)
  * Selection of mount points becomes important

# Aims

We want users identified as `instructors` to have access to `formgrader` whilst everyone else gets `assignments` [done]. We need to cope with multiple classes, and cannot have students able to "borrow" another students submission. [doing]

# Current Plan

Our current plan is to dynamically create `nbgrader_config .py` at notebook startup, and set `CourseDirectory.root` to `/courses/<course_id>` - this will separate each course into it's own _namespace_

For the exchange stuff, the notebook Docker images will volume-mount:

For `outbound`:

    /srv/nbgrader/exchange/outbound -> [external_stroage_point]/<course_id>/outbound

For `inbound`, it depends on whether you're an `instructor` or a `student` (this is what stops students seeing each others work):

[...]

PR #1113, opened 2019-05-31
Add a course list extension that shows all courses an instructor can manage

### LINKED ISSUE #1060: Link to course services for teachers/TA's in their Jupyterhub user-home.
I notice some teachers/TA's want a link in their users home to their service where the nbgrader course is managed. Instead they have to remember and write their path /services/{course-id} 

I want them to be able to see their course either as a menu-list in the toolbar or under the Assignments tab.

In #1040 multiple course PR we are working from the idea that every Jupyter group that is named nbgrader-{courseID} is for the students and groups named formgrade-{courseID} are for teachers and TA's. So the idea would be to query the Jupyterhub API and only list the link to the services if the user is in some group called formgrade-{courseID}. 
Here is the api resource for how the services can be listed and the "prefix" could be fetched, it include the url path to the service.
https://jupyterhub.readthedocs.io/en/0.8.1/_static/rest-api/index.html#path--services

Any help would be appreciated.

jhamrick:
Fixes #1060 and related to #1030

This creates a new extension called `course_list` that displays all courses available to an instructor, including local ones (that are loaded into the same single user notebook) and JupyterHub ones (which are found by querying the JupyterHub services). Currently it looks something like this:

image

- [x] add docs
- [x] add tests
- [x] tes with a full JupyterHub setup

@damianavila @BertR @sigurdurb

### DIFF
--- a/nbgrader/__init__.py
+++ b/nbgrader/__init__.py
@@ -38,6 +38,14 @@ def _jupyter_nbextension_paths():
                 require="assignment_list/main"
             )
         )
+        paths.append(
+            dict(
+                section="tree",
+                src=os.path.join('nbextensions', 'course_list'),
+                dest="course_list",
+                require="course_list/main"
+            )
+        )
 
     return paths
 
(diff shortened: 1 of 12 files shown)


jhamrick:
There are minimal docs at least for how to install/disable the extension, so I will go ahead and merge and once #1114 is merge I will add a little more in a separate PR.

PR #1116, opened 2019-05-31
Autograde & Assign: create missing students/assignments by default

takluyver:
In a demo yesterday, @jhamrick suggested that the `nbgrader autograde` should behave as though `--create` was specified by default. This implements that change.

The `--create` flag can still be passed, to allow for scripts written with it, but it's a no-op now. You can still access the old behaviour (error if student doesn't exist in the database) by passing `--Autograde.create_student=False`.

**Edit:** The same now also applies to `nbgrader assign` - `--create` is now the default.

### DIFF
--- a/nbgrader/apps/assignapp.py
+++ b/nbgrader/apps/assignapp.py
@@ -35,7 +35,8 @@ flags.update({
     ),
     'create': (
         {'Assign': {'create_assignment': True}},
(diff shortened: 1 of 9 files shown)


takluyver:
As requested offline, this now makes the same change for `nbgrader assign`, to automatically create assignments in the database.

PR #1118, opened 2019-05-31
Integrate course_id into the api and apps

damianavila:
Python tests are passing for now... I need to check if the nbextensions tests are passing yet.
I could be missing some obvious things because I don't know the codebase deeply.
But happy to iterate on this one on the next few days.


jhamrick:
This is great! Thanks a lot Damian. It looks like the nbextensions tests are indeed failing---it's not immediately obvious to me why but my guess is there are some places in the nbextensions that need to change too. A trick for debugging them is that if you run the nbextensions tests locally, they'll save a screenshot of the browser when they fail so you can at least get a visual on what's going wrong.

nbgrader/apps/baseapp.py
```diff
@@ -190,6 +190,11 @@ def _load_config(self, cfg, **kwargs):
             cfg.Exchange.course_id = cfg.NbGrader.course_id
```
jhamrick: I took a look myself and I think the only thing that might be missing is just to update this check fo the course_id being in NbGrader---I think you can change this to `cfg.CourseDirectory.course_id = cfg.NbGrader.course_id` (and then also update the warning just above as well)

damianavila: Good catch... I forgot about that one! Pushing the change now!

jhamrick:
Thanks!

Issue #1117, opened 2019-05-31
Sanitize notebook documentation output


jhamrick:
Currently when building the docs, they produce a lot of outputs that involves user-specific information like username, path to the nbgrader repo, dates and times, etc. Ideally we could write some sort of postprocessing sanitizer to remove this information so it doesn't result in such large diffs when committing changes to the docs.

cc @danielmaitre @dsblank

danielmaitre:
I made it my train journey task. Should have a PR shortly...

jhamrick:
Closed by #1119

PR #1119, opened 2019-06-01
added a sanatizing step to the doc creation. 

danielmaitre:
It removes user specific information and replaces some timestamps and file sizes with placeholders. It address #1117.


jhamrick:
This is awesome, thanks a lot for doing this!

PR #1135, opened 2019-06-02
section on grading in docker container

### LINKED ISSUE #483: Improve security by running autograder in isolation

gertingold:
This PR proposes to add to the advanced topics section of the docs a section on grading in a docker container with the help of ``envkernel``. The idea was to concentrate on docker as discussed in #483. Singularity is mentioned as a possibility at the end. I also did not follow the ``envkernel`` README very closely but refer to the README in a couple of places instead. 

Any suggestions for improvements are very welcome.

Closes #483

### DIFF
--- a/nbgrader/docs/source/user_guide/advanced.rst
+++ b/nbgrader/docs/source/user_guide/advanced.rst
@@ -187,8 +187,8 @@ nbgrader from the command line
 
     nbgrader autograde --ExecutePreprocessor.kernel_name=NAME ASSIGNMENT_NAME
 
-In addition to docker, ``envkernel`` also supports `singularity
-<https://www.sylabs.io/docs/>`_ as a containerization system. For details 
-on using ``envkernel`` with singularity, see the `README
+In addition to docker, ``envkernel`` also supports singularity as a
+containerization system. For details on using ``envkernel`` with
+singularity, see the `README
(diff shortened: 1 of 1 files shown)


gertingold:
I had to remove a link, because it caused problems with ``linkcheck`` even though the link itself is correct...

gertingold:
The test failing now does not seem to be related to my PR.

rkdarst:
Looks good to me.  I guess I should make a proper release of envkernel would be good, but shouldn't stop anything here.

rkdarst:
@gertingold, I just did a major update to envkernel, which shouldn't affect docker but should make things more extendable.  And most importantly, adding extensive tests.

Would you like to make sure it still works for your case?  I have no reason to think it wouldn't, just an abundance of caution since first experiences will be important (and reference here since it's still new).  I'll try to make a proper release once possible.

gertingold:
I will take a look at it as soon as possible and will let you know about the result. Thanks for your work!

gertingold:
@rkdarst, I just tested your new version of envkernel together with docker. As far as I can tell, everything works fine.

rkdarst:
Thanks!  So I guess things should be mostly stable now, but definitely let me know if things break in the future, or more features are needed.

PR #1164, opened 2019-07-19
assignment_dir: Add into several missing places

rkdarst:
This adds assignment_dir config to several places it was missing (see #1001).

Incidentally, while testing this I noticed something.  Without this, you have to be in the notebook root directory to fetch/submit/etc.  With this, you can be in any directory.  That seems to imply it would be beneficial to set the default to abspath(Jupyter notebook_dir) (but we can't use that directly, because it still needs to be configurable separately, that's the point of #1001).


rkdarst:
This is a failure of `TestNbGraderAutograde.test_infinite_loop` on only Python 3.5 and Python nightly.  I don't think this should affect it - do you think it's a flaky test?

jhamrick:
Thanks!

Issue #1039, opened 2018-10-30
Language support


sigurdurb:
I would like to add language support for Octave kernel.
And make it easier to add more languages and not have the code that checks for errors in two places like it is now in validator._extract_error and utils.determine_grade

You can see the code in this PR I just made that describes the changes better: #1038

jhamrick:
Is there more that needs to be done on this? I'm not sure I know what you have in mind by adding additional language support for Octave.

sigurdurb:
Hey @BertR, hope you are doing well. I am just checkin, are you guys using Octave assignments at Noteable? I remember someone mentioning using Octave at their school during the Hackathon but can't remember who it was. In any case, did you guys change the grader when you are validating notebooks with Octave for example ? 
You can see the problems I had with trying to validate with Octave assignments [here](https://github.com/jupyter/nbgrader/pull/870#issuecomment-430788312)

perllaghu:
Noteable definitely uses nbgrader with `Python`, `R`, `Sage`, and `Stata`.

We have, in the past, used it with `Haskell` too.

The solution is using the `ClearSolutions.xxx` settings in your `nbgrader_config.py` file 

Eg:
```
#-----------------------------------------------------
# config chages to allow nbgrader to work with stata
# NOTE: this *stops* python stuff working!
#-----------------------------------------------------
c.ClearSolutions.code_stub = dict(
    stata = '* YOUR CODE HERE\nerror 192'
    )
c.ClearSolutions.begin_solution_delimeter = "* BEGIN SOLUTION"
c.ClearSolutions.end_solution_delimeter = "* END SOLUTION"
```

PR #1173, opened 2019-08-15
Langsupport: Validation refactoring and better language support for autograder test cells.

sigurdurb:
* Refactors repeated code. The code could maybe be refactored even more.
* Better validation of errors, now supports Gnu Octave.
To get the kernel name I use lang = nb.metadata.get('kernelspec',{}).get("language","python"). Even though it says python it gets the real name of the kernel. So 'octave' is returned for the octave kernel for example. So contributers can now use the lang variable to do custom error handling for their kernel if it does not support the default functionality.
Octave kernel error matching now works,  The code for this is in is_validate_error in utils.py, Octave does not have a cell output_type 'error' but we re.match the error message from the "text" output that has the "stream" output_type. That is the only way I found that works for Octave. Just don't put display('error: ') in a your autograder test cell as a teacher, those cells are read only for students anyway so for students that should not be a problem.

### DIFF
--- a/nbgrader/utils.py
+++ b/nbgrader/utils.py
@@ -67,7 +67,7 @@ def is_validate_error(cell, lang = None):
         for output in cell.outputs:
             if output.output_type == "stream": 
                 if hasattr(output, "text"):
-                    if re.match("error: ", output.text): 
+                    if match("error: ", output.text): 
                         errors.append(True)
                         return errors
                     else:
--- a/nbgrader/utils.py
+++ b/nbgrader/utils.py
@@ -97,12 +97,8 @@ def determine_grade(cell, lang = None):
             return None, max_points
 
     elif cell.cell_type == 'code':
-        if any(is_validate_error(cell, lang, 1)):
+        if any(is_validate_error(cell, lang)):
                 return 0, max_points
-        elif output.output_type == 'stream':
-            if hasattr(output, 'text'):
-                if re.match('error: ', output.text):
-                    return 0, max_points
         return max_points, max_points
 
     else:
--- a/nbgrader/utils.py
+++ b/nbgrader/utils.py
@@ -10,6 +10,7 @@ import logging
 import traceback
 import contextlib
 import fnmatch
+from re import match
 
 from setuptools.archive_util import unpack_archive
 from setuptools.archive_util import unpack_tarfile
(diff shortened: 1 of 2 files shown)


sigurdurb:
Continuation of PR:#1038 and Issue: #1039

jhamrick:
Thanks for the work you put into this!

However, I'm not sure this is something that I want to merge into master as is, as I would really like to keep all of the code in nbgrader as language agnostic as possible, and I would like to avoid any special casing for particular languages.

It seems to me that this is more of an issue with the octave kernel: it should be returning errors using the "error" output type rather than a stream output. I guess though for some kernels it might be challenging to do this (e.g. for bash)---really all you have is the stream output. In that case, I would be ok with merging a PR which checks to see if there is a stream with `"name": "stderr"` and if so treating that as an error, i.e. something like:

```
def get_errors(cell):
    errors = []
    for output in cell.outputs:
        if output.output_type == "stream" and output.name == "stderr":
            errors.append(output.text)
        elif output.output_type == 'error':
            errors.append(output.traceback)
    return errors
```

However, I do not want to merge something that checks for a language type explicitly.

Also, it unfortunately looks like these changes have undone some other stuff surrounding the logic of `self.validate_all`. The logic of `_get_failed_cells` and `_get_passed_cells` is legitimately different so I don't think it makes sense to factor that out into a shared method. Would you mind undoing those changes, and just making changes individually in those two methods for appropriately handling errors?

sigurdurb:
There were two use cases with this PR:
* If there is a use case use case to having a variable for language, "lang" being passed to determine_grade. But if nbgrader is language agnostic it would probably not be used in that function unless it is just an option for people to make their own custom changes more easily for kernels like octave.

* A general refactoring PR could be useful. The [utils.is_validate_error](https://github.com/jupyter/nbgrader/blob/ef06d0e583cc2895aa2283fe3a2d2749df375162/nbgrader/utils.py#L64) function I added refactored the error checks for:

```
for output in cell.outputs:
    if output.output_type == "error":
         errors.append("\n".join(output.traceback))
```
That were in utils.determine_grade and validator._extract_error (That now also have a check for partial credit)

is_validate_error first checked if the language was 'octave'. Also in [_extract_error](https://github.com/jupyter/nbgrader/blob/ef06d0e583cc2895aa2283fe3a2d2749df375162/nbgrader/validator.py#L108) there is a check for 'octave' because there the error is in the cell text but else it is in the cell traceback. But it is not ideal to have octave hardcoded in two places.

But if anyone wants to make language support for weird kernels they could use this code for where to implement changes.

sigurdurb:
This code would need to be separated into several PR's. Going to close this PR and refer it to people who are interest in custom support for languages.

williamstein:
I have no choice but to deal with this same problem in CoCalc (because paying customers expect to be able to use nbgrader + octave).  In case you're curious, here's how it works there (like you, I check for `error: ` at the beginning of the line):

https://github.com/sagemathinc/cocalc/blob/master/src/smc-webapp/jupyter/nbgrader/autograde.ts#L219

Note that this is Typescript code, rather than Python.

Also, note that there are subtle similar problems that happen when using the R kernel.

williamstein:
I also agree that this is all working around bugs in these kernels.  However, I'm not optimistic that these bugs will be fixed in a timely manner.

PR #1177, opened 2019-08-16
make path relative

phaustin:
the validate button is failing for me on current master with error messages like:

```
FileNotFoundError: [Errno 2] No such file or directory: '/source/Assign_1'
```

this is because path is being set with '/' as the first character, making the os.path.join with the notebook_dir a noop.  My fix using pathlib.relative_to works, but obviously it would be better to catch this at the initialization and make sure path is relative.

### DIFF
--- a/nbgrader/server_extensions/validate_assignment/handlers.py
+++ b/nbgrader/server_extensions/validate_assignment/handlers.py
@@ -6,6 +6,7 @@ import traceback
 
 from tornado import web
 from textwrap import dedent
+from pathlib import Path
 
 from notebook.utils import url_path_join as ujoin
 from notebook.base.handlers import IPythonHandler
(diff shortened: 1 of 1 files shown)


jhamrick:
Hmm, I'm not able to reproduce this unfortunately. Do you know why the path is starting with a `/` in the first place? I would think that shouldn't happen... can you open a JavaScript console on the offending notebook and look at what `Jupyter.notebook.notebook_path` is?

But I agree if this is possible we should check that the path is relative rather than absolute. However, we haven't *quite* yet dropped support for Python 2 (happening very soon though). Would you mind writing this without pathlib so that the python 2 tests will still pass?

jabalazs:
I am having the same issue. The `Jupyter.notebook.notebook_path` in my case is `/source/assignment_3/assignment_3.ipynb`.

Steps to replicate:
1. Create a new conda environment: 
  `conda create -n nbgrader python=3.7`
2. Install `nbgrader`: 
  `conda install nbgrader -c conda-forge`
  At this point, the validate button still works.
3. Install `jupytext`: 
`conda install jupytext -c conda-forge`
  The validate button does not work anymore, and displays the error referenced above.

Issue #1157, opened 2019-06-12
Exclude students when autograding


rkdarst:
Autograde has a `--student` option, which limits to autograde only one student.  Sometimes, autograde fails for unknown reasons (currently bumping against metadata updates), but it's concievable that there could be more.  It would be useful for me to have the equivalent of `--skip-students` option which skips the given student IDs (for example, comma separated list).

Right now I have a hack where, if it fails, I run `ls submitted/ | grep -v (bad usernames) | cut -d/ -f1 | xargs -i nbgrader autograde [...] [assignment_name] --student={}`, which basically re-invokes for each student explicitly using `--student` - it's not perfect because it also invokes for students who haven't submitted.

This is related to "don't stop if there is an error with one notebook", and #1083 somewhat.  There have been some recent updates to metadata handling, which may solve an immediate problem, but this could still be useful...

I'm using what was approximately master before the hackathon, Linux.

jhamrick:
Yes, seems reasonable to add something like this. Ideally we'll find all the ways that nbgrader fails during grading and handle them a bit more gracefully, but I think this is a nice fallback option as I expect there will always be some edge cases.

rkdarst:
done in #1179 long ago.

PR #1179, opened 2019-08-19
Add CourseDir.student_id_exclude option to exclude students

rkdarst:
This adds a `--CourseDirectory.student_id_exclude` which can exclude the given students from autograding and other tasks, as in #1157.  This is useful if there is a bug which causes autograding to halt if there is a problem with one particular student.

There are no tests yet and I haven't used much.  I'm not sure if I found the right place to add this.  If it is, then let me know and I'll see about adding the tests and documentation.

-----
- This is the counterpart of the CourseDir.student_id option in that
  it rejects single students, defaulting to them all.  Really, first
  student_id is applied, and then student_id_exclude ones are skipped.
- This is useful if a single student is erroring and needs to be left
  out.

### DIFF
--- a/nbgrader/converters/base.py
+++ b/nbgrader/converters/base.py
@@ -172,6 +172,11 @@ class BaseConverter(LoggingConfigurable):
         initialization was successful).
 
         """
+        if self.coursedir.student_id_exclude:
+            exclude_ids = self.coursedir.student_id_exclude.split(',')
+            if student_id in exclude_ids:
+                return False
+
         dest = os.path.normpath(self._format_dest(assignment_id, student_id))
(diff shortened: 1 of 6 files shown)


jhamrick:
Seems reasonable to me. If you wanted to add a test or two that would be great!

rkdarst:
Tests added (and found that release_feedback wasn't supported yet, fixed now).  Tests are added for autograde, generate_feedback, release_feedback.  Anything that goes through the converter system (but perhaps not the exchange system) should work, though, but I think these are the three main ones.

jhamrick:
Thanks a lot!

PR #1183, opened 2019-08-23
Feedback: update fetch_feedback command line help

rkdarst:
- I don't really know if everthing in this description update is
  correct, but I did the best I could.  Please sanity-check first.
- As noticed by @sefriol

### DIFF
--- a/nbgrader/apps/fetchfeedbackapp.py
+++ b/nbgrader/apps/fetchfeedbackapp.py
@@ -34,7 +34,7 @@ class FetchFeedbackApp(NbGrader):
 
         To fetch an assignment by name into the current directory:
 
-            nbgrader fetch assignment1
+             nbgrader fetch_feedback assignment1
 
         To fetch the assignment for a specific course (or if your course_id is
         not set in a configuration file already), you must first know the
--- a/nbgrader/apps/fetchfeedbackapp.py
+++ b/nbgrader/apps/fetchfeedbackapp.py
@@ -26,25 +26,28 @@ class FetchFeedbackApp(NbGrader):
     flags = flags
 
     examples = """
-        Fetch feedback for an assignment that an instructor has released. MORE INFO NEEDED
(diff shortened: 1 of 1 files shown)


nbgrader/apps/fetchfeedbackapp.py
```diff
@@ -26,25 +26,22 @@ class FetchFeedbackApp(NbGrader):
...
+        Fetch feedback for an assignment, if an instructor has released it.
         For the usage of students.
 
-        You can run this command from any directory, but usually, you will have a
-        directory where you are keeping your course assignments.
+        You can run this command from any directory, but usually, you will run
+        it from the directory where you are keeping your course assignments.
 
-        To fetch an assignment by name into the current directory:
+        To fetch feedback, you must first know the `course_id` for your course.
```
jhamrick: If the course_id is set in the `nbgrader_config.py` file, then you don't need to explicitly pass it as a flag, so I would maybe leave the current documentation that shows it without the flag.

rkdarst:
Updated.  One thing I wasn't clear on is if it had to be run from the main notebook dir/assignment dir, or could be run anywhere.  The code wasn't completely clear to me last time I checked.

Sefriol:
Atleast in our environtment you can run it in any folder. Feedback destination is always the same.

nbgrader/apps/fetchfeedbackapp.py
```diff
@@ -26,25 +26,28 @@ class FetchFeedbackApp(NbGrader):
...
         For the usage of students.
 
-        You can run this command from any directory, but usually, you will have a
-        directory where you are keeping your course assignments.
+        You can run this command from any directory, but usually, you will run
+        it from the directory where you are keeping your course assignments.
 
         To fetch an assignment by name into the current directory:
 
             nbgrader fetch assignment1
```
jhamrick:
```suggestion
             nbgrader fetch_feedback assignment1
```

Issue #1181, opened 2019-08-21
Fetching feedback should not be part of the formgrader extension


jhamrick:
As pointed out by @QuantumEntangledAndy in #1178: 

> Also is it intentional that the assignment list make calls to the formgrader server extension? I was under the impression that students only required access to the assignment list extension but perhaps I was mistaken? I had to enable the formgrader server extension for all the students as well as apply the above patch.

The route for fetching feedback should indeed be part of the assignment list extension, not the formgrader extension.

PR #1184, opened 2019-08-23
Move the fetch feedback API from formgrader to assignment_list

rkdarst:
- I was basically operating by guessing, so someone should check if
  this is appropriate - there's quite a difference between the way
  that assignment_list and formgrader access the API.  I copied from
  the "fetch" is assignment_list and adjusted ExchangeFetchAssignment
  to ExchangeFetchFeedback
- Also there's a bit too much copy-and-paste, but fixing that is
  beyond me...
- But I did just run this and it *seems* to work... as in it can
  fetch.
- Closes: #1181

### DIFF
--- a/nbgrader/nbextensions/assignment_list/assignment_list.js
+++ b/nbgrader/nbextensions/assignment_list/assignment_list.js
@@ -595,11 +595,7 @@ define([
             button.attr('disabled', 'disabled');
             var url = utils.url_path_join(
                 that.base_url,
-                'formgrader',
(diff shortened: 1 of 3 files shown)


rkdarst:
Something about this - or one of my other PRs - causes a problem in the assignment list tab.  When you fetch feedback, assignments in the available/downloaded assignment areas are duplicated.  I haven't been able figure out why yet, and probably someone that knows more can do it quickly.

PR #1120, opened 2019-06-01
Integrate feedback distribution within nbgrader

danielmaitre:
This is the result of the work of @dsblank, @perllaghu, @billywardrop29 and @danielmaitre at the Edinburgh hackathon.  

It includes UI integration of the feedback step, release of the feedback to students in the exchange, and fetching of this feedback by the students.

Some UI work needs to be finished but starting a PR will help check the coverage of the existing tests.

Fixes #795
Replaces #1096

PR #1186, opened 2019-08-23
Remove student_id and change root to cache, permission check to only execute

Sefriol:
There might be some tests that need to be checked, but I put this as initial starting point.

Issue was talked in here #1120 

Discussion welcome and required.

### DIFF
--- a/nbgrader/tests/apps/test_api.py
+++ b/nbgrader/tests/apps/test_api.py
@@ -667,7 +667,7 @@ class TestNbGraderAPI(BaseTestApp):
     @notwindows
     def test_release_and_unrelease(self, api, course_dir, db, exchange):
         self._copy_file(join("files", "submitted-unchanged.ipynb"), join(course_dir, "source", "ps1", "p1.ipynb"))
-        result = api.assign("ps1")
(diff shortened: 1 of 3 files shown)


nbgrader/tests/apps/test_api.py
```diff
@@ -21,6 +21,7 @@ def api(request, course_dir, db, exchange):
     config = Config()
     config.CourseDirectory.course_id = "abc101"
     config.Exchange.root = exchange
+    config.Exchange.cache = exchange
```
jhamrick: Usually this should be a different directory... it's probably fine here but just for consistency you probably want to pass the "cache" fixture to the signature of this method and then set that as the path here (which should just automagically work, because `cache` is defined in `conftest.py` and pytest is magic):

```
def api(request, course_dir, db, exchange, cache):
    ...
    config.Exchange.cache = cache
    ...
```

Sefriol: Wasn't actually sure if that would just work and forgot to test it out in the end. Will change that.

jhamrick:
Thanks a lot!!

Issue #1189, opened 2019-08-24
Student courses not appearing


jhamrick:
When using the "multiple courses" setup with JupyterHub authentication, it does not seem that students can actually view assignments in the courses they are in.

PR #1191, opened 2019-08-24
Allow access to formgrader when not using JuptyerHub auth

jhamrick:
If just running a single class on JupyterHub and not using JupyterHub auth, it should still be possible to access the formgrader from the courses list. This makes this possible, and also clarifies some of the documentation around setting this up.

Closes: #1189


jhamrick:
I also added to this a fix for #1189, which wasn't really so much a bug as much as a misconfiguration. But hopefully it should be clearer now that the course id needs to be included in the config files for the grader accounts.

Issue #1193, opened 2019-08-25
Update releasing instructions to use twine


jhamrick:
Currently the releasing instructions say to use `python setup.py sdist upload` but the recommended way to upload to pypi these days is with twine, so this should be updated.

PR #1194, opened 2019-08-25
Update releasing instructions

jhamrick:
This also removes running tests when creating the release, but I think these are mostly redundant and take a long time, so I don't think it's worth it to include them.

Closes: #1193


jhamrick:

jhamrick:

Issue #1188, opened 2019-08-24
Assignments duplicated when fetching feedback


jhamrick:
As reported by @rkdarst, when fetching feedback, this causes the list of currently available assignments to be duplicated for some reason.

Before:
<img width="1164" alt="Screen Shot 2019-08-24 at 5 12 39 PM" src="https://user-images.githubusercontent.com/83444/63639982-9642dd00-c692-11e9-920d-00b384d78637.png">

After:
<img width="1157" alt="Screen Shot 2019-08-24 at 5 14 19 PM" src="https://user-images.githubusercontent.com/83444/63639988-a2c73580-c692-11e9-9503-40c0c284fe35.png">

PR #1197, opened 2019-08-25
Do not duplicate assignments when fetching feedback

jhamrick:
There was some logic which was causing the fetch assignments to not be refreshed when fetching feedback, which then had the result of not removing the previous items before adding the new items to it. This removes that logic, so that the full assignment list gets refreshed when feedback is fetched, thus ensuring the assignments aren't duplicated.

Ping @dsblank @danielmaitre -- do you remember what the idea behind the original logic was? Is this PR doing the right thing?

Closes: #1188

cc @rkdarst @QuantumEntangledAndy

### DIFF
--- a/nbgrader/nbextensions/assignment_list/assignment_list.js
+++ b/nbgrader/nbextensions/assignment_list/assignment_list.js
@@ -225,26 +225,15 @@ define([
         }
     };
 
-    AssignmentList.prototype.handle_load_feedback_list = function (data, status, xhr) {
-        if (data.success) {
-            // update_fetched_only:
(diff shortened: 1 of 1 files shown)


dsblank:
@jhamrick That code was the trickiest to work on. If I remember correctly, it really needed a refactor to update the whole page, but we worked around that by just updating the fetched assignments. I think that was because the update lists came from two different sources: either all, or fetched. It looks like this refactors all of that.

I think if this works, then it probably is correct. :)

jhamrick:
Great, thanks!

jhamrick:

Issue #1141, opened 2019-06-05
Possible to fetch feedback for only some notebooks in an assignment


jhamrick:
If feedback has been generated and released for an assignment, and then one notebook changes and the assignment is resubmitted, then when the student clicks "fetch feedback" (before the new feedback is available) the feedback will be fetched for the unchanged notebook but not the changed notebook. This gives the reverse impression than what I think there should be: that there is new feedback available for the unchanged notebook (false) and there is no feedback available for the changed notebook (true at the moment, but might be false in the future). I think in this case it should detect that new feedback for the new submission is not yet available.

Steps to reproduce:
1. Submit an assignment.
2. Generate and release feedback for the assignment.
3. Fetch feedback for the assignment.
4. Make a change to one notebook in an assignment (but not all notebooks).
5. Resubmit assignment.
6. Fetch feedback for the assignment. Note that this succeeds, and that there is only feedback available for the notebook(s) that did not change.

dsblank:
@jhamrick Are you doing the steps through the UI or the command line?

I see a related issue (or same issue):  I can't generate new feedback if I regrade one of two notebooks. I will explore further.

jhamrick:
This was through the web interface. Thanks for investigating!

dsblank:
These are notes to myself as I explore the issue (so I don't forget... may take a few days to resolve). Also, not sure these notes are the same, related, or different issue as above.

1. It appears that `converters/base.py:init_destination` isn't detecting newer graded assignments so that we can re-generate
2. It appears that even if forced to generate feedback, and then released, the student fetch operation is not detecting newer feedback (which should show "ready to fetch" message next to submission and the "Fetch Feedback" button.

PR #1200, opened 2019-08-26
Associate feedback files with unique submission attempts

jhamrick:
This updates the logic for computing the feedback hashes so that the hash depends both on the feedback contents as well as the course, assignment id, notebook id, student id, and timestamp. Feedback files using the previous hashing scheme should still be detected, however, so this should be backwards compatible with previously generated feedback files.

This alleviates the issue in #1141 where it was possible to fetch feedback for only some notebooks in an assignment, which had feedback available because the notebook contents hadn't changed. It also fixes the confusing behavior whereby if a student submits an unchanged assignment, they immediately have feedback available---technically it will be the same behavior, but I found this surprising and confusing since I hadn't gone through the actual process of generating feedback yet.

In the process of doing this, I restructured a bit the integration between `nbgrader list` and the Assignment List extension:
* `nbgrader list` now returns a list of submissions for each assignment, rather than a flat list of assignment submissions. 
* `nbgrader list` also now lists in the command line output what the status of feedback is (available, ready to be fetched, not available).
* Fixed a few subtle bugs relating to if you have set custom values for the `feedback` or `submitted` directories, or the way the nbgrader directory is structured.
* Also did some general code cleanup (there were some variables that were defined but not used, or two variables defined for the same thing).
* In the Assignment List extension, I created a new `Submission` class to handle the view of each individual submission.
* Separate submissions for an assignment are now listed indented under the assignment name:

Screen Shot 2019-08-26 at 5 33 18 PM

Closes: #1141

### DIFF
--- a/nbgrader/exchange/fetch_feedback.py
+++ b/nbgrader/exchange/fetch_feedback.py
@@ -3,7 +3,7 @@ import shutil
 import glob
 
 from .exchange import Exchange
-from ..utils import check_mode, notebook_hash
+from ..utils import check_mode, notebook_hash, make_unique_key, get_username
 
 
 class ExchangeFetchFeedback(Exchange):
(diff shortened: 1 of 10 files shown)


rkdarst:
One idea I had was to, at submission time, generate a random token and put it into the submitted notebook and some cache.  Then, there is less coupling between a hash algorithm and retrieving the notebook later (which is probably a good thing)...

On the other hand, tracking tokens separately from the notebook makes more work.  They could be embedded in the notebook metadata, but you have the same problem as the next paragraph.

All this actually this makes me think: if a student A shares their notebook with student B, is B then able to retrieve A's results and grade?  It would be best to avoid this if possible...

jhamrick:
> One idea I had was to, at submission time, generate a random token and put it into the submitted notebook and some cache. Then, there is less coupling between a hash algorithm and retrieving the notebook later (which is probably a good thing)...

Yeah, that's not a bad idea. Probably more efficient too as then you don't have to compute the hash of the entire notebook. Really I think you actually shouldn't need the hash at all; I think it should probably be enough to have the unique key of `(course, student, assignment, notebook, timestamp)`...

> if a student A shares their notebook with student B, is B then able to retrieve A's results and grade? It would be best to avoid this if possible...

Yes, I think the current setup would allow this. But I'm not really sure of a way around this barring having better access controls (which we should have once hubshare exists...).

jhamrick:

Issue #1201, opened 2019-08-26
Feedback for regraded assignments not detected


jhamrick:
Originally reported by @dsblank  in #1141:

> I see a related issue (or same issue): I can't generate new feedback if I regrade one of two notebooks.
> 1. It appears that converters/base.py:init_destination isn't detecting newer graded assignments so that we can re-generate
> 2. It appears that even if forced to generate feedback, and then released, the student fetch operation is not detecting newer feedback (which should show "ready to fetch" message next to submission and the "Fetch Feedback" button.)

Steps to reproduce:

1. As instructor: generate and release assignment.
2. As student: fetch and submit assignment.
3. As instructor: collect, grade, generate feedback, release feedback.
4. As student: fetch feedback.
5. As instructor: regrade, generate feedback (here is problem 1: through the formgrader interface, this fails, though it works if you generate feedback on the command line with `--force`).
6. As instructor: release feeback
7. As student: assignment list shows that feedback is still viewable, but this should be changed to "ready to fetch" as there is new feedback available.

jhamrick:
(Note: if in the last step the student fetches feedback, then it does get the new feedback. However, there should be something in the UI telling the student that they can do this.)

PR #1204, opened 2019-08-26
Force generate feedback by default in API

jhamrick:
Fixes the first problem in #1201


jhamrick:

PR #1205, opened 2019-08-26
Check for newer feedback in nbgrader list

jhamrick:
If feedback exists both locally and in the exchange, compute the checksums of both feedback files and compare them. If they're different, then report that newer feedback can be fetched.

Closes: #1201


jhamrick:

PR #1222, opened 2019-09-18
CourseDir.format_path: supports absolute paths in nbgrader_step

nthiery:
This enables configuring absolute paths -- and in particular paths
outside of the main course directory for e.g.
c.CourseDirectory.submitted_directory.

Use case: for our course here, we have two separate shared folders (actually git repositories) among instructors. One for the course material, including the source notebooks, which is public. Another for everything involving student submissions, which is private and deleted every year. This PR gives more flexibility in the layout of the two shared folders.

### DIFF
--- a/nbgrader/tests/apps/test_config.py
+++ b/nbgrader/tests/apps/test_config.py
@@ -1,0 +1,15 @@
+from traitlets.config import Config
+from ...coursedir import CourseDirectory
+from .base import BaseTestApp
+from .conftest import notwindows
+
+class TestCourseDirectory(BaseTestApp):
+
+    @notwindows
+    def test_submit_rectoryget_source_assignments(self):
+        config = Config()
+        config.Exchange.course_id = "abc101"
+        config.CourseDirectory.root = "/root"
+        coursedir = CourseDirectory(config=config)
+        assert coursedir.format_path("submitted", "alice", "HW1") == "/root/submitted/alice/HW1"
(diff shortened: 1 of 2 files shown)


nthiery:
Note: I haven't found an existing location for unit tests for CourseDirectory. So I created one, randomly guessing where it could be reasonable to locate it.

nthiery:
Weird, Travis reports that some checks haven't completed above, yet the details state that they have all passed.

nthiery:
Now that's really weird: two minutes ago, the continuous integration results above complained for Python 2.7 and 3.6 about exactly the error message of #1239 ! Now, it's just pending.

nbgrader/tests/apps/test_config.py
```diff
@@ -0,0 +1,15 @@
+from traitlets.config import Config
+from ...coursedir import CourseDirectory
+from .base import BaseTestApp
+from .conftest import notwindows
+
+class TestCourseDirectory(BaseTestApp):
+
+    @notwindows
+    def test_submit_rectoryget_source_assignments(self):
```
jhamrick: Is the title of this method a typo?

nthiery: Oops, yes definitely. No idea how it became this :-) Fixed and pushed!

nthiery:
I merged in the latest master and pushed. Let's see how the tests go!
Thank you @jhamrick for the review.

jhamrick:
Thanks!

PR #1228, opened 2019-09-26
Fix formgrader API

BertR:
Formgrader was using the default course id for every call to the Gradebook.
By passing in the course id, every assignment in the gradebook will be created
with the course_id that was set in the config.


jhamrick:
Thanks!

jhamrick:

PR #1239, opened 2019-10-07
Fix: Application._load_config_files returns pairs (config, filename)

nthiery:
Steps to reproduce the original bug:
- Put a nbgrader_config.py file somewhere in the configuration path (e.g. /etc/jupyter)
- Run jupyter
- Switch to the Courses or Assignments tabs
- A error is issued with the following chunk of traceback:
        File "/opt/miniconda3/envs/info-114/lib/python3.7/site-packages/nbgrader/server_extensions/assignment_list/handlers.py", line 44, in load_config
          full_config.merge(config)
        File "/opt/miniconda3/envs/info-114/lib/python3.7/site-packages/traitlets/config/loader.py", line 185, in merge
          for k, v in other.items():
        AttributeError: 'tuple' object has no attribute 'items'

Analysis: `Application._load_config_files` from `IPython`'s `traitlets` package yields not config objects but pairs (config object, filename).

I am quite surprised this did not strike anyone earlier, and I did not find any recent change in IPython's traitlets or nbgrader that would cause this to be a recent issue. Weird.

This PR fixes the issue. I am not sure how to add a regression test though. For now I'll use a patched nbgrader in production, but if the PR could be merged, that would be great. Thanks in advance!

### DIFF
--- a/nbgrader/server_extensions/assignment_list/handlers.py
+++ b/nbgrader/server_extensions/assignment_list/handlers.py
@@ -40,7 +40,7 @@ class AssignmentList(LoggingConfigurable):
 
         config_found = False
         full_config = Config()
-        for config in NbGrader._load_config_files("nbgrader_config", path=paths, log=self.log):
+        for config, filename in NbGrader._load_config_files("nbgrader_config", path=paths, log=self.log):
             full_config.merge(config)
             config_found = True
 
@@ -60,7 +60,7 @@ class AssignmentList(LoggingConfigurable):
 
         # now cd to the full assignment directory and load the config again
         with chdir(assignment_dir):
-            for new_config in NbGrader._load_config_files("nbgrader_config", path=[os.getcwd()], log=self.log):
+            for new_config, filename in NbGrader._load_config_files("nbgrader_config", path=[os.getcwd()], log=self.log):
                 config.merge(new_config)
             yield config
 
--- a/nbgrader/server_extensions/course_list/handlers.py
+++ b/nbgrader/server_extensions/course_list/handlers.py
@@ -49,7 +49,7 @@ class CourseListHandler(IPythonHandler):
 
         config_found = False
         full_config = Config()
-        for config in NbGrader._load_config_files("nbgrader_config", path=paths, log=self.log):
+        for config, filename in NbGrader._load_config_files("nbgrader_config", path=paths, log=self.log):
             full_config.merge(config)
             config_found = True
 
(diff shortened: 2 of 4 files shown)


nthiery:
Thanks for the quick review :-)

willingc:
> I am quite surprised this did not strike anyone earlier, and I did not find any recent change in IPython's traitlets or nbgrader that would cause this to be a recent issue. Weird.

This strikes me as odd too. The PR looks fine.

I'm wondering though if there is another change in nbgrader that caused the error. Looking at the traitlets code it appears to return a Config object in its merge function. 

@jhamrick I wonder if server_extensions should have been imported for use with `def all_configurable_classes(self)` https://github.com/jupyter/nbgrader/commit/da642213882d6759772e2579a37ee7fa5c67d3ca#diff-1e25ed8caaadbe3bb17baa10bc1bd456R20

lzach:
I found the cause of the sudden breakage. There was recently a new release of traitlets (7 days ago, according to https://github.com/ipython/traitlets/releases). This release included a commit https://github.com/ipython/traitlets/commit/eaf1ae862cf7f6a1654d164d8526fb7d80528f4a#diff-fe5444e44c82c0d9e20f0f7efe9042d7 (from May 2017) that changed the return value of _load_config_files to be a tuple of (config, filename). If traitlets<= 4.3.2  are used, this fix will break.

lzach:
@nthiery my suggestion would be to set traitlets>=4.3.3 in the requirements. As the method is meant to be internal to the Application class, there's always a chance that it will change in the future, but I think the easiest solution for now is to just ensure that we're using a version of traitlets 4.3.3 or higher.

nthiery:
Ah, good, this explains that. I should have thought that a 2017 change could have been released only recently.
I added the dependency on traitlets >= 4.3.3 in the setup.py.

Of course, this is all more like a work around. In a perfect world, `load_config` would be shared among the various server extensions and we would not use an internal traitlets function. But those would be more extensive changes (I actually started doing them but then it felt too extensive of the purpose).

willingc:
@minrk Looks like the 4.3.3 release of traitlets changed `_load_config_files` return type. I looked at the release tag and there appears to be 291 commits to master since the 4.3.3 tag of traitlets. Does that make sense to you?

minrk:
Yeah, 4.3.3 is a backport of some fixes (there's loads in master for 5.0 that's not ready yet), and it includes this change to a private API, needed for one of the fixes. I was afraid of exactly this issue when making 4.3.3, but noticed that IPython had a fix a couple years ago. I didn't realize that nbgrader also used this private API.

See [this IPython patch](https://github.com/ipython/ipython/pull/10676/files) from a couple years ago for removing calls to this private API and sticking with public APIs.

willingc:
@nthiery Would you be willing to update your PR with Min's suggested approach? Thanks.

jhamrick:
Thanks so much for investigating this @nthiery (and sorry to be slow on reviewing this)! I think this is now not needed since I've merged #1245 so I will go ahead and close this.

PR #1245, opened 2019-10-17
Move away from using the internal Traitles API to load default configuration.

BertR:
As discovered by @nthiery, the changed internal API in https://github.com/ipython/traitlets
broke the nbgrader build.

This PR implements the change proposed by @minrk, based on the change
to went into ipython here https://github.com/ipython/ipython/pull/10676/files


nthiery:
I don't know enough the internals to foresee potential side effects, but the code certainly looks nicer after this change!

nklever:
IMHO Issue #1247 will also be solved with this PR

jhamrick:
Awesome, thanks!

jhamrick:

PR #1258, opened 2019-11-02
add pytest-testmon to allow only affected tests to be rerun

### LINKED ISSUE #1252: Consider adding caching to the tests so they don't need to run exhaustively
For example, with the pytest-cache extension.

HanTeo:
Introducing [Pytest Testmon](https://testmon.org) into the project to improve developer experience in running `pytest`. 

pytest-testmon is a pytest plugin which selects and executes only tests you need to run. It does this by collecting dependencies between tests and all executed code (internally using Coverage.py) and comparing the dependencies against changes. testmon updates its database on each test execution, so it works independently of version control.

This fixes #1252

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -111,3 +111,4 @@ gradebook.db
 package-lock.json
 .goutputstream*
 
+.testmondata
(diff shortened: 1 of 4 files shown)


HanTeo:
trying `testmon` sans `xdist`.

jhamrick:
After looking at this further, we've determined that using the incremental tests doesn't work reliably so we are closing this PR. Too bad, it would have been a nice feature to have 🙁

PR #1259, opened 2019-11-02
Remove Python 2 compatibility code

### LINKED ISSUE #1251: Remove all python 2 compatibility code
This should be done after #1082.

For example, remove all imports of six, or any other edge cases where python 2 is handled differently from python 3.

aliniknejad:
Removing references to six and a couple of python 2 imports

Fixes #1251

### DIFF
--- a/nbgrader/apps/api.py
+++ b/nbgrader/apps/api.py
@@ -2,7 +2,6 @@ import glob
 import re
 import sys
 import os
-import six
 import logging
 import warnings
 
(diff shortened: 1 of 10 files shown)


jhamrick:
This looks great, thanks a lot!

I noticed one other place that has a python 2 thing, which is the `to_bytes` function in utils.py. Would you mind updating that as well while you're at it?

aliniknejad:
Sure, I will. Thanks Jess :)

PR #1267, opened 2019-11-02
Improve the error message on the assignments page

enisnazif:
Improves the error message on the assignments page in the case that there is an ExchangeError.


jhamrick:
Thanks!

jhamrick:

PR #1270, opened 2019-11-02
Add requirements for readthedocs

jhamrick:
readthedocs doesn't know about sphinx_autodoc_typehints, but we don't want to add it into the regular dependencies, so I'm creating a special dependencies file for readthedocs specifically.


jhamrick:

PR #1278, opened 2019-11-03
Fix course list hanging when exchange has not been created 

enisnazif:
Fixes #1278

### DIFF
--- a/nbgrader/server_extensions/course_list/handlers.py
+++ b/nbgrader/server_extensions/course_list/handlers.py
@@ -175,8 +175,6 @@ class CourseListHandler(IPythonHandler):
                 local_courses = yield self.check_for_local_formgrader(config)
                 jhub_courses = yield self.check_for_jupyterhub_formgraders(config)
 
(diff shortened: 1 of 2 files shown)


nbgrader/server_extensions/course_list/handlers.py
```diff
@@ -75,9 +75,13 @@ def check_for_local_formgrader(self, config):
             self.log.error(traceback.format_exc())
             raise gen.Return([])
 
+        coursedir = CourseDirectory(config=config)
+        if not coursedir.course_id:
```
jhamrick: I am not sure this if statement has an effect, since the `gen.Return` is not actually raised. I think you can just remove it, since the javascript handles the case where the course id is missing.

enisnazif: ah yes, good point! I've removed it, thanks

jhamrick: Thanks!

jhamrick:

PR #1313, opened 2020-02-06
Improve formatting for high point value questions and assignments.

jperryhouts:
I have a class in which I'm offering 4.54 billion available points (age of the Earth). Each assignment is then worth a couple hundred million points, which presents a challenge in the nbgrader interface. I've found the point value fields way too small to be useful for this purpose. This pull request does three things:
* Make the input cells wider.
* Improve the point total, by switching it to a `<div>` element with a pretty-formatted number.
* Allow cell headers to expand vertically to fit their content. This improves compatibility with Firefox, where the cell toolbar is otherwise awkwardly small for some reason.
Screenshot from 2020-02-05 19-57-36

### DIFF
--- a/nbgrader/nbextensions/create_assignment/create_assignment.css
+++ b/nbgrader/nbextensions/create_assignment/create_assignment.css
@@ -3,7 +3,7 @@ div.nbgrader-points, div.nbgrader-id {
 }
 
 input.nbgrader-points-input {
-    width: 3.5em;
+    width: 8em;
     height: 22px;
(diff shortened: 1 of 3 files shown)


jperryhouts:
I've since found that there's no reason to use high point values in nbgrader assignments, and I've reverted to using scaled versions of the full point value. This PR also didn't include some important changes that would have made it more usable. Closing it for now, unless someone else chooses to pick it back up.

PR #1319, opened 2020-03-23
use rapidfuzz instead of fuzzywuzzy

maxbachmann:
FuzzyWuzzy is GPLv2 licensed which would force you to licence the whole project under GPLv2. I had the same problem on one of my projects and so I wrote [rapidfuzz](https://github.com/rhasspy/rapidfuzz) which is implementing the same algorithm but is based on a version of fuzzywuzzy that was MIT Licensed and is therefor MIT Licensed aswell, so it can be used in here without forcing a License change. As a nice bonus it is fully implemented in C++ and comes with a few Algorithmic improvements making it between 5 and 100 times faster than FuzzyWuzzy. (actually even faster since python-Levenshtein was not used)

### DIFF
--- a/nbgrader/converters/base.py
+++ b/nbgrader/converters/base.py
@@ -5,6 +5,7 @@ import shutil
 import sqlalchemy
 import traceback
 
+from rapidfuzz import fuzz
 from traitlets.config import LoggingConfigurable, Config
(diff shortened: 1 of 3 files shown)


willingc:
I'm going to merge. If anyone experiences any regressions from the change, please file an issue. Thanks.

PR #1324, opened 2020-04-10
Ensure errors are written to cell outputs to prevent the autograder from awarding points for failed tests

elesiuta:
If a student disables showtraceback by including:
```python
import IPython.core.interactiveshell
IPython.core.interactiveshell.InteractiveShell.showtraceback = lambda *args, **kwargs : None
```
somewhere in their code, errors won't be recorded in cell outputs, therefore passing and receiving points for those test cases.

If the student provides attempted solutions for each answer that look plausible, this would be hard to notice during manual grading. They could further obfuscate this by representing this exploit as an integer, fetch it from elsewhere using urllib, etc then running it with exec(). The easiest way an instructor could detect this now without having to update nbgrader would probably be to grep all their submissions for "showtraceback" or "exec".

### DIFF
--- a/nbgrader/preprocessors/execute.py
+++ b/nbgrader/preprocessors/execute.py
@@ -1,4 +1,4 @@
-from nbconvert.preprocessors import ExecutePreprocessor
+from nbconvert.preprocessors import ExecutePreprocessor, CellExecutionError
 from traitlets import Bool, List, Integer
 from textwrap import dedent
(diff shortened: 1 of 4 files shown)


elesiuta:
Sure, I just committed the test for this

jhamrick (approved):
Looks good to me, thanks for the fix!

Issue #1326, opened 2020-04-15
assignment_dir issues


bomma:
The `c.Exchange.assignment_dir` config setting is not behaving as expected when fetching assignments through the web interface (assignment list). We have a Jupyterhub setup where the notebooks of the users are placed in `~/Jupyter`:

in `jupyterhub_config.py`:
`c.Spawner.notebook_dir = '~/Jupyter'

Additionally, we use following settings in `nbgrader_config.py` of the normal users:

```
c = get_config()
c.CourseDirectory.course_id = "somecourse"
c.Exchange.path_includes_course = True
```

### not setting `c.Exchange.assignment_dir`
When fetching assignments though the web interface, the files are placed in the home directory of users, instead of in `~/Jupyter`

e.g. `~/somecourse/ps1` instead of `~/Jupyter/somecourse/ps1`

### setting `c.Exchange.assignment_dir` to a the `Spawner.notebook_dir`
Files are placed in `~/Jupyter/Jupyter/somecourse/ps1` rather than `~/Jupyter/somecourse/ps1`

### setting `c.Exchange.assignment_dir` to a relative path
e.g. `c.Exchange.assignment_dir = 'foo'`

Files are placed in `~/foo/ps1` instead of `~/Jupyter/foo`

### setting `c.Exchange.assignment_dir` to an absolute path
We initially solved the issue by setting
```c.Exchange.assignment_dir = os.path.expanduser("~/Jupyter")```

This places files in `~/Jupyter/somecourse/ps1` as expected.

However, this introduces a bug with the `(view feedback)` links. They now link to:

`https://hostname/jupyter/user/someuser/tree/home/someuser/Jupyter/somecourse/ps1/feedback/2020-04-14%2018:01:49.392273%20UTC`

Note the absolute path following `tree`

Where we should have 

`https://hostname/jupyter/user/someuser/tree/somecourse/ps1/feedback/2020-04-14%2018:01:49.392273%20UTC`

(removal of `home/someuser/Jupyter/`)

Note this does not happen with the links to the notebooks within a fetched course. The paths used to generate the links are relative in:

```
course_id: "somecourse"
assignment_id: "ps1"
status: "fetched"
path: "somecourse/ps1"
notebooks: [
  {notebook_id: "problem1", path: "somecourse/ps1/problem1.ipynb"},
   …]
```

The paths listed for feedback are absolute in `local_feedback_path: /home/someuser/Jupyter/somecourse/ps1/feedback/...`

### Recap

[...]

PR #1327, opened 2020-04-16
make paths relative to notebook_dir for use with jupyterhub

bomma:
fixes both issues described in https://github.com/jupyter/nbgrader/issues/1326


moorepants:
This change probably needs some unit tests to act as regression tests.

PR #1329, opened 2020-05-09
Update the test matrix on Travis to Python 3.6+

willingc:
- Drop 3.5
- Disable mypy
- allow nightly to fail


willingc:
This PR gets tests passing again. Going to merge and we can work on reenabling mypy in a different PR.

PR #1343, opened 2020-06-25
Update deprecated jquery functions and update jquery

tmetzl:
According to jquery, the .load() function is [deprecated since release 1.8 (2012)](https://api.jquery.com/load-event/).
I think after 8 years it might be time to adopt this.

I changed all the occurences of:
```
$(window).load(function () {
..
});
```
to

```
$(window).on('load', function () {
...
});
```

While doing this I also updated the jquery version to 3.5.1.

Any comments are appreciated.


choldgraf:
This seems pretty reasonable to me, particularly if JQuery is out-of-date and we're using deprecated functions. I'm not sure if I'm the best person to confirm the changes though, so perhaps someone w/ more experience with nbgrader could weigh in?

tmetzl:
This somehow broke the striped tables in the *Manage Assignment* and *Manage Students* tabs. Enclosing the table rows in a ```<tbody>``` tag should fix it.

PR #1369, opened 2020-09-15
Pin nbconvert to 5.6.1, traitlets to 4.3.3 and pytest to <6.0.0

### LINKED ISSUE #1367: nbconvert version 6.0.0 breaks nbgrader feedback
Basically, nbconvert version 6.0.0 breaks nbgrader's `generate_feedback`. Need to pin nbconvert to 5.6.1 for now for compatibility.

### `nbgrader --version`
Python version 3.8.5 | packaged by conda-forge | (default, Aug 29 2020, 01:22:49) 
[GCC 7.5.0]
nbgrader version 0.6.1

6.1.3

### Expected behavior
Should be able to run `generate_feedback`

### Actual behavior
Can't run `generate_feedback` -- incompatibility with `HTMLExporter` from latest nbconvert (versions 6 and later)

Downgrading nbconvert to 5.6.1 resolves the issue.

Try using `generate_feedback` with nbconvert 6.0.0 or later installed. Note that these were in pre-release until just yesterday.

BertR:
Fixes #1367

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,4 +1,4 @@
-pytest>=4.5
+pytest>=4.5,<6.0.0
 pytest-cov
 pytest-rerunfailures
 pytest-xdist==1.34.0
--- a/setup.py
+++ b/setup.py
@@ -91,7 +91,7 @@ setup_args = dict(
         "python-dateutil",
         "jupyter",
         "notebook>=4.2",
-        "nbconvert>=5.6",
+        "nbconvert==5.6.1",
         "nbformat",
         "traitlets",
         "jupyter_core",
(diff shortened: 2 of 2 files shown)


rkdarst:
Is there documentation for why the other pins were made, or is it a generic something broke without deeper investigation at the time?

BertR:
> Is there documentation for why the other pins were made, or is it a generic something broke without deeper investigation at the time?

It was definitely the latter. I would welcome any PRs that would let us upgrade the dependencies. Is this something you could look at @rkdarst ?

rkdarst:
I submitted the forward-port to nbconvert 6.0 which so far seems working, #1405 

Interestingly, I am running with traitlets 5.0.5 and it seems to work now.  Maybe it was an upstream problem that got fixed since then?  (though my fork is based on upstream 4c5d6f60, so the nbgrader-incompatible part could have been )

No info on pytest yet, but I haven't had much luck running the tests at all...

PR #1376, opened 2020-10-06
Make preprocessors of generate assignment, autograde and generate feedback configurable

tmetzl:
This would allow users to add custom preprocessors to the autograde and generate assignment converters without maintaining a fork of nbgrader. 

Any comments are appreciated.

### DIFF
--- a/nbgrader/converters/autograde.py
+++ b/nbgrader/converters/autograde.py
@@ -56,14 +56,14 @@ class Autograde(BaseConverter):
         OverwriteKernelspec,
         OverwriteCells,
         CheckCellMetadata
-    ])
+    ]).tag(config=True)
     autograde_preprocessors = List([
         Execute,
         LimitOutput,
         SaveAutoGrades,
         AssignLatePenalties,
         CheckCellMetadata
-    ])
+    ]).tag(config=True)
 
     preprocessors = List([])
 
--- a/nbgrader/converters/base.py
+++ b/nbgrader/converters/base.py
@@ -7,7 +7,7 @@ import traceback
 
 from rapidfuzz import fuzz
 from traitlets.config import LoggingConfigurable, Config
-from traitlets import Bool, List, Dict, Integer, Instance, Type
+from traitlets import Bool, List, Dict, Integer, Instance, Type, Any
 from traitlets import default
 from textwrap import dedent
 from nbconvert.exporters import Exporter, NotebookExporter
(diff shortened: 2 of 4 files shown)


jgwerner:
@tmetzl 👍 - it would be great to also implement this pattern for hooks, for example `pre_generate_assignment()`, `post_generate_assignment()`, `pre_release_assignment()` ...

tmetzl:
Yes that might be a good idea.

I could implement a general converter hook that could then be configured like:
```
c.Autograde.preconvert_hook = ...
c.Autograde.postconvert_hook = ...
```
Should this be a preprocessor or a general class / function?

jgwerner:
> Should this be a preprocessor or a general class / function?

@tmetzl I could be wrong here, but from what I gather the patterns used with the Jupyter / Tornado / Traitlets setup is something like:

- Define the hook as a configurable from a traitlet type
- Run the hook if it's defined as the app's config
- Implement the hook within a function

The [pre_save_hook](https://github.com/jupyter/notebook/blob/70d74d21ac051fbeaa81d4ef4fff9fa759de96d7/notebook/services/contents/manager.py#L86) from the Jupyter Notebook project and the [post_auth_hook](https://github.com/jupyterhub/jupyterhub/blob/e5a6119505f89a293447ce4e727c4bd15e86b145/jupyterhub/auth.py#L260) from the JupyterHub project seem like good examples.

What do you think? I would of course be glad to help with these other hooks, I can just piggyback on whatever you send to add more.

tmetzl:
@jgwerner 

I added hooks but the ```nbgrader_upgrade_old_db``` test is failing for me on the PR but also locally on the master branch of nbgrader. Any idea why this test case fails?

jgwerner:
@tmetzl I started debugging the tests this weekend and will get back to this mid-week. I think I'm honing in on the issue but need a little more time ... apologies I took so long in responding, things got a little crazy with work 😄

tmetzl:
@jgwerner Don't worry, just let me know if I need to make any changes :)

jhamrick (approved):

jhamrick:
Thanks, this is a good idea!

denisalevi:
How should theses hooks be configured? I defined a hook function called `post_feedback` in `hooks.py`. I tried to use it with
```
nbgrader generate_feedback <assignment-id> --GenerateFeedback.post_convert_hook=hooks.pre_feedback
```
That doesn't work. Whatever I pass via command line (or via `nbgrader_config.py`) is stored as string in the `BaseConverter.pre_convert_hook` attribute. When the `BaseConverter` tries to call it, I get
```
      File "/usr/lib/python3.9/site-packages/nbgrader/converters/base.py", line 464, in run_post_convert_hook
        self.post_convert_hook(
    TypeError: 'str' object is not callable
```

Am I missing something or is there some implementation missing? I don't know how the `traitlet` module works, but I couldn't find any code that would take care of importing whatever is passed as hook function.

tmetzl:
I am not sure if you can actually call it this way via the command line. It looks like it tries to interpret your argument as the literal string ```hooks.pre_feedback``` instead of as a function. Also I don't know how it would figure out that the hook is defined in the file ```hooks.py```.

But what works is if you specify the hook in your ```nbgrader_config.py``` like so:

```
# nbgrader_config.py

def my_hook(assignment, student, notebooks):
    print('Calling my hook with arguments:', assignment, student, notebooks)

c.GenerateFeedback.post_convert_hook = my_hook
```

denisalevi:
Ah yes of course, the config file is a Python file 😀, great, thank you!

PR #1381, opened 2020-10-17
find cell failure when stderr is used

AnotherCodeArtist:
Recently I came across multiple kernels (eg. kotlin-jupyter or tslab) that do not create cells of type `error` but rather use stream/stderr to indicate a problem. While this was apparently already fixed for grading

https://github.com/jupyter/nbgrader/blob/7f916faaa6fdad35113c0b80e818ebb87f8cf376/nbgrader/utils.py#L116-L119

it was not considered in the validator. Therefore, this PR changes the validator to be aware of this type of error-reporting.

### DIFF
--- a/nbgrader/tests/test_validator.py
+++ b/nbgrader/tests/test_validator.py
@@ -33,6 +33,14 @@ class TestValidator(object):
         ))
         return cell
 
+    def _add_error_2(self, cell):
+        cell.outputs.append(new_output(
+            "stream",
+            name="stderr",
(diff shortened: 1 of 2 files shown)


AnotherCodeArtist:
Would be cool to consider this PR. The effect of the current situation is that students allways get a success message when clicking the validation button, although nothing might work at all.

jhamrick (approved):

jhamrick:
Thanks for fixing this!

PR #1384, opened 2020-10-27
Fix migrations. 

### LINKED ISSUE #1383: Nbgrader migrations break with SQLAlchemy 1.3.20
This is related to a bugfix in SQLAlchemy. See https://github.com/sqlalchemy/sqlalchemy/issues/5669 for a similar issue
Two of the migrations (e43177bfe90b and 167914646830) depend on behaviour that
will error out in SQLAlchemy > 1.3.19 (https://docs.sqlalchemy.org/en/13/changelog/changelog_13.html#change-1.3.20-orm)

BertR:
The old code behaviour raises an error in SQLAlchemy 1.3.20 and up.
Fixes #1383

### DIFF
--- a/nbgrader/alembic/versions/167914646830_added_task_cells.py
+++ b/nbgrader/alembic/versions/167914646830_added_task_cells.py
@@ -17,9 +17,13 @@ depends_on = None
 
 
 def _get_or_create_table(*args):
-    try:
+    ctx = op.get_context()
+    con = op.get_bind()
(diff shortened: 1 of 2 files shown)


BertR:
The pipeline for Python 3.8 on Windows 10, because the output of "python -m nbgrader --version" differs in
some way from the output on Python 3.7, but I have no access to a Windows machine at this time to check.

PR #1388, opened 2020-11-09
Move from travis ci to github actions

### LINKED ISSUE #1387: Move from Travis CI to GitHub actions
With the changing policies for Travis CI, we should move to GitHub actions.
A similar migration is underway across many projects in the Jupyter ecosystem (see rf. https://github.com/jupyterhub/team-compass/issues/353)

BertR:
In line with other Jupyter projects, we're moving the CI for nbgrader from Travis to GitHub actions.
Closes #1387


dev-requirements.txt
```diff
@@ -5,7 +5,7 @@ pytest-xdist==1.34.0
 #pytest-mypy
 coverage
 selenium
-sphinx
+sphinx==3.2.1
```
jhamrick, line 8: Is pinning this required?

jhamrick (approved):
Thanks!

dev-requirements.txt
BertR: Building the docs fails with newer versions of sphinx, but that should be a different PR to this. I will remove this change in this PR.

jhamrick: Sounds good, thanks!

Issue #1311, opened 2020-02-05
Question: how to generate a solution notebook?


adimajo:
Hi all,

I would like, parallel to the feedback generation mechanism, which is done per student, to generate a solution notebook, i.e. where the solution appears systematically (and eventually all tests) but without the nbgrader specific contents (cell types and points attributed to each question, etc.).

N.B.: I've also asked that question on Google Groups but there seems to be less activity + less specificity (https://groups.google.com/forum/#!topic/jupyter/77NdmnQLoBM).

Thanks in advance,
Adrien

sigurdurb:
I don't think this had been implemented but it is a good idea we should mark it as a feature request. 
I guess since these notebooks are just json text files, you could just write a script that takes out the text you don't want and go from there.

rmjarvis:
+1 on this feature request.  Currently I manually copy the solution code into a copy of they student version.  But it would be very nice if nbgrader could automate this step for me.

PR #1394, opened 2020-11-22
Added CLI for generating solution notebooks

liuq:
Following issue #1311, I added a few classes for generating the notebooks that contain the solutions to the given assignments. Currently they're available only in the command line interface (I am currently teaching and, unfortunately, I do not have time to get into the details of the Jupyter interface) as an additional subcommand `nbgrader generate_solution <assignment_id>`.
The subcommand creates a solution directory with the notebook containing the solution.
Also I added essential documentation of the new subcommand.


BertR:
Hi @liuq, I really like this! Thanks for your work.
Let's get some input from other users and see if we can make this part of a future release.

Issue #1382, opened 2020-10-18
jinja2.exceptions.TemplateNotFound: formgrade


OptimusPrimus:
### Operating system
Ubuntu 20.04.1 LTS.

### `nbgrader --version`
nbgrader version 0.6.1

### `jupyterhub --version` (if used with JupyterHub)
n.a.

### `jupyter notebook --version` 
6.1.4

### Expected behavior
When clicking on Manual Grading > ps1 > Problem 1 > Submission #1 the interface for manual grading should open.

### Actual behavior
However, I get an error page with the error message `Sorry, the formgrader encountered an error. Please contact the administrator of the formgrader for further assistance.`

Jupyter notebook outputs the following error message:

```
 [E 16:48:47.716 NotebookApp] Uncaught exception GET /formgrader/submissions/5e4366763f3d4591a51127cca26c5df0/ (127.0.0.1)
    HTTPServerRequest(protocol='http', host='localhost:8888', method='GET', uri='/formgrader/submissions/5e4366763f3d4591a51127cca26c5df0/', version='HTTP/1.1', remote_ip='127.0.0.1')
    Traceback (most recent call last):
      File "/home/paul/miniconda3/lib/python3.8/site-packages/tornado/web.py", line 1701, in _execute
        result = method(*self.path_args, **self.path_kwargs)
      File "/home/paul/miniconda3/lib/python3.8/site-packages/tornado/web.py", line 3178, in wrapper
        return method(self, *args, **kwargs)
      File "/home/paul/miniconda3/lib/python3.8/site-packages/nbgrader/server_extensions/formgrader/base.py", line 108, in wrapper
        return f(self, *args, **kwargs)
      File "/home/paul/miniconda3/lib/python3.8/site-packages/nbgrader/server_extensions/formgrader/base.py", line 117, in wrapper
        return f(self, *args, **kwargs)
      File "/home/paul/miniconda3/lib/python3.8/site-packages/nbgrader/server_extensions/formgrader/handlers.py", line 125, in get
        html, _ = self.exporter.from_filename(filename, resources=resources)
      File "/home/paul/miniconda3/lib/python3.8/site-packages/nbconvert/exporters/exporter.py", line 181, in from_filename
        return self.from_file(f, resources=resources, **kw)
      File "/home/paul/miniconda3/lib/python3.8/site-packages/nbconvert/exporters/exporter.py", line 199, in from_file
[...]

Issue #1373, opened 2020-09-29
Formgrader error when trying to manually grade a specific submission: "Sorry, the formgrader encountered an error. Please contact the administrator of the formgrader for further assistance. "


LaurentHayez:
### Operating system
Ubuntu 20.04.1 LTS

### `nbgrader --version`
0.6.1

### `jupyterhub --version` (if used with JupyterHub)
1.0.0

(I'm getting a weird error when typing this:
```ERROR:asyncio:Task exception was never retrieved
future: <Task finished coro=<JupyterHub.launch_instance_async() done, defined at /opt/tljh/user/lib/python3.7/site-packages/jupyterhub/app.py:2477> exception=SystemExit(0)>
Traceback (most recent call last):
  File "/opt/tljh/user/lib/python3.7/site-packages/jupyterhub/app.py", line 2492, in launch_instance
    loop.start()
  File "/opt/tljh/user/lib/python3.7/site-packages/tornado/platform/asyncio.py", line 132, in start
    self.asyncio_loop.run_forever()
  File "/opt/tljh/user/lib/python3.7/asyncio/base_events.py", line 539, in run_forever
    self._run_once()
  File "/opt/tljh/user/lib/python3.7/asyncio/base_events.py", line 1775, in _run_once
    handle._run()
  File "/opt/tljh/user/lib/python3.7/asyncio/events.py", line 88, in _run
    self._context.run(self._callback, *self._args)
  File "/opt/tljh/user/lib/python3.7/site-packages/jupyterhub/app.py", line 2479, in launch_instance_async
    await self.initialize(argv)
  File "/opt/tljh/user/lib/python3.7/site-packages/jupyterhub/app.py", line 2085, in initialize
    super().initialize(*args, **kwargs)
  File "/opt/tljh/user/lib/python3.7/site-packages/traitlets/config/application.py", line 86, in inner
    return method(app, *args, **kwargs)
  File "/opt/tljh/user/lib/python3.7/site-packages/traitlets/config/application.py", line 326, in initialize
    self.parse_command_line(argv)
  File "/opt/tljh/user/lib/python3.7/site-packages/traitlets/config/application.py", line 86, in inner
    return method(app, *args, **kwargs)
  File "/opt/tljh/user/lib/python3.7/site-packages/traitlets/config/application.py", line 693, in parse_command_line
    self.exit(0)
  File "/opt/tljh/user/lib/python3.7/site-packages/traitlets/config/application.py", line 827, in exit
[...]

PR #1405, opened 2021-01-11
Fix missing templates by porting to nbconvert 6.*

rkdarst:
nbconvert 6.* broke nbgrader somehow.  This uses hints from  #1373 to fix the problem.  Individual commit messages describe what is happening, but basically I change some template names and fix a variable.

Previously it was pinned to solve the problem, but a pin is eventually going to cause un-installable software when combined with other things.  This is one of the banes of my daily work, so we need to do something about that.

Closes:  #1373, Closes: #1382

Review:
- I am by no means an expert here, so I can't vouch for if I have made the correct answers to things.  I have done some testing, and gotten others to as well, but it's not fully tested manually.
- Check if CI passes (I haven't gotten tests to run locally yet).


rkdarst:
Also fixes #1367.

rkdarst:
I gave up on this.

I got most of the problems fixed (took some work but wasn't too hard), but the feedback template has problems remaining.  I will elaborate later, but at first it seems that, while nbconvert tries to preserve a backwards-compatible template, the interaction with nbgrader customizations doesn't really work properly.  Before I could get there, I had to work around some problems with the nbconvert `classic` template not finding `static/style.css` from within the `classic/` directory - I think there is some bug with template paths and so on.

I will try to document my findings better later.  We could use someone that understands the nbgrader feedback template customizations.

BertR:
> I gave up on this.
> 
> I got most of the problems fixed (took some work but wasn't too hard), but the feedback template has problems remaining. I will elaborate later, but at first it seems that, while nbconvert tries to preserve a backwards-compatible template, the interaction with nbgrader customizations doesn't really work properly. Before I could get there, I had to work around some problems with the nbconvert `classic` template not finding `static/style.css` from within the `classic/` directory - I think there is some bug with template paths and so on.
> 
> I will try to document my findings better later. We could use someone that understands the nbgrader feedback template customizations.

Thanks for this! Please share your thoughts when you have the time to we kind think about how we could progress this.

rkdarst:
Here is what I remember:

- nbconvert template system completely reworked (see link in #1367)
- You might think you could simply change the template you inherit from.  I did this, and it got further.
- But, the nbgrader template seemed to have been made by copy-pasting from the nbconvert templates.  Most likely the old system wasn't flexible enough. Since the new compatibilty templates are slightly different, it doesn't work as a drop-in replacement.  I took and edited some, trying to port (what I thought were) the nbgrader specific parts to the new template system.
- But then... there was another template error!  I seem to recall it couldn't find `static/style.css` from within the template.  I didn't dig too deep, but I think it is something like this: `.../templates/` is added to the template search path.  The template `classic` tries to load `static/style.css`, but the path to that is `classic/static/style.css`, but tries to load it by a *relative* path.  But the relative path doesn't work, when called from nbgrader.  Somehow this is all tied up with template inheritance.
- I think I tried to run some manual nbconvert commands, specifying the other templates, to reproduce the `static/style.css` problem (to verify if it is upstream or not).  I think I decided it was an upstream problem, but I don't have those commands anymore, so I would try to verify before trusting this (also since more time has passed).
- I convinced myself that it was better to take a break, and see if something upstream could be worked out, then come back to it.

If anyone wants to work on this, please do!  Also feel free to fix up this PR and make it work, even if I don't have time to work on it.

jhamrick:
I have started trying to work on this too. So far have just fixed a few broken tests but haven't actually looked at the template stuff yet. Here are my changes for reference: https://github.com/jhamrick/nbgrader/commits/fix-nbconvert Hopefully I'll have some more time tomorrow to look at this too.

jhamrick:
And I should say as well, thanks so much @rkdarst for the work you've put into trying to fix it so far! Hopefully we can figure out the last bit that's needed and remove this dependency on the old version of nbconvert...

jhamrick:
I think I've figured this out! The missing piece was that in the new version of nbconvert, it expects a file called `confg.json` in the directory where the templates are, and this json file determines what the base template is you are inheriting from. Setting this to the `'classic'` value seems to have fixed the inheritance problems and it no longer has trouble finding the relevant static assets.

I'll make a new PR shortly and hopefully all the tests will still pass with that.

Thanks again for all the work you put into this @rkdarst , it was hugely helpful to have as a place to start from!

jhamrick:
Closing as this is superseded by #1421

PR #1410, opened 2021-02-05
partial credit returns zero when score is zero

### LINKED ISSUE #1374: Partial credit -- assigning zero marks gives full credit
When returning zero from a cell as partial credit, this awards full marks, due to the use of `> 0.0` rather than `>= 0.0` in the code below:

https://github.com/jupyter/nbgrader/blob/ed98bc9ddffec5900f99e310434d6f3c8cedc9b7/nbgrader/utils.py#L87-L94

e.g. consider the following:
```
score = test_one(...)
score += test_two(...)
score += test_three(...)

score
```

This may output zero if all tests fail.

kcranston:
This fixes the partial credit bug where the autograder would return full credit when the output was 0. 

closes #1374

### DIFF
--- a/nbgrader/tests/utils/test_utils.py
+++ b/nbgrader/tests/utils/test_utils.py
@@ -141,6 +141,10 @@ def test_get_partial_grade():
     test_data = { "data": { "text/plain": "6.0" } }
     assert utils.get_partial_grade(test_data,10.0) == 6.0
 
+    # test zero
+    test_data = { "data": { "text/plain": [ "0.0" ] } }
+    assert utils.get_partial_grade(test_data,1.0) == 0.0
+
     # test string in list, should assume partial credit not intended
     test_data = { "data": { "text/plain": [ "'this is a string'" ] } }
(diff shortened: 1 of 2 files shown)


kcranston:
I honestly don't understand the codecov failures. I changed one line of code, and added to the relevant test function.

lwasser:
👋 @BertR @jld23 and @jhamrick ! this fixes a bug in the partial credit code that we (@kcranston ) originally submitted after working with Jess on the idea. This bug has been causing us and others some issues when grading assignments . if there is any possibility of merging this and releasing a patch, we'd greatly appreciate it!!

many thanks for your work on nbgrader. we really value being able to use it for grading in our courses.

BertR:
Thanks for this contribution and fixing this bug. I'll merge this into master and start the conversation about cutting a new release, perhaps we can backport it into the 0.6 branch.

lwasser:
hi @BertR  is there any chance of backporting this or releasing a patch? Just checking in. we use nbgrader on our hub for grading so it would be wonderful to have this fix in place.

PR #1420, opened 2021-03-17
Fix various SQLAlchemy errors and warnings

jhamrick:
Various things seem to have changed in SQLAlchemy, which was causing some test cases to fail as well as many warnings. I have done my best to fix these:

- Ensure enum values are validated
- Remove unnecessary 'aliased' call which was causing errors
- Fix syntax for multiple joins, which was causing errors
- Fix relationships for different types of BaseCells, Grades, and Comments, which was causing warnings
- Use the inspector to get the table names (was causing warnings), and fix this in the alembic migration scripts too
- Use scalar subqueries where appropriate to suppress warnings

This does not fix all of the bugs. There is a strange one that seems to be failing where a submitted notebook doesn't have a student associated with it, and I haven't been able to figure out why. But I am not sure when I will have time to do more work on this so I figured I'd go ahead and make a PR with what I have!

### DIFF
--- a/nbgrader/alembic/versions/167914646830_added_task_cells.py
+++ b/nbgrader/alembic/versions/167914646830_added_task_cells.py
@@ -19,7 +19,7 @@ depends_on = None
 def _get_or_create_table(*args):
     ctx = op.get_context()
     con = op.get_bind()
-    table_exists = ctx.dialect.has_table(con.engine, args[0])
+    table_exists = ctx.dialect.has_table(con, args[0])
 
     if not table_exists:
         table = op.create_table(*args)
--- a/nbgrader/alembic/versions/e43177bfe90b_added_course_table_and_relationships.py
+++ b/nbgrader/alembic/versions/e43177bfe90b_added_course_table_and_relationships.py
@@ -19,7 +19,7 @@ depends_on = None
 def _get_or_create_table(*args):
     ctx = op.get_context()
     con = op.get_bind()
-    table_exists = ctx.dialect.has_table(con.engine, args[0])
+    table_exists = ctx.dialect.has_table(con, args[0])
 
     if not table_exists:
         table = op.create_table(*args)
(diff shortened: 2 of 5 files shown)


jhamrick:
Sounds possibly like the failure could be related to this: https://github.com/sqlalchemy/sqlalchemy/issues/6055

Maybe we can wait to see if that gets fixed in the near future and then try again.

zzzeek:
hi there -

sorry to barge in on your issue tracker.

I would love if you were able to provide a test case for your loading failure.  we are not getting anywhere with https://github.com/sqlalchemy/sqlalchemy/issues/6055 and I cannot reproduce this person's issue.

SQLAlchemy 1.4 was in betas for three months and had thousands of downloads and we didnt see any issues like this, so I am very motivated to get to the bottom of it.

zzzeek:
hello again -

as you predicted, we fixed the issue, and hopefully it will repair the same issue for this project in release 1.4.1. Thanks for using SQLAlchemy!

jhamrick:
@zzzeek Just came back to this to see if I could help debug and I'm amazed you were able to fix it already---just installed 1.4.1 and it seems to be working (at least locally) now :) Thanks so much for fixing it so quickly!

BertR:
Hi @jhamrick , this looks good to me and it will be great to have compatibility with SQLAlchemy 1.4.
However, it does break nbgrader with SQLAchemy < 1.4. 

So I would suggest we specify that in the setup.py

```diff --git a/setup.py b/setup.py
index 9fe0494f..6d81376c 100644
--- a/setup.py
+++ b/setup.py
@@ -87,7 +87,7 @@ setup_args = dict(
         'console_scripts': ['nbgrader=nbgrader.apps.nbgraderapp:main']
     },
     install_requires=[
-        "sqlalchemy",
+        "sqlalchemy>=1.4.1",
         "python-dateutil",
         "jupyter",
         "notebook>=4.2",
```

jhamrick:
Oh yes, good point, will do.

PR #1425, opened 2021-03-26
Use NBGRADER_VALIDATING env var during autograding

### LINKED ISSUE #1415: Skip executing certain cells during grading
Is it possible to somehow skip execution of *some* cells during autograding. The issue I have is that there are a few cells (which train a NeuralNetwork) will take a long time to execute. The results of these cells are saved elsewhere (which is read by tests and graded). 

Perhaps a related discussion can be found [here](https://discourse.jupyter.org/t/is-there-a-cell-tag-convention-to-skip-execution/5445).

jhamrick:
This also sets the NBGRADER_VALIDATING environment variable during autograding, which notebooks can use to selectively choose to not run certain cells if desired. This also adds a test for this, both for the autograder and for the validator.

Fixes #1415

### DIFF
--- a/nbgrader/converters/autograde.py
+++ b/nbgrader/converters/autograde.py
@@ -189,6 +189,6 @@ class Autograde(BaseConverter):
         self._init_preprocessors()
         try:
             with utils.setenv(NBGRADER_EXECUTION='autograde'):
-                super(Autograde, self).convert_sinfgle_notebook(notebook_filename)
+                super(Autograde, self).convert_single_notebook(notebook_filename)
         finally:
             self._sanitizing = True
--- a/nbgrader/converters/autograde.py
+++ b/nbgrader/converters/autograde.py
@@ -188,6 +188,7 @@ class Autograde(BaseConverter):
         self._sanitizing = False
         self._init_preprocessors()
         try:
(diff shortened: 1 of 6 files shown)


nbgrader/converters/autograde.py
```diff
@@ -188,6 +188,7 @@ def convert_single_notebook(self, notebook_filename: str) -> None:
         self._sanitizing = False
         self._init_preprocessors()
         try:
-            super(Autograde, self).convert_single_notebook(notebook_filename)
+            with utils.setenv(NBGRADER_VALIDATING='1'):
```
rkdarst, line 191:
```suggestion
            with utils.setenv(NBGRADER_VALIDATING='autograde'):
```
What if this was set to a different value, so that one could distinguish the validation and autograde phases if one wanted?

rkdarst (approved):
This seems useful to me!

nbgrader/converters/autograde.py
jhamrick: Ah yes good idea! I did this slightly differently, by introducing a new environment variable `NBGRADER_EXECUTION` that gives the execution type (either validate or autograde). Then `NBGRADER_VALIDATING` is unchanged, and is only set during validation.

PR #1437, opened 2021-05-03
Make converter exporter class configurable

tmetzl:
This allows users to set a custom exporter class in each converter. 

Use case:
We want to enrich student feedback with custom information and need custom filters for this. The easiest way to achieve this is is by using a custom exporter as the _GenerateFeedback_ converter.

Any comments are appreciated.


jhamrick (approved):
Seems reasonable to me, thanks!

PR #1438, opened 2021-05-04
Validate pre and post convert hooks

tmetzl:
Fixing the problem mentioned in #1376. 

If the hooks are given as an import string (see example) they are now properly imported.
```
nbgrader generate_feedback <assignment-id> --GenerateFeedback.post_convert_hook="my_package.my_hook"
```

### DIFF
--- a/nbgrader/converters/base.py
+++ b/nbgrader/converters/base.py
@@ -8,7 +8,7 @@ import traceback
 from rapidfuzz import fuzz
 from traitlets.config import LoggingConfigurable, Config
 from traitlets import Bool, List, Dict, Integer, Instance, Type, Any
-from traitlets import default
+from traitlets import default, validate
 from textwrap import dedent
 from nbconvert.exporters import Exporter, NotebookExporter
 from nbconvert.writers import FilesWriter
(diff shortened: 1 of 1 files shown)


nbgrader/converters/base.py
```diff
@@ -85,6 +87,24 @@ class BaseConverter(LoggingConfigurable):
...
+        if isinstance(value, string_types):
+            value = import_item(self.pre_convert_hook)
+        if not callable(value):
+            raise TraitError("pre_convert_hook must be callable")
+        return value
+
+    @validate('post_convert_hook')
+    def _validate_post_convert_hook(self, proposal):
+        value = proposal['value']
+        if isinstance(value, string_types):
```
jhamrick, line 102: nbgrader is python 3 only these days, so I think you can just check against str here (and above as well).

nbgrader/converters/base.py
```diff
@@ -19,6 +19,8 @@
 from ..nbgraderformat import SchemaTooOldError, SchemaTooNewError
 import typing
 from nbconvert.exporters.exporter import ResourcesDict
+from ipython_genutils.importstring import import_item
```
jhamrick, line 22: I believe genutils isn't officially supported, and it would probably be better to not need to add a new dependency in any case---could you achieve the same behavior using `importlib.import_module`? See https://docs.python.org/3/library/importlib.html

jhamrick (changes requested):
Thanks for this!

tmetzl:
Thanks for pointing out the changes. I think I addressed them with my new commit.

jhamrick (approved):
Great, thanks!

PR #1443, opened 2021-05-05
Fix broken windows tests

jhamrick:
Use `python -m pip install --upgrade` rather than `pip install -U`, which causes a permission error on Windows.


BertR:
Thanks @jhamrick, this bug confused me, since it seemed the behaviour of the windows tests changed suddenly.

jhamrick:
No problem! Yeah I was confused too, and I'm not 100% sure why this fix works exactly, but I got it from the github docs so figured it would work, and indeed it did 😄

PR #1458, opened 2021-06-16
Lock setup dependencies

jgwerner:
Locks setup dependencies to patch versions.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -12,4 +12,3 @@ codecov
 nbval
 requests-mock
 wheel
-
--- a/setup.py
+++ b/setup.py
@@ -85,20 +85,20 @@ setup_args = dict(
         'console_scripts': ['nbgrader=nbgrader.apps.nbgraderapp:main']
     },
     install_requires=[
-        "sqlalchemy>=1.4.1",
-        "python-dateutil",
-        "jupyter",
-        "notebook>=4.2",
+        "sqlalchemy==1.4.18",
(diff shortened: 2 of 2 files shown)


BertR (approved):

perllaghu:
The primary reason, from my point of view, is to ensure that the application works as intended.

The other version is that up-stream updates can change APIs, and break services..... witness the fun NBGrader had with NBConvert - when it changed from 5.6.1 to 6.0.0

BertR:
> 
> I clearly have opinions here, but I'm wondering what others tend to do. I would definitely support a separate lockfile or some such that says "these versions work". Also, practically speaking, it makes more sense for nbgrader rather than other projects, since it's not actively maintained, so maybe it _needs_ to be locked, because we know it's going to break someday anyway, and no one will be around to fix it quickly.

@rkdarst : I think your last paragraph hits a good point, this was done because we wanted to make sure there was a working branch, and new CI runs were failing because of broken dependencies.
I we could do some work to establish the range of working versions, we could relax the pinning of the dependencies. eg. `notebook<7.0.0`. For some packages that are fairly stable (requests, tornado) we can consider removing the pins altogether.

Thoughts @rkdarst , @jgwerner & @perllaghu ?

jgwerner:
@BertR I would agree that we need a range of working versions to avoid conflicts. I have something locally that accomplishes this requirement. I haven't updated the requirements in a few weeks so will review the packages and send a PR for review.

PR #1474, opened 2021-08-01
Downgrade traitlets to avoid installation errors with v5.0.5

jgwerner:
Downgrades the `traitlets` package to `v4.3.3` to remove errors when running the `pip install .` command. For some reason, `pip` cannot resolve version `5.0.5` [even though it does appear as a valid version in PyPi](https://pypi.org/project/traitlets/).


jgwerner:
Closing now that traitlets have been upgraded to 5.1.0.

Issue #1292, opened 2019-11-18
empty string in token


rebenkoy:
https://github.com/jupyter/nbgrader/blob/4c5d6f606c8a2ec7b51e05f7f08797be6252f7f5/nbgrader/server_extensions/course_list/handlers.py#L60

I get empty string a self.token here.
Where the problem might be?

I am following the instructions for many courses

PR #1476, opened 2021-08-07
Fix Issue with Courses tab on Multi courses

QuantumEntangledAndy:
When using multiple courses multiple instructors the courses tab will crash with an Unauthorized error. This happens when it tries to find local form graders without a token.

The fix is to just add `ConnectionRefusedError` exception handler. In which case it will gracefully continue to looking for Jupyter hub formgraders instead of erroring out.

Fixes: #1292


BertR:
Thanks!

QuantumEntangledAndy:
Welcome and thanks for merging

PR #1493, opened 2021-09-08
docs: revise "Managing assignment files" pages

rkdarst:
- Rename to "Exchanging assignment files" - the previous name could
  mean almost anything related to nbgrader, and uses the standard
  terminology.

- Adjust the introduction sections so that each page describes the
  reason you would read it.  For the 'manual' page, there are more
  rearrangements here, since the focus has changed a lot.  But nothing
  is completely lost (though maybe it should be, to reduce redundancy)

- There is more to do to improve, but I am trying to start small and
  get some basic stuff done.

- Review: General check, but not too much, we should keep iterating.

### DIFF
--- a/nbgrader/docs/source/user_guide/managing_assignment_files.ipynb
+++ b/nbgrader/docs/source/user_guide/managing_assignment_files.ipynb
@@ -11,7 +11,9 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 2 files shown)


perllaghu:
Looks OK to me
(though the reference to an `exchange directory` is possibly misleading as there are multiple solutions to exchange through an external micro-service)

rkdarst:
Good point about multiple services.  I clarified (on the exchange directory page) that the exchange directory is only one of the options, and it is pluggable.  I said plugins would have the same UI, which may not be entirely true but close enough for this page.  And the options that don't use the plugin abilities need some other dedicated documentation, I think.

perllaghu:
Further to the docs on the exchange & it's pluggable nature - the [developer] reference in the existing docs is https://nbgrader.readthedocs.io/en/latest/exchange/

..... this may, or may not, be relevant to these docs :)

rkdarst:
I do agree that there could be some more links here, but was trying to make the user guide not link to the developer/admin docs that much

Anyone should feel welcome to directly push and add more links to any of these PRs - they could be useful.

perllaghu (approved):

Issue #1491, opened 2021-09-05
Documentation suggestions: more on notebook creation and arrange to notebooks, instructor view, student view


rkdarst:
I am at the point where I would like to write some better nbgrader docs  for my own team, but I thought perhaps I should contribute directly to nbgrader docs.  Since the changes are more significant than usual, I thought I would post an issue with the idea. (I would do most of these myself)

My fundamental goal is to make it easier to get someone started with nbgrader.  My main issue with the current documentation is that it doesn't seem to provide a gentle introduction to nbgrader for an outsider.  In particular, there aren't good examples of what a minimal notebook would look like (it is really simple, and that's the benefit!  It should be show as early as possible, and make the connection to unit testing).  Instead, most of the starting information is about the interface itself.

So, what I am thinking about right now is:
a) Interface highlights → add an example of what a notebook would look like.  Emphasize that this is like unit tests and the benefits of the single notebook format.
b) Also on "Interface highlights", divide it into three or four main topics: "creating notebooks", "managing assignments" (instructor view), "student view", and possibly "exchanging".
c) Add a "Creating notebooks" section before "Creating and grading assignments".  Move some of the material about notebooks to here.
d) Possibly rename/adjust the "User documentation" section to fit these three or four categories. I wouldn't do much, I think, but possibly try to clarify some of the names (eg. "Managing assignment files"→"Exchanging assignment files") or minor rearrangements.

Any thoughts on this, yes or no?  I doubt I would get all the way to the bottom of this list, but at least I can get started. If this isn't desired here, I would do at least a few of these in our own local (but open source) documentation.

perllaghu:
Yes, definitely*...... I suspect there are several versions of this, in various levels of completeness & usefulness

We have ours at
* Instructor side: https://noteable.edina.ac.uk/nbguide/
* student side: https://noteable.edina.ac.uk/student-guide/

.... and there's some general stuff at https://noteable.edina.ac.uk/user-guide/

If there was a complete, central, repository for all this information, that would certainly save _me_ time in the future.

It would also be good to include some templates or examples to show how `### BEGIN SOLUTION` & `### END SOLUTION` work, what `### MARK SCHEME` does, what `### HIDDEN TESTS` do, and the difference between `manually-graded answer` and `manually-graded task` cells - again, there's some stuff to possibly use at https://github.com/edina/Exemplars2020/tree/master/TeachingDocs/

[...]

PR #1494, opened 2021-09-12
docs: update highlights to introduce the notebook format

rkdarst:
Rearrange the highlights page into three sections, and introduce a section giving a minimal example of the notebook format.  This introduces some of the changes discussed in #1491.  I recommend looking at the commits separately, to easily understand what is going on and see more about each change.

To review: general text and all, also the general clarity and message of the change.


perllaghu:
I like....

(if I were to be _really_ picky, should students not `validate` before they submit?)

PR #1518, opened 2021-10-28
[converters/autograde] Fix autograded notebook permission

mhwasil:
When groupshared is enabled (`c.CourseDirectory.groupshared = True`), the default permission in `BaseConverter` class will be changed to `664`, which is expected. However, since `autograde.py` is also inherited from `BaseConverter` class in `convertes/base.py`, this will make permission of autograded notebooks to be `664` as well, which is not wanted, because those files should be read-only (444) to grader as it's when `groupshared = False`.

This PR will set the default permission of `Autograde` class to `444`, regardless of whether the groupshared is True or False.

### DIFF
--- a/nbgrader/converters/autograde.py
+++ b/nbgrader/converters/autograde.py
@@ -39,6 +39,10 @@ class Autograde(BaseConverter):
 
     _sanitizing = True
 
+    @default("permissions")
+    def _permissions_default(self) -> int:
+        return 444
+
     @property
     def _input_directory(self) -> str:
         if self._sanitizing:


jhamrick:
Thanks!

jhamrick:

jhamrick:
Ah hmm I didn't realize this actually breaks some tests---it seems that in #1000 this was intended behavior. Unfortunately I've run out of time to determine whether 644 or 444 is really the right behavior (perhaps @rkdarst remembers from the original PR?) so for now I am going to revert this until a conclusion is reached.

rkdarst:
My intention here was that multiple different users, with different user accounts on a shared system with unix permissions controlling the course directory, can share instructor responsibility.

I think that 664 would be appropriate here.  If it's 444, then when one instructor autogrades, others can not anymore, and it requires the original autograder to remove the files before others can.  Then, they'd ask the sysadmin to investigate and do something.  So, while it is theoretically better for them to be read-only since they shouldn't be modified, in practice it breaks the ability to share, which was my intention in #1000.  If that's good is another matter, but at least if it's changed here, we would need to locally patch it.

(by the way, I think there is at least one of these still in nbgrader somewhere, that instructors need my help to fix - my notes say that it is in release_feedback.  I hope that a new person working with me can take a look.  I guess the same question as here would come up: it shouldn't be modified, but if it's not group-writeable, then it can't be shared).

Thanks for pinging me!

mhwasil:
First of all, I did not know that this has been discussed a long time ago in #1000.

I agree with @rkdarst, 444 would not work in a group shared system. However, in our use case, setting 444 does not really change the behaviour of our group shared system, because we use the same UID for all graders, which is set in our spawner. Of course, this wouldn't be ideal for some other systems (e.g., the system where the UID for graders is different).

Nevertheless, we could, of course, force the base permission to 444 in our [nbgrader_config](https://github.com/DigiKlausur/e2xgrader/issues/85), and also set the other to 664 so that the generated assignment is not read-only:
```
c.BaseConverter.permissions = 444
c.GenerateAssignment.permissions = 664
c.GenerateFeedback.permissions = 664
c.GenerateSolution.permissions = 664
```

rkdarst:
Hm, I initially added this specifically for the case of multiple different UIDs accessing things. So, I wonder if you need this option on and you could leave it false, and get the behavior you need?  If so, should documentation be approved?

Hm, I wonder if there could be a difference between "shared group readable" and "shared group writeable" or "shared group writeable including feedback files".  But maybe this is thinking too hard about it, and editing the config directly like you do is a much better option.

Thanks for the thoughts about this.

mhwasil:
> Hm, I initially added this specifically for the case of multiple different UIDs accessing things. So, I wonder if you need this option on and you could leave it false, and get the behavior you need? If so, should documentation be approved?
> 
Do you mean an option to set `multi` vs `single` UID in nbgrader config? If that's what you meant I think that would be a good idea. So that we would not have to modify the permission in the nbgrader config manually, and use this option instead.

PR #1519, opened 2021-10-29
Make generate solutions preprocessors configurable

ykazakov:
Most preprocessors were made configurable but not for the generate_solutions converter.

see #1376


BertR:
Thanks @ykazakov. Apologies it took so long to merge this.

PR #1539, opened 2022-02-08
Improve CI by running sphinx linkcheck

rkdarst:
Sphinx linkcheck reported one broken link, which caused docs CI to
fail.  This fixes the one broken link, and updates several other
redirects to https.  Other ones *could* be updated but seem reasonable
to leave as well.

Review: general check and merge (other CI is broken, too).  Check if
the docs build group passes, I'm curious if other things are broken.


rkdarst:
Well, at least the docs test passes now...

PR #1560, opened 2022-04-11
Fix selenium tests

BertR:


BertR:
Has now been addressed in https://github.com/jupyter/nbgrader/pull/1567

PR #1561, opened 2022-04-11
Add missing 'self' argument to `late_submission_penalty`

gymreklab:
When implementing my own late policy, the example code provided did not work until I added the `self` argument to my `late_submission_penalty`.


BertR:
Thanks @gymreklab !

Issue #1570, opened 2022-05-04
Adopt the Jupyter Releaser


jtpio:
It would be great if we could start using the Jupyter Releaser for making new releases: https://jupyter-releaser.readthedocs.io/en/latest/

The releaser is already being used in many projects. Using it for `nbgrader` would be great for consistency and would make the setup similar to the other projects.

The checklist for adoption: https://jupyter-releaser.readthedocs.io/en/latest/how_to_guides/convert_repo.html

jhamrick:
That's a great idea, I didn't know about Jupyter Releaser.

If you start using it could you update the releasing docs accordingly too? (They may also be helpful in ensuring all the release steps are followed when switching to Jupyter Releaser): https://nbgrader.readthedocs.io/en/stable/contributor_guide/releasing.html

brichet:
I can start setting up the Jupyter Releaser.
I don't know if I need special permissions on this repository. I'll let you know.

jtpio:
Thanks @brichet.

We should be able to start with a draft PR, by adding the config files and the `check-release` CI workflow like that one: https://github.com/jupyter-server/jupyter_server/blob/main/.github/workflows/check-release.yml

The check release workflow will help us know if the repo is correctly setup and ready to be released with the releaser.

PR #1571, opened 2022-05-06
Add workflow to enforce GitHub labels

jtpio:
Similar to the workflows used in other Jupyter projects:

- https://github.com/jupyterlab/jupyterlab/blob/master/.github/workflows/enforce-label.yml
- https://github.com/jupyter-server/jupyter_server/blob/main/.github/workflows/enforce-label.yml

This will be useful when adopting the Jupyter Releaser (#1570), so each PR has a label and the changelog be nicely generated.

### DIFF
--- a/.github/workflows/enforce-label.yml
+++ b/.github/workflows/enforce-label.yml
@@ -1,0 +1,13 @@
+name: Enforce PR label
+
+on:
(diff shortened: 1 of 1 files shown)


jhamrick:
Thanks!

PR #1572, opened 2022-05-06
Fix a false positive test

brichet:
This PR fix a false positive in `test_create_assignment`.

Sometime the `blank` notebook was loaded instead of the `task` notebook, which made the test pass.


jhamrick:
Thanks!

PR #1577, opened 2022-05-09
Set up jupyter releaser

brichet:
This PR aims to use jupyter releaser for futures releases of nbgrader (#1570)
- uses setup.cfg file
- adds check-release workflow in CI
- inits tbump
- moves changelog from documentation directory to root directory


jtpio:
Thanks @brichet for starting this!

pyproject.toml
```diff
@@ -0,0 +1,82 @@
...
+    "nbval",
+    "requests-mock",
+    "wheel",
+]
+
+[project.scripts]
+nbgrader = "nbgrader.apps.nbgraderapp:main"
+
+[tool.tbump.version]
+current = "0.8.0.dev"
```
davidbrochart, line 67: Isn't it `0.7.0`?

pyproject.toml
```diff
@@ -0,0 +1,82 @@
...
+  (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)
+  ((?P<channel>a|b|rc|.dev)(?P<release>\d+))?
+'''
+
+[tool.tbump.git]
+message_template = "Bump to {new_version}"
+tag_template = "v{new_version}"
+
+[[tool.tbump.file]]
+src = "nbgrader/__init__.py"
```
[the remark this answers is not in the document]
davidbrochart: But I think you should have `version_info = (0, 7, 0,)` 
 [on this line](https://github.com/jupyter/nbgrader/blob/2e1df794eda5a5ce3c00ed34360542ab0a7783e0/nbgrader/_version.py#L1), to reflect the current version.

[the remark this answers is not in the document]
davidbrochart: Actually I think you need an explicit `__version__ = "0.0.7"` in `nbgrader/__init__.py` for tbump to work.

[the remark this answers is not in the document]
brichet: Thanks @davidbrochart, let's try it.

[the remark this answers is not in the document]
brichet: Actually I don't understand the error : 
`Error: Current version string: (0.7.0) not found in setup.py`

The version is already explicit in `nbgrader/_version.py` (imported in `nbgrader/__init__.py`), in `pyproject.toml` and in `tbump.toml`.
Do I have to explicit it in `setup.py` too ?
cc @jtpio

nbgrader/_version.py
```diff
@@ -1,2 +1,2 @@
-version_info = (0, 8, 0, 'dev')
+version_info = (0, 8, 0)
```
jtpio, line 1: Probably we'll want to stick to what is on `master` currently? And let the releaser do the bump to the new version on release.

brichet: I agree but I'm not sure how to handle this using `tbump`. 
`tbump` uses python string format() to patch version, and complains if I try to change from `(0, 8, 0, "dev")` to `(0, 8, 0)` :

```
Error: nbgrader/_version.py: refusing to replace by version containing 'None'
More info:
 * version groups:  {'major': '0', 'minor': '8', 'patch': '0', 'channel': None}
 * template:        ({major}, {minor}, {patch}, "{channel}")
 * version:         (0, 8, 0, "None")
```

Does the releaser handle this ? Otherwise, it means that a *channel* must be included for future versions.

.github/workflows/check-release.yml
```diff
@@ -0,0 +1,29 @@
...
+  push:
+    branches: ["master"]
+  pull_request:
+    branches: ["*"]
+
+jobs:
+  check_release:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
```
jtpio, line 12: Maybe we could set `fail-fast: false` here so the check release continues to run even if the link check fails:

```suggestion
      fail-fast: false
      matrix:
```

brichet: Thanks @jtpio

jtpio:
Nice, looks like CI is now passing :tada:

.github/workflows/check-release.yml
```diff
@@ -0,0 +1,30 @@
...
+        run: |
+          pip install -e .
+      - name: Check Release
+        if: ${{ matrix.group == 'check_release' }}
+        uses: jupyter-server/jupyter_releaser/.github/actions/check-release@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Run Link Check
+        if: ${{ matrix.group == 'link_check' }}
+        uses: jupyter-server/jupyter_releaser/.github/actions/check-links@v1
```
jtpio, line 30: Maybe we can add the additional step that uploads the built assets as artifacts. This makes it easier to troubleshoot potential issues, and make it possible to double check everything looks good manually if needed.

Similar to: https://github.com/jupyterlab/jupyterlab/blob/a41b9816fdedb936be1589dbc98363f62f5f4574/.github/workflows/check-release.yml#L37-L43

```yaml
- name: Upload Assets
  if: ${{ matrix.group == 'check_release' }}
  uses: actions/upload-artifact@v2
  with:
    name: jupyterlab-releaser-dist-${{ github.run_number }}
    path: |
      .jupyter_releaser_checkout/dist/*.*
```

pyproject.toml
```diff
@@ -0,0 +1,98 @@
...
+
+    "Framework :: Jupyter",
+    "Framework :: Jupyter :: JupyterLab",
+    "Framework :: Jupyter :: JupyterLab :: 3",
+    "Framework :: Jupyter :: JupyterLab :: Extensions",
+    "Framework :: Jupyter :: JupyterLab :: Extensions :: Prebuilt",
+]
+urls = {Homepage = "https://github.com/jupyter/nbgrader"}
+requires-python = ">=3.7"
+dependencies = [
```
jtpio, line 20: Do we need the duplication here is the same information is already available in `setup.cfg`?

nbgrader/_version.py
brichet: You are right, I had missed the `field` option

pyproject.toml
brichet: :+1:

.github/workflows/check-release.yml
brichet: :+1:

jtpio:
Kicking CI to try to get the ReadTheDocs preview: https://github.com/jupyter/nbgrader/issues/1580

brichet:
Nice, thanks @jtpio for reviews.
The documentation still needs to be updated with the use of `jupyter_releaser` and `tbump`.
I think we can merge this PR and open an issue on the documentation.

setup.cfg
```diff
@@ -0,0 +1,67 @@
...
+classifiers =
+    License :: OSI Approved :: BSD License
+    Programming Language :: Python
+    Programming Language :: Python :: 3
+
+    Framework :: Jupyter
+    Framework :: Jupyter :: JupyterLab
+    Framework :: Jupyter :: JupyterLab :: 3
+    Framework :: Jupyter :: JupyterLab :: Extensions
+    Framework :: Jupyter :: JupyterLab :: Extensions :: Prebuilt
```
jtpio, line 22: Maybe we should remove these classifiers for now, and add them back when adding the JupyterLab extension?

.github/workflows/check-release.yml
```diff
@@ -0,0 +1,35 @@
...
+      - name: Upload Distributions
+        uses: actions/upload-artifact@v2
+        with:
+          name: jupyter-releaser-dist-${{ github.run_number }}
```
jtpio, line 34: Maybe with the `nbgrader` prefix to make it more explicit when downloading the artifact?

```suggestion
          name: nbgrader-jupyter-releaser-dist-${{ github.run_number }}
```

jtpio (approved):
Looks good thanks!

Just checked the built assets uploaded as artifact and the version seems to be bumped correctly there.

jtpio:
cc @jhamrick who did the previous releases and might want to have a look too

nbgrader/docs/source/contributor_guide/releasing.rst
```diff
@@ -5,12 +5,15 @@ Releasing a new version
...
 
+This repository uses `jupyter_releaser <https://jupyter-releaser.readthedocs.io/en/latest/>`_.
+
 Prerequisites
 -------------
 
 - All PRs should be tagged with the appropriate milestone. Several of the
   releasing scripts depend on this being done.
 - You will need a GitHub API key: https://github.com/settings/tokens
+- Install packaging requirement `tbump <https://github.com/dmerejkowsky/tbump>`_ : ``pip install tbump``
```
jhamrick, line 16: I guess this should already be installed when you do the dev install, so perhaps it would be more general to say something like "make sure you have done a development install" and link to the relevant docs?

jhamrick (approved):
Looks great, thanks a lot!

jtpio:
Thanks @brichet and @jhamrick!

PR #1579, opened 2022-05-09
Only add extra_template_basedirs if it has not been set

tmetzl:
In nbgrader v0.6.x we only set the `template_path` option if it is not already in the config (see [here](https://github.com/jupyter/nbgrader/blob/0.6.x/nbgrader/converters/generate_feedback.py#L60)). 
In nbgrader v0.7.x we should also only set the the `extra_template_basedirs` conditionally in order to not break any custom exporters.

### DIFF
--- a/nbgrader/converters/generate_feedback.py
+++ b/nbgrader/converters/generate_feedback.py
@@ -57,9 +57,11 @@ class GenerateFeedback(BaseConverter):
         c = Config()
         if 'template_name' not in self.config.HTMLExporter:
             c.HTMLExporter.template_name = 'feedback'
-
-        template_path = os.path.abspath(os.path.join(os.path.dirname(__file__), '..', 'server_extensions', 'formgrader', 'templates'))
-        c.HTMLExporter.extra_template_basedirs = [template_path]
+            
+        if 'extra_template_basedirs' not in self.config.HTMLExporter:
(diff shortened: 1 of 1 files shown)


jhamrick:
Thanks @tmetzl !

Can you modify this to merge into master rather than 0.7.x? I think you just would need to cherry-pick your commit (so it excludes the '[Bump version to 0.7.1.dev](https://github.com/jupyter/nbgrader/pull/1579/commits/cfe89748cbe060e622f577b7b7018ca8b08fcaa2)' commit).

tmetzl:
@jhamrick I think I messed by dropping the commit. Should I create a new clean PR in favor of this one?

jhamrick:
I was able to fix it! 😄 

(If you're curious, here are the git commands I used to do it)

```
# get the latest changes
git fetch upstream
# create a temp branch off of master to fix the PR
git checkout master
git branch patch-1-fix
# checkout the PR, then cherry-pick the relevent commit; this gives us a branch that only has the one commit in it, and not the others
gh pr checkout 1579
git checkout patch-1-fix
git cherry-pick 61df6a5
# reset the PR branch to be the same as the temp branch (throws away the other commits)
git checkout patch-1
git reset --hard patch-1-fix 
git push -f
```

tmetzl:
Perfect 👍 
Thank you for the git commands and the comments.

jhamrick:
Tests are passing locally, so should be good to merge.

jhamrick:

PR #1586, opened 2022-05-18
Jupyter server

### LINKED ISSUE #1465: Migrate server extensions to jupyter_server
## Overview

Migrate the nbgrader [server extensions](https://github.com/jupyter/nbgrader/tree/master/nbgrader/server_extensions) to support _both_ the Jupyter classic notebook server and the  `jupyter_server`. This would be one of the first steps to allow nbgrader to work with the newer version of [Jupyter NB Classic](https://github.com/jupyterlab/nbclassic) or with [JupyterLab](https://jupyterlab.readthedocs.io/en/stable/).

## Tasks

- [ ] Add new reference name for `load_jupyter_server_extension` with `_load_jupyter_server_extension`
- [ ] Update readme/docs
- [ ] Update tests

I don't see any `nbgrader` extension files located in the `/path/to/etc/jupyter/jupyter_server_config.d` directory so it doesn't look like the `setup.py` needs to be updated, but perhaps I'm missing something.

brichet:
This PR is a starting point for using labextensions :

- Switch from IpythonHandler from notebook to JupyterHandler from jupyter_server
- Enable jupyter-server and classic notebook server extensions during installation (fix #1465)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["jupyter_packaging~=0.5.0", "setuptools>=40.8.0", "wheel"]
+requires = ["jupyter_packaging~=0.12.0", "setuptools>=40.8.0", "wheel"]
 build-backend = "setuptools.build_meta"
 
 [tool.tbump.version]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,7 @@
+[build-system]
+requires = ["jupyter_packaging~=0.5.0", "setuptools>=40.8.0", "wheel"]
+build-backend = "setuptools.build_meta"
+
 [tool.tbump.version]
 current = "0.8.0.dev"
 regex = '''
--- a/etc/jupyter/jupyter_notebook_config.json
+++ b/etc/jupyter/jupyter_notebook_config.json
@@ -1,0 +1,10 @@
+{
+  "NotebookApp": {
+    "nbserver_extensions": {
+      "nbgrader.server_extensions.formgrader": true,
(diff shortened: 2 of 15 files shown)


pyproject.toml
```diff
@@ -1,3 +1,7 @@
+[build-system]
+requires = ["jupyter_packaging~=0.5.0", "setuptools>=40.8.0", "wheel"]
```
jtpio, line 2: Wondering if we need to stick to `0.5.x` for jupyter packaging here.

Or if it could be bumped to a more recent version, for example: `jupyter_packaging~=0.12.0`

jtpio: Also of note, we will probably soon be able to drop `jupyter-packaging` and switch to `hatch` (in a later PR):

https://github.com/jupyterlab/team-compass/issues/146

brichet:
It's not ready to be merged yet, this version still uses the `notebook_dir` (notebook server property) instead of `root_dir` (jupyter server property).
At this point both servers should be usable, it must be possible to choose between these 2 properties.
I will fix this in a future commit.

brichet:
@jhamrick @BertR @jtpio do you think this PR can be merged ?

jhamrick (approved):
Looks good to me! Perhaps @jtpio can take another look as well before merging.

jtpio:
Nice thanks @brichet!

brichet:
> Also, do we already test that the extensions are loaded and enabled on CI?
> 
> Otherwise maybe we should add a check on CI that runs `jupyter server extension list` and check the extensions are correctly listed?

No, there is no test on the installation and activation of extensions yet. We may add it.

PR #1587, opened 2022-05-18
Fix readthedocs automatic build

brichet:
According to the documentation, the `readthedocs.yaml` file should be renamed `.readthedocs.yaml`.
The `readthedocs-requirements.txt` should not be necessary anymore as the yaml file can run `pip install .[docs]` now.


jhamrick (approved):
Ah great, thanks!

PR #1588, opened 2022-06-01
Jupyterlab extensions

brichet:
This PR brings the jupyterlab extensions to nbgrader.

It is an improvement of the work done by @Lawrence37 and @aalmanza1998 in https://github.com/LibreTexts/nbgrader-to-jupyterlab/tree/lab-common.

It works with jupyterlab < 4 and is still compatible with classic notebook (nbextensions still work).

Still missing some tests on labextensions (formgrader and course list), which could be added in a future PR.

@fcollonval also helped with this PR.


jhamrick:
Awesome---thanks a lot for this! Will try to have a look sometime next week...

jtpio:
Nice!

Probably this could in some way (or partially) also be tested on Binder with the following link: https://mybinder.org/v2/gh/brichet/nbgrader/jupyterlab_extensions?urlpath=lab

perllaghu:
> Nice!
> 
> Probably this could in some way (or partially) also be tested on Binder with the following link: https://mybinder.org/v2/gh/brichet/nbgrader/jupyterlab_extensions?urlpath=lab

Poodles.... I'm getting a `No space left on device` error from Binder :(

perllaghu:
I'm looking at this in a notebook, How does one access `Formgrader` & `Assignment list` sections?

If I switch to the classic interface, they're there....

brichet:
> I'm looking at this in a notebook, How does one access `Formgrader` & `Assignment list` sections?
> 
> If I switch to the classic interface, they're there....

I will update the documentation soon.
Extensions are available from command palette (`Ctrl + Shift + c`) and then you can type `formgrader` or `assignment list`, it will open a new tab in Jupyter Lab.

brichet:
> Working on documentation I just realize that there are some problems with images being duplicated. This could be related to nbconvert as the images in error are from the Notebook files in the documentation ("Creating and grading assignments" and "Exchanging assignment files" in TOC).

For traceability :
According to https://github.com/spatialaudio/nbsphinx/issues/162 it seems that using `pandoc>=2` changes the way images are linked into RST file (from Markdown). This was causing errors when using nbconvert to convert some documentation Notebook to RST (`WARNING: Duplicate substitution definition name: "image0"`). 
Whenever a markdown cell containing images was converted, the first image was substituted by "image0" definition name, causing the duplication error  if more than one cell contained images (it was the same for "image1" and so on).
I assume that using readthedoc.yaml v2 brought that behavior.
This is simply fixed by adding alternative text for the images in the markdown (https://github.com/spatialaudio/nbsphinx/issues/162#issuecomment-360977244).

perllaghu:
Everything's looking good..... I've now got the full assignment cycle running - even with my plugin exchange service!

The only thing not working now (as far as I can see) is the "View Feedback" link. The feedback is downloaded, and the link is added to the submitted item, however the link is failing.

Works fine in the classic interface, where it has [for me] an href of `...tree/<course>/<assignment>/feedback/<timestamp>`

In the lab UI, the `<a>` element is completely attribute-less: `<a> (view feedback)</a>`

brichet:
> Everything's looking good..... I've now got the full assignment cycle running - even with my plugin exchange service!
> 
> The only thing not working now (as far as I can see) is the "View Feedback" link. The feedback is downloaded, and the link is added to the submitted item, however the link is failing.
> 
> Works fine in the classic interface, where it has [for me] an href of `...tree/<course>/<assignment>/feedback/<timestamp>`
> 
> In the lab UI, the `<a>` element is completely attribute-less: `<a> (view feedback)</a>`

Thanks @perllaghu for your feedback.
From my side the `view feedback` is working, but maybe the feature could be handled better. As there can be several feedbacks, nbextension opens a `tree file` tab. 
In this extension, instead of opening a new tab, it simply opens the `file browser` (left panel) in the correct directory.
If you change the directory in the `file browser`, the link should open `...tree/<course>/<assignment>/feedback/<timestamp>` in it.

perllaghu:
> Thanks @perllaghu for your feedback. From my side the `view feedback` is working, but maybe the feature could be handled better. As there can be several feedbacks, nbextension opens a `tree file` tab. In this extension, instead of opening a new tab, it simply opens the `file browser` (left panel) in the correct directory. If you change the directory in the `file browser`, the link should open `...tree/<course>/<assignment>/feedback/<timestamp>` in it.

OK..... that's not happening in the Docker Images I have running in the K8 cluster.

Let me try in a basic image - there may be something in my alternative exchange code that's throwing it..

brichet:
> > Thanks @perllaghu for your feedback. From my side the `view feedback` is working, but maybe the feature could be handled better. As there can be several feedbacks, nbextension opens a `tree file` tab. In this extension, instead of opening a new tab, it simply opens the `file browser` (left panel) in the correct directory. If you change the directory in the `file browser`, the link should open `...tree/<course>/<assignment>/feedback/<timestamp>` in it.
> 
> OK..... that's not happening in the Docker Images I have running in the K8 cluster.
> 
> Let me try in a basic image - there may be something in my alternative exchange code that's throwing it..
> 
> edit... Yep, works in a basic Docker image.

Do you think this is something to fix ? How can we reproduce it ?

perllaghu:
> > OK..... that's not happening in the Docker Images I have running in the K8 cluster.
> > Let me try in a basic image - there may be something in my alternative exchange code that's throwing it..
> > edit... Yep, works in a basic Docker image.
> 
> Do you think this is something to fix ? How can we reproduce it ?

I've not got any further over the weekend.... however I think it's more likely to be a local problem that a core one - I'll have some pythonic library installed that's conflicting.

perllaghu:
In a separate thought-stream: did you consider adding the `formgrader` and/or `assignment list` commands to the Jupyterlab Interface?

I _think_ these are correct - it's part of the typescript in `src/index.ts`:

```
  // Add the command to the launcher
  if (launcher) {
    launcher.add({
      command,
      category: 'Other',
      rank: 1
    });
  }
```

and/or

```
  // Add the command to the menu
  if (menu) {
    menu.fileMenu.newMenu.addGroup([{ command }], 30);
  }
 ```
 
 (I was playing with something different over at https://github.com/edina/noteable-htmlProject )

**Comment by perllaghu** [comment]
> > > OK..... that's not happening in the Docker Images I have running in the K8 cluster.
> > > Let me try in a basic image - there may be something in my alternative exchange code that's throwing it..
> > > edit... Yep, works in a basic Docker image.
> > 
> > 
> > Do you think this is something to fix ? How can we reproduce it ?
> 
> I've not got any further over the weekend.... however I think it's more likely to be a local problem that a core one - I'll have some pythonic library installed that's conflicting.

Basic image works in the cluster.... so there's something added to my default image that breaks it.... and that [mostly] makes it my problem :)

**Comment by brichet** [comment]
> In a separate thought-stream: did you consider adding the `formgrader` and/or `assignment list` commands to the Jupyterlab Interface?

I added menu entries for `formgrader`, `assignment list` and `course list`.

File: nbgrader/docs/source/user_guide/managing_assignment_files_jlab.ipynb [hunk 1]
```diff
@@ -0,0 +1,911 @@
+{
```
jhamrick, line 1: Same comment here as above about duplication in the documentation.

tasks.py
```diff
@@ -115,7 +124,11 @@ def install(args):
         cmd = 'pip install .[docs,tests]'
     else:
         cmd = 'pip install -e .[tests]'
-    run(cmd)
+
+    env = os.environ.copy()
+    if args.group != 'labextensions':
```
jhamrick, line 129: Is this logic correct for when the group is 'all'?

jhamrick (changes requested):
This is looking really nice, thanks for the work on it! I found a few issues in my testing, described below. I tested in a clean mamba env using a fresh course directory created by `nbgrader quickstart`. I didn't really look at the code much, just a skim through.

- In the assignment list, the formatting causes the status messages to be clipped on the left side. This is partially aesthetic, but if you get the error that the exchange is not setup correctly, it hides the "h" from "http" in the the provided documentation url which will definitely confuse some people. Can the formatting be adjusted to fix this so there's no clipping?

image


- The "Create Assignment" extension doesn't seem to work for me. It appears on the right hand side as in the screenshots, but when I click on it nothing happens.

- When I run "Validate" on the instructor notebook through lab after having tried to activate the "Create Assignment" extension, I get the following error: "The nbgrader ID "null" has been used for more than one cell. Please make sure all grade cells have unique ids." This does not occur when running validate from the traditional notebook interface. It also does not occur if I haven't tried to use the "Create Assignment" extension.

image


- Mathjax does not seem to render when viewing generated feedback in a tab in lab, but it does render when viewing the same file directly in the browser. Example in lab:

image

Same file in browser:

image

- Validation output looks different depending on whether you run it from the assignment list extension or in the notebook itself (and both of these look different from how it appears when using the classic notebook, which shows identical output regardless of whether it's generated from the assignment list extension or from the notebook). When running validation from the notebook in lab, the errors are displayed with a pink background and are not scrollable:

image
 
When running validation from the assignment list extension in lab, the errors do not have a pink background and are displayed in scrollable boxes:

image

Classic notebook:

image

- When building the docs, I get a bunch of warnings that should probably be addressed:

```
/Users/jhamrick/project/nbgrader/nbgrader/docs/source/configuration/jupyterhub_config.rst:2: WARNING: Duplicate explicit target name: "jupyterhub documentation".
/Users/jhamrick/project/nbgrader/nbgrader/docs/source/user_guide/creating_and_grading_assignments_nb.rst:124: WARNING: duplicate label manually-graded-cells, other instance in /Users/jhamrick/project/nbgrader/nbgrader/docs/source/user_guide/creating_and_grading_assignments_jlab.rst
/Users/jhamrick/project/nbgrader/nbgrader/docs/source/user_guide/creating_and_grading_assignments_nb.rst:148: WARNING: duplicate label manually-graded-task-cells, other instance in /Users/jhamrick/project/nbgrader/nbgrader/docs/source/user_guide/creating_and_grading_assignments_jlab.rst
/Users/jhamrick/project/nbgrader/nbgrader/docs/source/user_guide/creating_and_grading_assignments_nb.rst:179: WARNING: duplicate label manually-graded-task-cell-mark-scheme, other instance in /Users/jhamrick/project/nbgrader/nbgrader/docs/source/user_guide/creating_and_grading_assignments_jlab.rst
/Users/jhamrick/project/nbgrader/nbgrader/docs/source/user_guide/creating_and_grading_assignments_nb.rst:191: WARNING: duplicate label autograded-answer-cells, other instance in /Users/jhamrick/project/nbgrader/nbgrader/docs/source/user_guide/creating_and_grading_assignments_jlab.rst
/Users/jhamrick/project/nbgrader/nbgrader/docs/source/user_guide/creating_and_grading_assignments_nb.rst:240: WARNING: duplicate label autograder-tests-cell-hidden-tests, other instance in /Users/jhamrick/project/nbgrader/nbgrader/docs/source/user_guide/creating_and_grading_assignments_jlab.rst
/Users/jhamrick/project/nbgrader/nbgrader/docs/source/user_guide/creating_and_grading_assignments_nb.rst:267: WARNING: duplicate label read-only-cells, other instance in /Users/jhamrick/project/nbgrader/nbgrader/docs/source/user_guide/creating_and_grading_assignments_jlab.rst
/Users/jhamrick/project/nbgrader/nbgrader/docs/source/user_guide/creating_and_grading_assignments_nb.rst:342: WARNING: duplicate label assign-and-release-an-assignment, other instance in /Users/jhamrick/project/nbgrader/nbgrader/docs/source/user_guide/creating_and_grading_assignments_jlab.rst
/Users/jhamrick/project/nbgrader/nbgrader/docs/source/user_guide/creating_and_grading_assignments_nb.rst:494: WARNING: duplicate label autograde-assignments, other instance in /Users/jhamrick/project/nbgrader/nbgrader/docs/source/user_guide/creating_and_grading_assignments_jlab.rst
/Users/jhamrick/project/nbgrader/nbgrader/docs/source/user_guide/managing_assignment_files_nb.rst:145: WARNING: duplicate label fetching-assignments, other instance in /Users/jhamrick/project/nbgrader/nbgrader/docs/source/user_guide/managing_assignment_files_jlab.rst
/Users/jhamrick/project/nbgrader/nbgrader/docs/source/configuration/jupyterhub_config.rst:7: WARNING: unknown document: /user_guide/creating_and_grading_assignments
/Users/jhamrick/project/nbgrader/nbgrader/docs/source/configuration/jupyterhub_config.rst:10: WARNING: unknown document: /user_guide/managing_assignment_files
/Users/jhamrick/project/nbgrader/nbgrader/docs/source/configuration/jupyterhub_config.rst:23: WARNING: unknown document: /user_guide/managing_assignment_files
/Users/jhamrick/project/nbgrader/nbgrader/docs/source/configuration/jupyterhub_config.rst:23: WARNING: unknown document: /user_guide/creating_and_grading_assignments
/Users/jhamrick/project/nbgrader/nbgrader/docs/source/configuration/student_version.rst:7: WARNING: unknown document: /user_guide/creating_and_grading_assignments
/Users/jhamrick/project/nbgrader/nbgrader/docs/source/user_guide/advanced.rst:25: WARNING: unknown document: managing_assignment_files
/Users/jhamrick/project/nbgrader/nbgrader/docs/source/user_guide/creating_and_grading_assignments_jlab.rst:448: WARNING: unknown document: managing_assignment_files
/Users/jhamrick/project/nbgrader/nbgrader/docs/source/user_guide/creating_and_grading_assignments_jlab.rst:462: WARNING: unknown document: managing_assignment_files
/Users/jhamrick/project/nbgrader/nbgrader/docs/source/user_guide/creating_and_grading_assignments_jlab.rst:464: WARNING: unknown document: managing_assignment_files
/Users/jhamrick/project/nbgrader/nbgrader/docs/source/user_guide/creating_and_grading_assignments_nb.rst:449: WARNING: unknown document: managing_assignment_files
/Users/jhamrick/project/nbgrader/nbgrader/docs/source/user_guide/creating_and_grading_assignments_nb.rst:463: WARNING: unknown document: managing_assignment_files
/Users/jhamrick/project/nbgrader/nbgrader/docs/source/user_guide/creating_and_grading_assignments_nb.rst:465: WARNING: unknown document: managing_assignment_files
/Users/jhamrick/project/nbgrader/nbgrader/docs/source/user_guide/faq.rst:75: WARNING: unknown document: managing_assignment_files
/Users/jhamrick/project/nbgrader/nbgrader/docs/source/user_guide/managing_assignment_files_jlab.rst:65: WARNING: unknown document: creating_and_grading_assignments
/Users/jhamrick/project/nbgrader/nbgrader/docs/source/user_guide/managing_assignment_files_jlab.rst:465: WARNING: unknown document: creating_and_grading_assignments
/Users/jhamrick/project/nbgrader/nbgrader/docs/source/user_guide/managing_assignment_files_nb.rst:65: WARNING: unknown document: creating_and_grading_assignments
/Users/jhamrick/project/nbgrader/nbgrader/docs/source/user_guide/managing_assignment_files_nb.rst:457: WARNING: unknown document: creating_and_grading_assignments
```

jhamrick:
One more comment too, after installing and building the docs `git status` looks like this:

```
$ git status
On branch jupyterlab_extensions
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   yarn.lock

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	nbgrader/docs/source/user_guide/creating_and_grading_assignments_jlab.rst
	nbgrader/docs/source/user_guide/creating_and_grading_assignments_nb.rst
	nbgrader/docs/source/user_guide/managing_assignment_files_jlab.rst
	nbgrader/docs/source/user_guide/managing_assignment_files_nb.rst
	nbgrader/labextension/
	tsconfig.tsbuildinfo
```

Can the untracked files be added to `.gitignore`? And perhaps `yarn.lock` shouldn't be committed if it changes when doing a development install?

nbgrader/docs/source/user_guide/creating_and_grading_assignments_jlab.ipynb
```diff
@@ -0,0 +1,1149 @@
+{
```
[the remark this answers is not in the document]
brichet: I agree.
As this branch is not supposed to keep compatibility with Classic Notebook, maybe we should drop the documentation on this for the next release ?

brichet:
Thanks @jhamrick for the feedback. I'll take a look at it as soon as possible.

nbgrader/docs/source/contributor_guide/installation_developer.rst
```diff
@@ -29,16 +35,17 @@ whenever you update the nbgrader repository.
 The serverextension is a Python module inside nbgrader, so only an
 enable step is needed.
 To install and enable all the frontend nbextensions (*assignment list*,
-*create assignment*, and *formgrader*) along with the server extensions
-(*assignment list* and *formgrader*) run::
+*create assignment*, and *formgrader*) and the frontend labextensions
+along with the server extensions (*assignment list* and *formgrader*) run::
 
+    jupyter labextension develop --overwrite .
```
jhamrick, line 41: This command will fail unless `pip install` has been run first, as I guess the pip command builds the lab extension. Perhaps it's worth stating this explicitly, to avoid confusion?

brichet:
> * In the assignment list, the formatting causes the status messages to be clipped on the left side. This is partially aesthetic, but if you get the error that the exchange is not setup correctly, it hides the "h" from "http" in the the provided documentation url which will definitely confuse some people. Can the formatting be adjusted to fix this so there's no clipping?

Fixed in https://github.com/jupyter/nbgrader/pull/1588/commits/47fb9b9a158e3ca90af2a615f3d73789c006348a

brichet:
> * The "Create Assignment" extension doesn't seem to work for me. It appears on the right hand side as in the screenshots, but when I click on it nothing happens.

Fixed in https://github.com/jupyter/nbgrader/pull/1588/commits/df462bd75b746a0ebb814cb56f86c1598fa030c8
The extension worked but the accordion did not open properly (width of 0px).

brichet:
> * Validation output looks different depending on whether you run it from the assignment list extension or in the notebook itself (and both of these look different from how it appears when using the classic notebook, which shows identical output regardless of whether it's generated from the assignment list extension or from the notebook). When running validation from the notebook in lab, the errors are displayed with a pink background and are not scrollable:

Fixed in https://github.com/jupyter/nbgrader/pull/1588/commits/5c52bd16d6a06fb2e9b4fd31dc411965ac8e0a23 by using the same functions to validate Notebook and display modal, from Notebook itself and from `Assignment list`

brichet:
> * Mathjax does not seem to render when viewing generated feedback in a tab in lab, but it does render when viewing the same file directly in the browser. Example in lab:

The formgrader frontend extension is essentially an iframe including the output of the formgrader server extension. It seems that the Mathjax problem is a security issue. There is a "Trust HTML" button in the upper left corner of the formgrader tab, which renders correctly. I have not yet found a way to trust it automatically.

brichet:
> * When I run "Validate" on the instructor notebook through lab after having tried to activate the "Create Assignment" extension, I get the following error: "The nbgrader ID "null" has been used for more than one cell. Please make sure all grade cells have unique ids." This does not occur when running validate from the traditional notebook interface. It also does not occur if I haven't tried to use the "Create Assignment" extension.

I can't reproduce that behaviour.

tasks.py
brichet: Right, fixed in https://github.com/jupyter/nbgrader/pull/1588/commits/ca01cd66c7664fcf254cccbd3c67e5cdac3f466d

nbgrader/docs/source/contributor_guide/installation_developer.rst
brichet: fixed in https://github.com/jupyter/nbgrader/pull/1588/commits/ca01cd66c7664fcf254cccbd3c67e5cdac3f466d

brichet:
> Can the untracked files be added to `.gitignore`? And perhaps `yarn.lock` shouldn't be committed if it changes when doing a development install?

fixed in https://github.com/jupyter/nbgrader/pull/1588/commits/ca01cd66c7664fcf254cccbd3c67e5cdac3f466d

nbgrader/docs/source/user_guide/creating_and_grading_assignments_jlab.ipynb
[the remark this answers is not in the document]
jhamrick: Sounds ok to me.

jhamrick:
Great, now I can test "Create Assignment"---thanks! It looks good, except that I am unable to scroll down if there are many cells in the notebook. e.g in the screenshot below, you can see that the cell on the left can be scrolled into view, but its corresponding area in the extension is cut off, and I can't seem to make it scroll down:

image

I also am still seeing `yarn.lock` in the repo. As it gets regenerated and changed upon running `pip install`, I think perhaps it should just not be committed and then added to `.gitignore`? Or is there a reason it needs to be committed?

Regarding the error message about the null cell ids, what notebook are you testing with? I get it with both of the notebooks included from `nbgrader quickstart`.

brichet:
> Menu items should not be added unless the extension is enabled... more specifically _removed_ if the extension is disabled.

@perllaghu currently the labextensions are installed all together, there is no way to install them separately.
I did some tests to disable some labextension using the settings (https://github.com/jupyterlab/extension-examples/tree/master/settings). This works fine (for *formgrader*, *assignment list* and *course list*), but activation is still available from Jupyter Lab `advanced settings editor`.
Do you think that we need to find a way to permanently disable some extensions (on be able to install them separately) ?

perllaghu:
> Do you think that we need to find a way to permanently disable some extensions (or be able to install them separately) ?

We currently _enable_ extensions separately... and I know _we_ only enable formgrader for instructors [though I'd be interested what others do, and if we're in a minority in this behaviour]

I would encourage breaking things into separate extensions, installed separately.


Oh, and remember the 80:20 rule: "The last 20% of the work will take 80% of the time"
Major kudos for getting things actually _running_ and to a discussable PR!

perllaghu:
A variation to @jhamrick 's discovery above: the Assignment List extension doesn't scroll when the content is longer than the browser window
nbgrader_iframe-bound

jtpio:
Mentioning that as another option: it is also possible to disable individual extensions (and plugins) with the `jupyter labextension disable` command, or by setting `disabledExtensions` in the page config: https://jupyterlab.readthedocs.io/en/stable/user/extensions.html?#enabling-and-disabling-extensions

perllaghu:
Question - is jupyterlab 3.4.2 _actually_ needed?

It was only released a few days ago, and seems quite a stringent dependency

Oh - and the extension is installed & working in 'lab 3.3.4 [for me]

```
"nbgrader@0.8.0-dev" is not compatible with the current JupyterLab
Conflicting Dependencies:
JupyterLab           Extension      Package
>=3.3.4 <3.4.0       >=3.4.2 <4.0.0 @jupyterlab/launcher
```

brichet:
> Great, now I can test "Create Assignment"---thanks! It looks good, except that I am unable to scroll down if there are many cells in the notebook. e.g in the screenshot below, you can see that the cell on the left can be scrolled into view, but its corresponding area in the extension is cut off, and I can't seem to make it scroll down:

My bad, I changed the container widget to have a minimum width when opening (from BoxPanel to Panel), but the new widget was not scrollable. I undid the change and applied a minimum width to BoxPanel in https://github.com/jupyter/nbgrader/pull/1588/commits/08a702f3249b1c0f5936f1406b2a270197dedca6.
Minimum width now works on my side.

brichet:
> I also am still seeing `yarn.lock` in the repo. As it gets regenerated and changed upon running `pip install`, I think perhaps it should just not be committed and then added to `.gitignore`? Or is there a reason it needs to be committed?

According to https://classic.yarnpkg.com/en/docs/migrating-from-npm, yarn.lock file is supposed to ensure dependencies : 
> When you run either yarn or yarn add <package>, Yarn will generate a yarn.lock file within the root directory of your package. You don’t need to read or understand this file - just check it into source control. When other people start using Yarn instead of npm, the yarn.lock file will ensure that they get precisely the same dependencies as you have.

The labextensions built using `jlpm`, a jupyterlab's bundled version of yarn. As I understand it, the installation should not generate a new yarn.lock file each time. So far I've removed it from committed files, but I wonder if that's why the `check release` workflows never end.

brichet:
> Regarding the error message about the null cell ids, what notebook are you testing with? I get it with both of the notebooks included from `nbgrader quickstart`.

Right, all the cells were checked instead of only those with nbgrader metadata. Fixed in https://github.com/jupyter/nbgrader/pull/1588/commits/7d9373e88463a94becb1036df85e3da55b187225

jhamrick:
Awesome, thanks for the changes! 

> Oh, and remember the 80:20 rule: "The last 20% of the work will take 80% of the time"
> Major kudos for getting things actually running and to a discussable PR!

Huge +1 to this, thanks a lot for working on this!

- Another small thing I noticed in "Create Assignment" -- the total points box is horizontally very small, and can cause the total number of points displayed to be clipped. e.g. here you can see the 3 being clipped, and if the value is 10 you don't see the 0 at all:

image

- I occasionally seem to run into a bug where when I click on links in the formgrader that should update the folder shown in the explorer, nothing happens (maybe related to what @perllaghu was experiencing?). I see this error in the javascript console:

image

- I am seeing the same MathJax issue as before with the manual grading interface too, though I don't see a way to trust the HTML:

image

- Related to the point above, perhaps we can say in the documentation something like, instructors will need to instruct their students to click the "trust HTML" button for it to render properly?

- Regarding the yarn file, hmm that's strange that it's meant to be committed but keeps getting regnerated. I don't really know anything about npm so I'm afraid I don't really know what the best course of action is, but it does seem suboptimal to me that it gets changed on install. It doesn't seem like something to me that should be getting changed with every PR?

- I'm not sure I'll have time, but if I do I'll try to test this out with a JupyterHub install and multiple courses and such too.

brichet:
> Mentioning that as another option: it is also possible to disable individual extensions (and plugins) with the `jupyter labextension disable` command, or by setting `disabledExtensions` in the page config: https://jupyterlab.readthedocs.io/en/stable/user/extensions.html?#enabling-and-disabling-extensions

Thanks @jtpio , it works great.

@perllaghu I updated the documentation in https://github.com/jupyter/nbgrader/pull/1588/commits/9226994e900eaf825d28fcc8fa27d9842679c9ce, but basically the labextensions are installed all together but can be disabled individually
```
jupyter labextension disable nbgrader/formgrader
jupyter labextension disable nbgrader/assignment-list
jupyter labextension disable nbgrader/course-list
jupyter labextension disable nbgrader/create-assignment
jupyter labextension disable nbgrader/validate-assignment
```

jtpio:
`yarn.lock` is indeed supposed to be checked in the repo. It's pretty much a snapshot of the dependencies at a given point in time.

If the `check-release` check fails on CI, it's probably because the `yarn` command resulted in changes to the `yarn.lock`, making the repository not clean anymore since it has local changes.

Probably running `yarn` locally will update `yarn.lock` which will have to be committed and pushed. Maybe a `rm -r node_modules` first can also help.

SylvainCorlay:
As we were looking at the PR with Nicolas Brichet @brichet, we realized that bootstrap's CSS was on the page and polluting the general CSS namespace, causing some breakage in the rest of the JupyterLab UI. (Bootstrap is not a good citizen of the DOM).

This is not too much of an issue for formgrader, which is displayed in an iframe, but it is more problematic for the assignment list. One way to go would be to refactor the assignment list page to not rely on bootstrap for layout.

brichet:
> A variation to @jhamrick 's discovery above: the Assignment List extension doesn't scroll when the content is longer than the browser window

@perllaghu fixed in https://github.com/jupyter/nbgrader/pull/1588/commits/b17c4f597a1dd794a1f7826e14c0f601fc2adcc2 (thanks @SylvainCorlay)



> As we were looking at the PR with Nicolas Brichet @brichet, we realized that bootstrap's CSS was on the page and polluting the general CSS namespace, causing some breakage in the rest of the JupyterLab UI. (Bootstrap is not a good citizen of the DOM).

@SylvainCorlay fixed in https://github.com/jupyter/nbgrader/pull/1588/commits/b17c4f597a1dd794a1f7826e14c0f601fc2adcc2

brichet:
> * I am seeing the same MathJax issue as before with the manual grading interface too, though I don't see a way to trust the HTML:

This issue also exists in the 0.7.x branch. It may be due to the nbconvert upgrade. I will try to fix it in another PR.

jhamrick:
Ok, running locally I think everything is looking pretty good to me!

However, I believe there are some issues when running with JupyterHub, particularly in the context of running the formgrader as a service and allowing multiple users to access it. Specifically, if the formgrader is being run as a service (as in the multiple classes and multiple graders [demos](https://github.com/jupyter/nbgrader/tree/main/demos)) the course list extension should provide a link to the formgrader running as a service, *not* a formgrader running on the instructor's account. However, it seems that with the lab extension if I click on the course name I get an instructor-specific formgrader rather than the service. Additionally, if there are multiple courses, it seems that the extension will open the same formgrader for both of them rather than multiple tabs. I think there's a bit of a deeper problem here, which is how to run the formgrader as a service now---previously we were doing this by just running the singleuser server and providing access to it, which worked ok since we would just open the formgrader as a new browser tab. But since everything now lives within lab, I'm not sure that would work?

I also am seeing a similar behavior that @perllaghu saw where the extensions are not being disabled properly for students. I am running:

```
jupyter labextension develop --overwrite .
jupyter labextension disable --level=sys_prefix nbgrader
```

and then only enable formgrader for the instructor accounts, but it still appears in the command palette for students. Similarly for the course list and create assignment.

To reproduce these issues, you can run the demos linked above---you'll need to patch in #1597 (until it's merged) as well as the following patch: [jupyterlab-patch.txt](https://github.com/jupyter/nbgrader/files/8890144/jupyterlab-patch.txt)

nbgrader/docs/source/user_guide/creating_and_grading_assignments_jlab.ipynb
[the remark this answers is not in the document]
brichet: The documentation has been updated in https://github.com/jupyter/nbgrader/pull/1588/commits/db5eb6a23a53dbf0b6f0f7c379b09d374297fae1

nbgrader/docs/source/user_guide/managing_assignment_files_jlab.ipynb
brichet: The documentation has been updated in https://github.com/jupyter/nbgrader/pull/1588/commits/db5eb6a23a53dbf0b6f0f7c379b09d374297fae1

brichet:
> I think there's a bit of a deeper problem here, which is how to run the formgrader as a service now---previously we were doing this by just running the singleuser server and providing access to it, which worked ok since we would just open the formgrader as a new browser tab. But since everything now lives within lab, I'm not sure that would work?

Opening the correct formgrader in a new JupyterLab tab should be OK, this is the URL of the formgrader to link to in the `iframe`.
The main problem with using formgrader as a service is being able to change the root directory, as the formgrader server extension runs in its own root directory, which is not the same as the jupyterlab instance.
I wonder if the best way is to open a new browser tab with a new instance of jupyterlab (if possible). This means that running a `local formgrader` would work as now (all in jupyterlab), but when it's a `service formgrader` it would open a new jupyterlab instance in a new tab (with the correct root directory).

jhamrick:
Ah fantastic! Thanks for the patch, this looks great.

So I think the remaining two issues are:

1. Extensions being enabled, even after having been disabled.
2. The `yarn.lock` file getting regenerated when doing a development install. I think (2) is a problem as it will definitely cause merge conflicts and contributors will unknowingly try to commit it when making PRs with unrelated changes. But I also understand it's supposed to be committed into the repo. Is there any way we can avoid it getting changed on every install? Maybe we could have an explicit command to update it, rather than it getting changed automatically?

Once those issues are resolved I think I am happy to merge this and continue iterating on `main` as more people start to test it out!

jtpio:
> 2\. The `yarn.lock` file getting regenerated when doing a development install. I think (2) is a problem as it will definitely cause merge conflicts and contributors will unknowingly try to commit it when making PRs with unrelated changes. But I also understand it's supposed to be committed into the repo. Is there any way we can avoid it getting changed on every install? Maybe we could have an explicit command to update it, rather than it getting changed automatically?

Normally it should not changed, unless new dependencies are added to the `package.json`. Probably running `jlpm` locally one more time will update `yarn.lock`. The changes should then be committed and pushed to the branch.

brichet:
> 1. Extensions being enabled, even after having been disabled.

It's working on my side with other commands.
To disable extensions at *sys_prefix* level, I think the command `jupyter labextension disable --level=sys_prefix nbgrader` create confusion in that case (maybe due to the way the extensions are built). If you repeat commands for each extension individually it works fine.

To enable extensions at *user* level, it may be an issue on Jupyterlab, depending on what was the initial goal. I'll try to explain what I understood, but I may be wrong. 
If you disable an extension, it create a file `page_config.json` (at *user*, *sys_prefix* or *system* level). And the files will be read in that order, so if an extension is enabled or disabled explicitly  at *user* level, it will not look for another file at *sys_prefix* or *system* level.
But it also seems that enabling an extension at a level where it has not been explicitly disabled does nothing (no file is created).
In our case, we disable extensions at *sys_prefix* level, that's OK. Then we try to enable it at *user* level, which does nothing. So the extension remain disabled. A workaround is to disable it before enabling it at *user* level.

You can try applying this patch which works on my side :
```
diff --git a/demos/restart_demo.sh b/demos/restart_demo.sh
index bcc00420..abf76a43 100755
--- a/demos/restart_demo.sh
+++ b/demos/restart_demo.sh
@@ -58,7 +58,9 @@ install_nbgrader () {
     jupyter nbextension install --symlink --sys-prefix --py nbgrader --overwrite
     jupyter nbextension disable --sys-prefix --py nbgrader
     jupyter labextension develop --overwrite .
-    jupyter labextension disable --level=sys_prefix nbgrader
+    jupyter labextension disable --level=sys_prefix nbgrader/formgrader
+    jupyter labextension disable --level=sys_prefix nbgrader/course-list
+    jupyter labextension disable --level=sys_prefix nbgrader/create-assignment
     jupyter serverextension disable --sys-prefix --py nbgrader
 
     # Everybody gets the validate extension, however.
diff --git a/demos/utils.sh b/demos/utils.sh
index 9d3fe3ba..d3055659 100644
--- a/demos/utils.sh
+++ b/demos/utils.sh
@@ -70,7 +70,8 @@ enable_create_assignment () {
     local runas="sudo -u ${USER}"
 
     ${runas} jupyter nbextension enable --user create_assignment/main
-    # ${runas} jupyter labextension enable --level=user nbgrader/create-assignment
+    ${runas} jupyter labextension disable --level=user nbgrader/create-assignment
+    ${runas} jupyter labextension enable --level=user nbgrader/create-assignment
 }
 
 enable_formgrader () {
@@ -79,6 +80,7 @@ enable_formgrader () {
     local runas="sudo -u ${USER}"
 
     ${runas} jupyter nbextension enable --user formgrader/main --section=tree
+    ${runas} jupyter labextension disable --level=user nbgrader/formgrader
     ${runas} jupyter labextension enable --level=user nbgrader/formgrader
     ${runas} jupyter serverextension enable --user nbgrader.server_extensions.formgrader
 }
@@ -89,6 +91,7 @@ enable_assignment_list () {
     local runas="sudo -u ${USER}"
 
     ${runas} jupyter nbextension enable --user assignment_list/main --section=tree
+    ${runas} jupyter labextension disable --level=user nbgrader/assignment-list
     ${runas} jupyter labextension enable --level=user nbgrader/assignment-list
     ${runas} jupyter serverextension enable --user nbgrader.server_extensions.assignment_list
 }
@@ -99,6 +102,7 @@ enable_course_list () {
     local runas="sudo -u ${USER}"
 
     ${runas} jupyter nbextension enable --user course_list/main --section=tree
+    ${runas} jupyter labextension disable --level=user nbgrader/course-list
     ${runas} jupyter labextension enable --level=user nbgrader/course-list
     ${runas} jupyter serverextension enable --user nbgrader.server_extensions.course_list
 }
```

nbgrader/docs/source/user_guide/installation.rst
```diff
@@ -14,70 +14,69 @@ Or, if you use `Anaconda <https://www.anaconda.com/download>`__::
...
+    jupyter server extension disable nbgrader.server_extensions.assignment_list
+    jupyter server extension disable nbgrader.server_extensions.course_list
+    jupyter server extension disable nbgrader.server_extensions.validate_assignment
 
-To work properly, the **assignment list**, **formgrader**, and **validate**
-extensions require both the nbextension and serverextension. The **create
-assignment** extension only has an nbextension part.
+To disable all the labextensions run::
+
+    jupyter labextension disable nbgrader
```
jhamrick, line 36: Perhaps remove this, as it doesn't seem to work as one would expect, and just tell people they have to enable/disable the extensions individually.

brichet: Done in https://github.com/jupyter/nbgrader/pull/1588/commits/5519c65b14e49e0712f690357e2bc7de597f26bd

jhamrick (approved):
Looks great. Thanks so much!!

brichet:
Thanks @jhamrick @perllaghu @jtpio and @SylvainCorlay for testing and code review

orboan:
If anyone could take a look at this issue:

[1645](https://github.com/jupyter/nbgrader/issues/1645)

Thank you

PR #1597, opened 2022-06-10
Debug demos

### LINKED ISSUE #1594: JupyterHub demos are broken
With JupyterHub 2.0, the [demos](https://github.com/jupyter/nbgrader/tree/main/demos) that use services (i.e. `demo_one_class_multiple_graders` and `demo_multiple_classes`) no longer work in that the instructors do not have permission to access the formgrader. I thought it would be a matter of simply adding something like the following (for `demo_one_class_multiple_graders`) to `jupyterhub_config.py`:

```
c.JupyterHub.load_roles = [
    {
        'name': 'formgrader-course101',
        'groups': ['formgrader-course101'],
        'scopes': [
            'access:services!service=course101',
        ]

    }
]
```

but this doesn't seem to work; I get an error message that `[D 2022-06-08 19:38:28.329 SingleUserNotebookApp auth:995] Hub user instructor1 needs scope(s) {'access:services', 'access:services!service=course101'}, has scope(s) ['access:servers!user=instructor1', 'read:users:activity!user=instructor1', 'users:activity!user=instructor1']`. This is true also if I use `'users': ['instructor1']` rather than `'groups'`. I am not sure why the scopes are not being correctly set.

This is related to #1533, which documents that there is an incorrect scope configuration for the formgrader service itself too.

@perllaghu do you have any insight on this from your deployment of nbgrader?

minrk:
includes a Dockerfile which, when I run it, produces working examples

maybe closes #1594 though it would still be be good to understand what's not working for @jhamrick there.

### DIFF
--- a/demos/restart_demo.sh
+++ b/demos/restart_demo.sh
@@ -26,7 +26,7 @@ install_dependencies () {
     apt install -y npm
     npm install -g configurable-http-proxy
     apt install -y python3-pip
-    pip3 install -U notebook
+    pip3 install -U notebook nbclassic
     pip3 install -U jupyterhub
 }
 
--- a/demos/restart_demo.sh
+++ b/demos/restart_demo.sh
@@ -1,9 +1,9 @@
 #!/usr/bin/env bash
 
-set -e
+set -ex
 
 # Configuration variables.
-root="/root"
+root="$(realpath `dirname $0`)"
 srv_root="/srv/nbgrader"
 nbgrader_root="/srv/nbgrader/nbgrader"
 jupyterhub_root="/srv/nbgrader/jupyterhub"
@@ -26,7 +26,7 @@ install_dependencies () {
     apt install -y npm
     npm install -g configurable-http-proxy
     apt install -y python3-pip
-    pip3 install -U jupyter
+    pip3 install -U notebook
     pip3 install -U jupyterhub
 }
 
@@ -45,11 +45,10 @@ install_nbgrader () {
 
     # Update git repository.
     cd "${nbgrader_root}"
-    git checkout master
     git pull
 
     # Install requirements and nbgrader.
-    pip3 install -U -r requirements.txt -e .
+    pip3 install -U .
 
     # Install global extensions, and disable them globally. We will re-enable
     # specific ones for different user accounts in each demo.
@@ -71,8 +70,11 @@ install_nbgrader () {
 
 restart_demo () {
     local demo="${1}"
+    test ! -z "$demo"
 
     install_dependencies
+    setup_directory "/etc/jupyter" ugo+r
+
     setup_directory "${srv_root}" ugo+r
     install_nbgrader "${nbgrader_root}" "${exchange_root}"
 
--- a/demos/Dockerfile
+++ b/demos/Dockerfile
@@ -1,0 +1,20 @@
+# Dockerfile for running / testing demos
+# build: `docker build -t nbgrader-demo .`
+# build and run: `docker build -t nbgrader-demo . && docker run --rm -it -p 8000:8000 nbgrader-demo`
+# run a demo without rebuilding by mounting the demo dir as a volume (e.g. if you are editing the demos):
+# docker run --rm -it -p 8000:8000 -v $PWD:/root nbgrader-demo /root/restart_demo.sh demo_multiple_classes
+
+FROM ubuntu:22.04
+ENV DEBIAN_FRONTEND=noninteractive
+RUN apt-get update \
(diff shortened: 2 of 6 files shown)


jhamrick:
Everything looks good from the instructor end for me now---woohoo!

As you mentioned in https://github.com/jupyter/nbgrader/issues/1594 though it seems that in the "Assignment List" extension, the students are not able to see the list of courses and corresponding assignments. I *think* it's a similar issue---the students' servers need to be able to list the services (though not access them). I think I have enough to try fiddling with the config file to fix this now, let me give it a go :) 

Thank you so much for working on this!

jhamrick:
Ok, fixed the Assignment List for students. Here is a patch with my changes if you'd like to incorporate them:

```
diff --git a/demos/demo_multiple_classes/jupyterhub_config.py b/demos/demo_multiple_classes/jupyterhub_config.py
index ddfdecdd..e7492e6f 100644
--- a/demos/demo_multiple_classes/jupyterhub_config.py
+++ b/demos/demo_multiple_classes/jupyterhub_config.py
@@ -9,7 +9,10 @@ c.Authenticator.allowed_users = [
     'grader-course123',
 ]
 
-# instructor1 and instructor2 have access to different shared servers:
+# instructor1 and instructor2 have access to different shared servers.
+# Note that groups providing access to the formgrader *must* start with
+# 'formgrade-', and groups providing access to course materials *must*
+# start with 'nbgrader-' in order for nbgrader to work correctly.
 c.JupyterHub.load_groups = {
     'instructors': [
         'instructor1',
@@ -23,8 +26,14 @@ c.JupyterHub.load_groups = {
         'instructor2',
         'grader-course123',
     ],
-    'nbgrader-course101': [],
-    'nbgrader-course123': [],
+    'nbgrader-course101': [
+        'instructor1',
+        'student1',
+    ],
+    'nbgrader-course123': [
+        'instructor2',
+        'student1',
+    ],
 }
 
 c.JupyterHub.load_roles = roles = [
@@ -54,20 +63,30 @@ c.JupyterHub.load_roles = roles = [
         ],
     },
 ]
-# add grader roles
 for course in ['course101', 'course123']:
+    # access to formgrader
     roles.append(
         {
             'name': f'formgrade-{course}',
             'groups': [f'formgrade-{course}'],
             'scopes': [
                 f'access:services!service={course}',
+            ],
+        }
+    )
+    # access to course materials
+    roles.append(
+        {
+            'name': f'nbgrader-{course}',
+            'groups': [f'nbgrader-{course}'],
+            'scopes': [
                 # access to the services API to discover the service(s)
                 'list:services',
                 f'read:services!service={course}',
             ],
         }
     )
+
 # Start the notebook server as a service. The port can be whatever you want
 # and the group has to match the name of the group defined above.
 c.JupyterHub.services = [
diff --git a/demos/demo_one_class_multiple_graders/jupyterhub_config.py b/demos/demo_one_class_multiple_graders/jupyterhub_config.py
index 49f24cbb..678f0583 100644
--- a/demos/demo_one_class_multiple_graders/jupyterhub_config.py
+++ b/demos/demo_one_class_multiple_graders/jupyterhub_config.py
@@ -7,13 +7,22 @@ c.Authenticator.allowed_users = [
     'student1',
 ]
 
-# instructor1 and instructor2 have access to a shared server:
+# instructor1 and instructor2 have access to a shared server.
+# This group *must* start with formgrade- for nbgrader to work correctly.
 c.JupyterHub.load_groups = {
     'formgrade-course101': [
         'instructor1',
         'instructor2',
     ]
 }
+
+# Note: we don't need to add students to a nbgrader-course101 group to give
+# them access to coure materials (as we do in demo_multiple_classes) because
+# we explicitly set the course id in the nbgrader_config.py. If we didn't
+# set this explicitly, we would also have to create a nbgrader-course101
+# group and give it the scopes of:
+# ['list:services', 'read:services!service=course101'].
+
 c.JupyterHub.load_roles = [
     {
         'name': 'formgrade-course101',
```

demos/restart_demo.sh
```diff
@@ -45,11 +45,10 @@ install_nbgrader () {
 
     # Update git repository.
     cd "${nbgrader_root}"
-    git checkout master
     git pull
 
     # Install requirements and nbgrader.
-    pip3 install -U -r requirements.txt -e .
+    pip3 install -U .
```
jhamrick, line 51: I'm curious why `-e` doesn't work... is there any way we can keep the old behavior? It's useful for being able to debug/develop the demos.

jhamrick:
Superseded by #1601

PR #1603, opened 2022-06-16
Use jupyter css variable in labextension to manage colors with theme

brichet:
Change the hard coded CSS color by Jupyterlab CSS variables in labextension to allow changing theme (light and dark).
Only the formgrader extension remains in light style as it depends on bootstrap.

### DIFF
--- a/style/assignment_list.css
+++ b/style/assignment_list.css
@@ -4,7 +4,6 @@
 }
 
 #nbgrader-assignment-list .panel-group .panel .panel-heading {
-  background-color: #eee;
   padding-top: 4px;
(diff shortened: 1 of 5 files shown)


jhamrick:
Overall looks good, though I think the "lock" icon is the wrong color, as it is not visible in JupyterLab in either the white or dark theme (well you can almost see it in the dark theme).

image
image

Compare to the classic version:

image

brichet:
Thanks for feedback @jhamrick, it should be fixed now

jhamrick:
Thanks!

PR #1606, opened 2022-06-20
Create assignment panel only open if a Notebook is visible on main area

brichet:
This PR improves the behavior of the `create assignment` widget :
- If no Notebook is visible in main area, clicking on the button does not open the right panel (instead of opening an empty one)
- if a notebook is visible, it will be activated (can occur when the main area is split)
- if the selected tab in main area becomes a tab which is not a Notebook, the `create assignment` panel is closed

### DIFF
--- a/.github/workflows/test-labextensions.yml
+++ b/.github/workflows/test-labextensions.yml
@@ -57,7 +57,7 @@ jobs:
       - name: Install Playwright
         run: |
           npx playwright install
-      - name: Run pytest
+      - name: Run playwright tests
         run: |
           python tasks.py tests --group=labextensions
       - name: Upload Playwright Test report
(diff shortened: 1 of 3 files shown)


jtpio (approved):
Looks good thanks!

jtpio:
At some point we'll want to start looking into how the extension behaves in Notebook v7 application, which doesn't provide a `ILabShell`. 

But that can be tracked in a separate issue.

brichet:
> At some point we'll want to start looking into how the extension behaves in Notebook v7 application, which doesn't provide a `ILabShell`.
> 
> But that can be tracked in a separate issue.

Sure, but I think we'll do it in a separate issue. 
For now it's not working with Notebook v7, and doesn't seem to be fully compatible with JupyterLab 4 either.

jhamrick:
Nice, thanks!

PR #1610, opened 2022-06-23
Fix version bumping for pre-releases

jtpio:
Running the Draft Changelog on the `main` branch failed with the following error:

```
Error: Could not parse 0.8.0a0 as a valid version string
```

https://github.com/jtpio/jupyter_releaser/runs/7023597896?check_suite_focus=true

This looks related to bumping the version for pre-releases.

This change should fix it.

### DIFF
--- a/.github/workflows/check-release.yml
+++ b/.github/workflows/check-release.yml
@@ -19,7 +19,7 @@ jobs:
         uses: jupyterlab/maintainer-tools/.github/actions/base-setup@v1
       - name: Install Dependencies
         run: |
-          pip install -e .
+          pip install -e ".[dev]"
       - name: Check Release
         if: ${{ matrix.group == 'check_release' }}
         uses: jupyter-server/jupyter_releaser/.github/actions/check-release@v1
--- a/demos/restart_demo.sh
+++ b/demos/restart_demo.sh
@@ -48,7 +48,7 @@ install_nbgrader () {
     git pull
 
     # Install requirements and nbgrader.
-    pip3 install -e ".[docs,tests]"
+    pip3 install -e ".[dev,docs,tests]"
 
     # Install global extensions, and disable them globally. We will re-enable
     # specific ones for different user accounts in each demo.
--- a/nbgrader/_version.py
+++ b/nbgrader/_version.py
@@ -1,2 +1,2 @@
-version_info = (0, 8, 0, "dev")
-__version__ = '.'.join(map(str, version_info[:3])) + (("." + version_info[3]) if version_info[3] else "")
+version_info = (0, 8, 0, ".dev", "0")
+__version__ = ".".join(map(str, version_info[:3])) + "".join(version_info[3:])
--- a/nbgrader/docs/source/contributor_guide/installation_developer.rst
+++ b/nbgrader/docs/source/contributor_guide/installation_developer.rst
@@ -20,7 +20,7 @@ We recommand using `conda environment <https://docs.conda.io/en/latest/miniconda
     # activate the environment
     mamba activate nbgrader
 
-    pip install -e ".[docs,tests]"
+    pip install -e ".[dev,docs,tests]"
 
 Installing Jupyter labextensions
 --------------------------------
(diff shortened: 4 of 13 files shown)


jtpio:
This is what it looks like locally running `tbump --non-interactive --only-patch 0.8.0a0`:

image

jtpio:
Currently the version in the top-level `package.json` is set to `0.8.0`:

https://github.com/jupyter/nbgrader/blob/086a81d874a47acbcdbf9669e923c5f5cab8662d/package.json#L3

But for `0.8.0a0` it should be set to the corresponding `0.8.0-alpha.0`.

brichet:
> Currently the version in the top-level `package.json` is set to `0.8.0`:
> 
> https://github.com/jupyter/nbgrader/blob/086a81d874a47acbcdbf9669e923c5f5cab8662d/package.json#L3
> 
> But for `0.8.0a0` it should be set to the corresponding `0.8.0-alpha.0`.

I agree, do you think there is a way to do that using tbump ?

jtpio:
> I agree, do you think there is a way to do that using tbump ?

hmm not sure.

But an alternative could be to run a small script as a `after-bump-version` hook doing something like this:

```py
version.replace('a', '-alpha.').replace('b', '-beta.').replace('rc', '-rc.')
```

Similar to the cookiecutter but the other way around: https://github.com/jupyterlab/extension-cookiecutter-ts/blob/a706ffd9d5d2c2813c427d571caaab77014dae17/%7B%7Bcookiecutter.python_name%7D%7D/setup.py#L40-L45

And write the modified version back to `package.json`.

brichet:
> And write the modified version back to `package.json`.

It seems that the 2 hooks available in `tbumb ` are `before_commit` and `after_push`.
If we modify the `package.json` file, don't you think that `tbump` will complain with a `Current version string not found` error on the next release.
We should need a `before_bump` hook to reverse it.

jtpio:
Right I meant a Jupyter Releaser hook.

Just pushed https://github.com/jupyter/nbgrader/pull/1610/commits/e69c9b5128dcd348a954382a240ad426b0589cd1 to try something along these lines, will check what the check release CI job does.

brichet:
Nice @jtpio

In `tools/post-bump.py` l20 is that not `"-dev"` instead of `"dev"` ?

jtpio:
Thanks @brichet makes sense for consistency. Updated in https://github.com/jupyter/nbgrader/pull/1610/commits/4a606ea86bcc987c56b8f16ba98fca6fe26787d8

jtpio:
Looks like the `nbextensions` checks are now failing.

Restarting them to see if they are just flaky.

brichet:
This is related to `selenium.webdriver` :
`E           AttributeError: 'WebDriver' object has no attribute 'find_elements_by_css_selector'`
The current version of selenium is `4.3.0`. With the `4.2.0` the tests were passing.
As this branch will not support nbextension, we can just lock the version `selenium==4.2.0` for this PR, and open an issue on 0.7.x branch.

jtpio:
Ah indeed it looks like `selenium` `4.3.0` was released today:

image

Maybe we can fix it in a separate PR since it's likely to affect other PRs? Happy to rebase this one afterwards.

brichet:
> Maybe we can fix it by pinning on `4.2.0` in a separate PR since it's likely to affect other PRs? Happy to rebase this one afterwards.

Done in https://github.com/jupyter/nbgrader/pull/1611

jtpio:
@brichet this PR was rebased and CI is now passing, if you would like to have another look.

Happy to attempt a new release right after to see if that goes better that time.

brichet:
> @brichet this PR was rebased and CI is now passing, if you would like to have another look.
> 
> Happy to attempt a new release right after to see if that goes better that time.

Everything seems OK for me.

jtpio:
Alright then let's get it in and I'll start the pre-release process right after.

jtpio:
Looks like the "Draft Changelog" workflow went through this time: https://github.com/jupyter/nbgrader/pull/1613

PR #1611, opened 2022-06-24
Pins selenium version to 4.2

brichet:
Pins selenium version to `4.2` as the latest `4.3.0` breaks tests.


jtpio (approved):
Thanks for this!

jtpio:
Let's open an issue to track updating to the latest version?

PR #1616, opened 2022-06-28
Fix version error message

brichet:
Replace `nbextension` by `labextension` in the error message when the versions between the server and the extension mismatch.


jtpio (approved):
Thanks!

PR #1617, opened 2022-06-28
Move the jupyter_releaser hooks from package.json to pyproject.toml

brichet:
Brings together all `jupyter_releaser` hooks  in `pyproject.toml` file,  it seems that the jupyter_releaser do not run the hooks from 2 sources.

cc @jtpio for review

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ ignore = ["nbgrader/labextension/**", "yarn.lock"]
 
 [tool.jupyter-releaser.hooks]
 after-bump-version = ["python tools/post-bump.py"]
-before-build-npm = ["jlpm"]
+before-build-npm = ["python -m pip install 'jupyterlab<4'", "jlpm"]
 before-build-python = ["jlpm clean:all"]
 
 [tool.tbump.version]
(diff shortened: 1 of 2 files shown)


jtpio:
Thanks @brichet!

Indeed the releaser seems to read the config only from one source: https://github.com/jupyter-server/jupyter_releaser/blob/023124cbb7c7685205c39d653c2d90d6563481fb/jupyter_releaser/util.py#L399-L420

pyproject.toml
```diff
@@ -18,6 +18,8 @@ ignore = ["nbgrader/labextension/**", "yarn.lock"]
 
 [tool.jupyter-releaser.hooks]
 after-bump-version = ["python tools/post-bump.py"]
+before-build-npm = ["jlpm"]
```
jtpio, line 21: Wondering if we still need the `python -m pip install jupyterlab~=3.1` as well.

brichet: I removed it as `jupyterlab` was in `build-system` requires.

But after having a look at `jupyter_releaser` it seems that the npm is built before the python : 
https://github.com/jupyter-server/jupyter_releaser/blob/023124cbb7c7685205c39d653c2d90d6563481fb/jupyter_releaser/actions/draft_release.py#L68-L72.
So that requirement will not be available. 
Thanks @jtpio

jtpio (approved):
Thanks!

PR #1620, opened 2022-07-04
Automated Changelog Entry for 0.8.0a1 on main

jtpio:
Automated Changelog Entry for 0.8.0a1 on main
```
Python version: 0.8.0a1
npm version: nbgrader: 0.8.0-alpha.1
```

After merging this PR run the "Full Release" Workflow on your fork of `jupyter_releaser` with the following inputs
| Input  | Value |
| ------------- | ------------- |
| Target | jupyter/nbgrader  |
| Branch  | main  |
| Version Spec | next |
| Since | v0.8.0a0 |

PR #1621, opened 2022-07-04
Pin version of nbclassic<0.4.0

brichet:
The latest `nbclassic` version `0.4.0` breaks nbgrader so for now it should be pinned in a previous version.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -43,6 +43,7 @@ install_requires =
     traitlets<5.2.0
     jupyterlab<4
     jupyterlab_server
+    nbclassic<0.4.0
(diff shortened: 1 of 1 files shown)


brichet:
cc @jtpio 
Maybe for this release we can just pin the version of `nbclassic`, and depending on the answer in https://github.com/jupyter/nbgrader/pull/1620 we make the changes in `nbgrader`.

jtpio:
Thanks @brichet.

Yes it would make sense to add that pin for now. Maybe some fixes will also be landing in `nbclassic` directly.

PR #1626, opened 2022-07-07
Feature/mybinder

brichet:
Add binder configuration.
Try it : https://mybinder.org/v2/gh/brichet/nbgrader.git/feature/mybinder

cc @jtpio


brichet:
It seems the latest release of `jupyter_core` (`4.11.0`) breaks the tests : https://github.com/jupyter/jupyter_core/issues/275

brichet:
> If we want to go one step further we can also add a GitHub Action to post a Binder link for new PRs: https://github.com/jupyterlab/maintainer-tools#pr-binder-link

That would be nice.
Is the `github_token` mandatory? 
Does this mean that everyone pushing a PR needs to setup a personal token?

jtpio:
> Is the `github_token` mandatory?

hmm not sure why it's in the example snippet. We use this action in the `jupyter/notebook` repo without the token: https://github.com/jupyter/notebook/blob/main/.github/workflows/binder.yml

brichet:
> > Is the `github_token` mandatory?
> 
> hmm not sure why it's in the example snippet. We use this action in the `jupyter/notebook` repo without the token: https://github.com/jupyter/notebook/blob/main/.github/workflows/binder.yml

Thanks, I'll try without.
BTW the example you linked seems to use a token too.

brichet:
The new action didn't run.
I'm not sure if it should run on this PR or if it will be only available only for future PR.

brichet:
@jhamrick @jtpio do you think it can be merged ?

jtpio:
> I thought `opened` type was for all opened PR, but it only happen when the PR opens, so it's not supposed to be triggered now.

Right, and I think the action will first have to be merged before being effective.

jhamrick:
Very nice, thanks!

PR #1648, opened 2022-07-18
Add required github-token in workflow

brichet:
This PR adds a GITHUB-TOKEN to binder workflow, as required

### DIFF
--- a/.github/workflows/binder.yaml
+++ b/.github/workflows/binder.yaml
@@ -10,3 +10,5 @@ jobs:
       pull-requests: write
     steps:
       - uses: jupyterlab/maintainer-tools/.github/actions/binder-link@v1
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/binder.yaml
+++ b/.github/workflows/binder.yaml
@@ -11,4 +11,4 @@ jobs:
     steps:
       - uses: jupyterlab/maintainer-tools/.github/actions/binder-link@v1
         with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
+          github_token: ${{ secrets.github_token }}
--- a/.github/workflows/binder.yaml
+++ b/.github/workflows/binder.yaml
@@ -11,4 +11,4 @@ jobs:
     steps:
       - uses: jupyterlab/maintainer-tools/.github/actions/binder-link@v1
         with:
-          github-token: ${{ secrets.GITHUB_TOKEN }}
+          github_token: ${{ secrets.GITHUB_TOKEN }}


brichet:
@jtpio any idea why the link to Binder doesn't appear in the comments?
The Binder workflow passed but then disappeared.

.github/workflows/binder.yaml
```diff
@@ -10,3 +10,5 @@ jobs:
       pull-requests: write
     steps:
       - uses: jupyterlab/maintainer-tools/.github/actions/binder-link@v1
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
```
jtpio, line 14: Maybe we can try with `github-token`:

```suggestion
          github-token: ${{ secrets.GITHUB_TOKEN }}
```

jtpio: Hmm although this doesn't seem to be what the action expects here:

https://github.com/jupyterlab/maintainer-tools/blob/4af624a8519f8887712c7932931546399b0ff14d/.github/actions/binder-link/action.yml#L4

jtpio: Maybe lowercase `GITHUB_TOKEN`:

```suggestion
          github_token: ${{ secrets.github_token }}
```

brichet: My first attempt which failed was with `github-token`, that why I changed it in a new commit.
The `maintainer-tools` seems to wait for `github_token` input, and to rename it `github-token`  : 
https://github.com/jupyterlab/maintainer-tools/blob/4af624a8519f8887712c7932931546399b0ff14d/.github/actions/binder-link/action.yml#L16
I'll try to run again the failed workflow.

jtpio:
Re-opening the PR doesn't seem to trigger the check.

brichet:
> Re-opening the PR doesn't seem to trigger the check.

Yes, the workflow is triggered only when the PR opens the first time.
But I tried to open a PR on my repo and it didn't pass as well : https://github.com/brichet/nbgrader/pull/3

jtpio:
> But I tried to open a PR on my repo and it didn't pass as well : [brichet#3](https://github.com/brichet/nbgrader/pull/3)

Really odd. I also tried on my fork and same outcome.

brichet:
Yes, the token seems empty regarding at line 16 [here](https://github.com/jupyter/nbgrader/runs/7390458872?check_suite_focus=true#step:3:30). This is the same in my fork.

brichet:
@jtpio in fact the action used is that of the `main` branch while this PR is not merged. So it normal that it fails.

I tested it by :
- creating a fake branch from `main`
- merging this branch (`fix/binder-workflow`)  into it -> the action on Binder failed
- create an other PR (https://github.com/brichet/nbgrader/pull/6) -> the action on Binder passed

We should merge this one as the modified action will never been tested during the PR.

jtpio:
OK sounds good thanks @brichet for testing!

jtpio (approved):
Thanks!

PR #1650, opened 2022-07-28
New Feature: Automatic generation of test code

trevorcampbell:
This is a PR related to Discussion #1633 -- please see that discussion for context!

-------

Alright @jhamrick, as promised, here is the PR for the autotest functionality built into nbgrader. Major changes in this PR:
- `nbgrader/preprocessors/instantiatetests.py`: A new preprocessor that converts `### AUTOTEST` statements into test code. This preprocessor is included in the `generate_assignments` app and the `instantiate_tests` app.
- `nbgrader/converters/instantiate_tests.py`: A new converter that creates a `source_with_tests` folder to see the source copy of the assignment with autogenerated test code
- ~~`nbgrader/apps/instantiatetestsapp.py`: A new `instantiate_tests` app that allows instructors to generate the test code in their solution version (this is useful for checking/playing with/debugging the test code that gets generated). The instantiated version of the notebook still contains all solutions and appears in the `instantiated/` folder.~~ this was removed per @jhamrick request.
- ~~`nbgrader/preprocessors/limittestcellheights.py`: A new preprocessor that edits test code cell metadata with a cell height property to ensure they don't get too large/unwieldy in the release version. This is included in the `generate_assignments` app.~~ This was moved to a separate PR #1656 per @jhamrick request
- ~~`nbgrader/nbextensions/limit_cell_height/`: A new javascript extension responsible for making sure test cells don't get too large in the browser. Any test cell that would be larger than the max height gets a scroll bar instead.~~ This was moved to a separate PR #1656 per @jhamrick request
- Test code for all of the above
- Some new example notebooks in the `nbgrader/docs/source/user_guide/source` folder that illustrate the use of autotest, with test template YaML code in `nbgrader/docs/source/user_guide/tests.yml` illustrating multiple languages
- (minor unrelated change) an R codestub in the `clearsolutions` preprocessor

Please let @Alireza-Iranpour and myself know if you need any additional info!


trevorcampbell:
I see a bunch of the tests failing because the PyYAML package isn't installed. We included it in the `requirements.txt` file, but I'm sure that's the wrong place to put it. I'm trying to trace through the code to figure out where packages are listed for the `cmd = 'pip install -e .[tests]'` line in `tasks.py`, but struggling a bit... advice much appreciated.

trevorcampbell:
Tests pass on all ubuntu images now. There is a failure on the windows images, though, that occurs at test teardown. If we rerun the workflows here now, perhaps you'll have seen it before and know what to do about it.

trevorcampbell:
OK @jhamrick the PR is now officially ready. All tests (old and new) pass on all ubuntu images. 

There is still that [one issue](https://github.com/jupyter/nbgrader/pull/1650#issuecomment-1200106958) with test teardown on windows images; help or advice there would be much appreciated.

jhamrick:
Thanks so much! Sorry I haven't had a chance to look at this yet, but will be doing so today/tomorrow 😄

nbgrader/docs/source/user_guide/release/ps1/problem1.ipynb
```diff
@@ -1,31 +1,5 @@
 {
  "cells": [
-  {
```
jhamrick, line 3: It looks like there are some cells removed from this notebook---was this intentional?

nbgrader/docs/source/user_guide/release/ps1/problem2.ipynb
```diff
@@ -1,31 +1,5 @@
 {
  "cells": [
-  {
```
jhamrick, line 3: Same comment here about missing cells

nbgrader/docs/source/user_guide/release/ps2/problem3.ipynb
```diff
@@ -0,0 +1,468 @@
+{
```
jhamrick, line 1: Rather than creating two new assignments in the documentation, I think it might be cleaner to add some new example problems to the existing ps1. Could you remove ps2 and ps3 and instead add some example auto tests ps1?

nbgrader/__init__.py
```diff
@@ -55,6 +55,12 @@ def _jupyter_nbextension_paths():
             dest="validate_assignment",
             require="validate_assignment/main"
         ),
+        dict(
```
jhamrick, line 58: Hmm, so adding this new extension is going to require a bit more work in terms of updating documentation, demos, etc. in a variety of places than what you've done here. We'll also need to make sure it's compatible with JupyterLab, which may be a bit more work. What I'd suggest for now is to remove the extension code from this PR and made a new PR adding it in. Then we can focus on the autogenerated test functionality alone in this PR, and discuss what else needs to be added for the extension in the other PR.

nbgrader/docs/source/user_guide/release/ps1/problem1.ipynb
```diff
@@ -62,6 +36,12 @@
    "execution_count": null,
    "metadata": {
     "deletable": false,
+    "execution": {
```
jhamrick, line 39: The docs need to be rebuilt in order to exclude these sorts of extraneous metadata. Can you run `python tasks.py docs` from the root of the repo and commit the changes?

(For more details on updating documentation see https://nbgrader.readthedocs.io/en/stable/contributor_guide/documentation.html )

nbgrader/apps/instantiatetestsapp.py
```diff
@@ -0,0 +1,107 @@
+# coding: utf-8
```
jhamrick, line 1: I am a bit hesitant about adding a new command (especially one that requires creating a new directory in the nbgrader directory) as that will increase complexity and make things more confusing for folks new to nbgrader. But I guess the reason is so you can run validation on the source assignment with the actual tests that will be used (because if you run validation on the notebooks in `source`, they don't yet have the tests, and if you run validation in `release`, they don't have the solutions). Hmm.

What would you think about the following instead: rather than a new top-level command, we add an option to `GenerateAssignmentApp` (something like `nbgrader generate_assignment --instantiate_tests`) that will, as a byproduct of generating the assignment, save the intermediate notebooks to `source_with_tests/` (rather than `instantiated/`). Then the functionality is there to generate the notebooks if you want to get them with both the solutions and the hidden tests, but it's more of a debugging tool rather than a key step in the nbgrader process. I think it should be a pretty simple change, I think you can just add the following to `generateassignmentsapp.py` before running the generate converter (plus code to implement the option):

```
if self.instantiate_tests:
    converter = InstantiateTests(coursedir=self.coursedir, parent=self)
    try:
        converter.start()
    except NbGraderException:
        sys.exit(1)
```

nbgrader/coursedir.py
```diff
@@ -142,6 +142,18 @@ def _validate_notebook_id(self, proposal: Bunch) -> str:
         )
     ).tag(config=True)
 
+    instantiated_directory = Unicode(
+        'instantiated',
```
jhamrick, line 146: Maybe `source_with_tests` would be clearer than `instantiated`?

nbgrader/coursedir.py
```diff
@@ -142,6 +142,18 @@ def _validate_notebook_id(self, proposal: Bunch) -> str:
...
+        help=dedent(
+            """
+            The name of the directory that contains the version of the
```
jhamrick, line 149: This help string seems wrong

requirements.txt
```diff
@@ -1 +1 @@
--e .
```
jhamrick, line 1: This looks like an unrelated change

setup.cfg
```diff
@@ -44,6 +44,7 @@ install_requires =
     jupyterlab<4
     jupyterlab_server
     nbclassic<0.4.0
+    PyYAML==6.0
```
jhamrick, line 47: Rather than an exact version, can this be `PyYAML>=6.0` (or `<=`)?

nbgrader/docs/source/user_guide/tests.yml
```diff
@@ -0,0 +1,310 @@
+#kernel_name:
```
jhamrick, line 1: Is this commented out config needed?

jhamrick, line 1: This file is not getting copied when running `nbgrader quickstart` (which copies the example assignments from the docs), meaning that running `nbgrader generate_assignment` doesn't work for notebooks that include auto tests. Can you update the quickstart code so that it is? You can test that it works by running the following:

```
cd /tmp
nbgrader quickstart course101
cd course101
nbgrader generate_assignment ps2
```

jhamrick (changes requested):
This is really nice, thanks again for the PR and sorry for being slow to review it!

Overall I think it looks pretty good, but I have a few suggested changes (in comments) about the `instantiate_tests` command and the notebook extension, as well as a few more minor changes.

Also, I think this feature needs some more explicit documentation. Could you add an explanation of the auto test feature into the page on "creating and grading assignments"? You can base yours off of how the hidden tests are described: https://nbgrader.readthedocs.io/en/stable/user_guide/creating_and_grading_assignments.html#autograder-tests-cells-with-hidden-tests

nbgrader/apps/instantiatetestsapp.py
trevorcampbell: @jhamrick thanks for the review! We'll start working on these suggestions shortly. One clarification/question before we get started:

> I am a bit hesitant about adding a new command (especially one that requires creating a new directory in the nbgrader directory) as that will increase complexity and make things more confusing for folks new to nbgrader. But I guess the reason is so you can run validation on the source assignment with the actual tests that will be used (because if you run validation on the notebooks in source, they don't yet have the tests, and if you run validation in release, they don't have the solutions)

The `instantiate_tests` converter is not a required part of the workflow. Users can simply use nbgrader as they would right now (run `generate_assignment` to go from `source/` to `release/`). The `instantiate_tests` converter is optional -- just in case the instructor wants to play with an editable version of the assignment that has both solutions and test code in it.

I think I'd personally be confused that `generate_assignment --instantiate_tests` doesn't produce a `release/` version. I think I'd also assume (before digging into docs) that adding the `--instantiate_tests` flag forces autotest to run (and that it won't run without that flag), which isn't the case. 

Semi-related: I am happy to change `instantiated/` to `source_with_tests/`. But it is a bit unwieldy, and if someone is running the `instantiate_tests` converter (or `--instantiate_tests` flag), the name `instantiated/` would hopefully not be confusing... I don't feel too strongly about it though.

Thoughts? Let us know how you'd like to proceed and we'll get to it. I skimmed your other comments and I agree with all!

trevorcampbell: (oops -- we actually had one more question regarding adding a new example notebook to the docs! See above :) )

nbgrader/docs/source/user_guide/release/ps2/problem3.ipynb
jhamrick: Ah right, I see, I missed ps3 was in R! Yes it's probably a good idea to showcase that. Though in that case I'd definitely make sure to link to ps3 (maybe rename it to ps2) from the docs so that it's clear that example exists.

nbgrader/apps/instantiatetestsapp.py
jhamrick: > The instantiate_tests command is not a required part of the workflow.

Yes, I understand, it's just my gut feeling that making it a top-level command implies that it is (though I guess who knows, that's really an empirical question!)

> I think I'd personally be confused that generate_assignment --instantiate_tests doesn't produce a release/ version.

Ah so my thinking was that it would *both* generate a `release/` version and a `source_with_tests/` version. In other words, because making the tests is an intermediate step of `generate_assignment`, adding the flag tells nbgrader to save out that intermediate step separately. Maybe it would be clearer if, instead of calling the flag `--instantiate_tests`, we called it `--generate_source_with_tests` or something like that? And then that would address your other point about the name `source_with_tests/` being a little confusing. What do you think?

trevorcampbell: Ah -- generating both a `release/` and `source_with_tests/` (given the `--blah` flag) seems like reasonable behaviour to me!

But looking at the code again, I'm not sure how easy it would be to modify the `generate_assignment` converter to output two copies, with one only using a partial set of preprocessors. It seems the base converter assumes one input notebook / folder -> one output notebook / folder. I could be wrong here -- maybe you know a convenient way to pull that off. I think unless there's a clean way of doing that (without hacking the `generate_assignments` converter substantially) I'd rather keep it as a separate converter that follows existing code patterns.

(I'm of course happy to add documentation to make it super clear that the `instantiate_tests` converter is optional and what its purpose is, in case we go that route)

trevorcampbell: (if we keep the separate converter, would you prefer `nbgrader generate_source_with_tests` or `nbgrader instantiate_tests`?)

jhamrick: So it wouldn't be a change to the converter itself, but to the application. I think the code I posted above that modifies `generateassignmentsapp.py` should do the trick---the idea is basically just to run both the instantiate tests converter *and* the generate assignmemt converter.

(Technically, this would cause the instantiate tests preprocessor to run twice, but I don't think the overhead should be a problem)

trevorcampbell: I dig it -- happy to go with that solution then! We'll get to it.

nbgrader/docs/source/user_guide/release/ps1/problem1.ipynb
Alireza-Iranpour: Not sure why that happened. I replaced ps1 notebooks with their original version from the main repository

nbgrader/docs/source/user_guide/release/ps1/problem2.ipynb
Alireza-Iranpour: Replaced with their original version from the main repository to address this

nbgrader/docs/source/user_guide/release/ps2/problem3.ipynb
Alireza-Iranpour: The source folder is now organized as:
ps1 including 3 Python notebooks (problem3.ipynb is a new notebook with autotest examples)
ps2 including 1 R notebook (problem1.ipynb)

nbgrader/__init__.py
Alireza-Iranpour: The extension related changes have been removed from this PR. We will make a new PR for that separately.

nbgrader/docs/source/user_guide/release/ps1/problem1.ipynb
Alireza-Iranpour: We ran `python tasks.py docs` and committed the changes

nbgrader/apps/instantiatetestsapp.py
Alireza-Iranpour: We added a new flag called --generate_source_with_tests to the `generateassignmentsapp` and removed the other app

nbgrader/coursedir.py
Alireza-Iranpour: Renmaed to `source_with_tests`

nbgrader/coursedir.py
Alireza-Iranpour: Help sting is fixed
`    source_with_tests_directory = Unicode(
        'source_with_tests',
        help=dedent(
            """
            The name of the directory that contains the version of the
            assignment that will contain both the solutions and the instantiated 
            test cases to be used by instructors. This corresponds to
            the `nbgrader_step` variable in the `directory_structure` config
            option.
            """
        )
    ).tag(config=True)`

setup.cfg
Alireza-Iranpour: Changed to `PyYAML>=6.0`

nbgrader/docs/source/user_guide/tests.yml
Alireza-Iranpour: They were meant to serve as a template for new kernels
We removed them to make it cleaner

Alireza-Iranpour: We updated the `quickstartapp.py` to copy the `tests.yml` file as well

`
# copying the tests.yml file to the course directory
tests_file_path = os.path.abspath(os.path.join(
    os.path.dirname(__file__), '..', 'docs', 'source', 'user_guide', 'tests.yml'))
shutil.copyfile(tests_file_path, course_path)
`

requirements.txt
Alireza-Iranpour: We originally changed this because we put PyYAML there but we moved that to setup.cfg. We have replaced this file with the original requirements.txt, so there is no longer any change here

nbgrader/docs/source/user_guide/release/ps1/problem1.ipynb
Alireza-Iranpour: Btw, some new files were created as a result of running this command. Were those expected?
Example: user_guide/autograded/bitdiddle/ps1/problem3.html

trevorcampbell:
@jhamrick we've handled all of your comments now. Should be ready for you again!

One question did come up. Running `python tasks.py docs` generated some extra files, not sure if you wanted those to be committed. See https://github.com/jupyter/nbgrader/pull/1650#discussion_r954317094

trevorcampbell:
Hi @jhamrick -- it's been a few months now, and the PR has fallen behind the main branch a bit. This will need a bit of cleanup to get ready for review again. I'm happy to do the cleanup to bring it back up to speed, but I'll hold off on doing that until you let me know that you're ready!

trevorcampbell:
This PR is superceded by #1817 . We can close this one I think (hopefully the milestone tag can be transferred over to the new PR).

trevorcampbell:
After #1817 was merged this PR is no longer needed. Closing!

PR #1656, opened 2022-08-24
New Feature: Limit test cell height extension

Alireza-Iranpour:
See the thread for PR #1650. This PR was split from that one. This PR is just for the limit test cell height extension.

### DIFF
--- a/nbgrader/__init__.py
+++ b/nbgrader/__init__.py
@@ -55,6 +55,12 @@ def _jupyter_nbextension_paths():
             dest="validate_assignment",
             require="validate_assignment/main"
         ),
+        dict(
+            section="notebook",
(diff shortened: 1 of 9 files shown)


trevorcampbell:
`nbextensions` is gone now as of 0.9.x . So this one will need to be revised. But I think JLab comes with a configurable max cell height property / scrollable cells now, doesn't it? So perhaps we don't even need this feature any more.

trevorcampbell:
@brichet Yes -- I think we would need to re-do it from scratch anyway now with the new JLab

brichet:
@trevorcampbell thanks for answering, let's close it, then.
Feel free to reopen it if required.

PR #1666, opened 2022-09-21
Add maximum size for entire directory as submission option

franswel:
The previous maximum size option was applied to individual files only. This adds a separate option to limit the maximum size of the entire directory. The option is called max_dir_size and is implemented similarly to max_file_size. The default is same as the max file size.

### DIFF
--- a/nbgrader/tests/apps/test_nbgrader_submit.py
+++ b/nbgrader/tests/apps/test_nbgrader_submit.py
@@ -2,6 +2,7 @@ import os
 import datetime
 import time
 import stat
+import pytest
 
 from os.path import join, isfile, exists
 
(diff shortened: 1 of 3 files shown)


nbgrader/tests/apps/test_nbgrader_submit.py
```diff
@@ -231,7 +232,7 @@ def test_submit_include(self, exchange, cache, course_dir):
         filename, = os.listdir(join(exchange, "abc101", "inbound"))
         assert not exists(join(exchange, "abc101", "inbound", filename, "foo.txt"))
 
-    def test_submit_include(self, exchange, cache, course_dir):
```
rkdarst, line 234: This is renamed since it's the same name as above - it was probably named wrong.

rkdarst (approved):
This has also been in use at our site for quite a while now.  The default max size of 100M probably makes more sense to be interperted as max size of all files, rather than individula ones anyway, right?

This is a problem we've had plenty of times here.

brichet:
Thanks @franswel and @Gehock

PR #1667, opened 2022-09-21
groupshared feedback re-release fix

franswel:
Groupshared users could not re-release student feedback due to shutil.copy functionality used by release_feedback. Shutil.copy tries to chmod the destination file which does not work in groupshared scenarios where files have different owners but same groups. Therefore groupshared users cannot copy a new generated feedback with the same name over the old one owned by another user. In the fix the file gets copied in a temporary location and is moved over the old file. 

Tests were not implemented as I believe it is not possible to test interactions requiring multiple users automatically. To replicate the issue, you need two users with groupshared enabled that release feedback for the same student's assignment.

### DIFF
--- a/nbgrader/exchange/default/release_feedback.py
+++ b/nbgrader/exchange/default/release_feedback.py
@@ -67,9 +67,12 @@ class ExchangeReleaseFeedback(Exchange, ABCExchangeReleaseFeedback):
                 submission_secret = fh.read()
 
             checksum = notebook_hash(secret=submission_secret, notebook_id=notebook_id)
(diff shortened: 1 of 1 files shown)


rkdarst:
This is a follow up of #1000, and is live at our site for a while now.  It's also nice that it makes feedback releasing an atomic operation, so I think this should be "good to go".  Existing tests should cover the actual functionality, and like @franswel said testing the multi-user interaction is probably too hard.

brichet (approved):
Thanks @franswel for the fix and @Gehock for reviving this PR.
It looks good to me.

PR #1669, opened 2022-09-26
Switch to hatch build backend

blink1073:
Use `hatch` and the `hatch-jupyter-builder` plugin, while still honoring `NBGRADER_NO_LAB`.

I compared the sdist and the wheel contents before and after using the `hatch_jupyter_builder.compare_migration` script.


brichet:
Thanks @blink1073 
This look like good to me.

We should update the conda recipe after this PR, which doesn't work now anyway : https://github.com/conda-forge/nbgrader-feedstock/pull/56

blink1073:
Sounds good, if you want to make a release I can update the conda recipe.

brichet:
> Sounds good, if you want to make a release I can update the conda recipe.

It's released : https://github.com/jupyter/nbgrader/releases/tag/v0.8.1.

Issue #1664, opened 2022-09-16
`View Feedback` link not working whin in lab mode


perllaghu:
Dockerized notebooks running in a Kubernetes cluster

I'm testing NBgrader 0.8 for our service, and everything is working fine (including our external exchange service) - however the `View Feedback` link does not work.

I've tried both the `pypi` version and the latest `master` branch from gitlab.

I've tried cutting my docker image right back to remove possible clashes with other libraries

I can confirm that `fetch_feedback` works, and the files are pulled down; and I can confirm that the link works when in the `classic` interface - same docker image, just started with a different `default_url` value

From what I can see....

- In classic mode, the link is `..../tree/<feedback_path>`
- In lab mode, the link has no attributes, but has an event link that runs 

         function() {
          n.commands.execute("filebrowser:go-to-path", {
            path: i
          })
        }

and viewed from a javascript console, this calls `..../api/contents/<feedback_path>` to get to `..../lab/tree/<feedback_path>`

Can someone confirm that the `View Feedback` link works for them, in lab mode?

perllaghu:
(and to bring a note of levity.... can we call this `The Kronenbourg Issue`?

I'll get my coat...)

perllaghu:
Hmm...... I wonder if there's some proxy shenanigans going on.

Do you know of any other API calls that I can test to see if there's a general problem with my docker-in-k8s setup, or it's specifically this call?

brichet:
I just came across this again, it looks like you already had this problem during the first tests of nbgrader on JupyterLab with your environment: 
https://github.com/jupyter/nbgrader/pull/1588#issuecomment-1146054704

A possible reason could be that your *feedback path* is not in the JupyterLab root path.
In your installation, is there JupyterHub and a service JupyterLab instance running (see https://nbgrader.readthedocs.io/en/stable/configuration/jupyterhub_config.html?highlight=service#example-use-case-one-class-multiple-graders) ?
If so, the *assignment list* may not be running on the correct server and may not have access to the *feedback path*.

perllaghu:
Indeed - I noticed it when the changes first came round.... then I had to go and finalise all the notebook offerings for our service in time for testing & rollout for the new academic year [with all the lead-time that requires] - so wasn't able to pursue properly at the time.

My feedback path is standard.... though with `path_includes_course` set true (so `~/<course_coce>/<assignment_code>/feedback/`

There's no 'hub involved, we have a custom front-end that spawns into a k8s cluster directly.

[...]

PR #1675, opened 2022-09-30
Fix view feedback link when root_dir is set

brichet:
The `view feedback` link does not work when the `root_dir` of JupyterLab is different from './'.
The reason is that the feedback path is not relative to the `root_dir` but to the directory where nbgrader started.

This PR solves the problem by changing the feedback path to a relative path from `root_dir`.

I observed the same issue with classic notebook (in version `0.7.x`) if the `notebook_path` is different from './'.
This PR should probably be back ported to `0.7.x` branch.

Close https://github.com/jupyter/nbgrader/issues/1664


brichet:

perllaghu:
I can confirm this branch works for my docker notebook images, which are deployed in a K8s cluster.

Issue #1612, opened 2022-06-24
Tests on nbextensions are broken with latest selenium version (4.3.0)


brichet:
The latest version of `selenium 4.3.0` breaks the nbextensions tests with the following error :
`E           AttributeError: 'WebDriver' object has no attribute 'find_elements_by_css_selector'`

This has been fixed in `main` branch in https://github.com/jupyter/nbgrader/pull/1611 by pinning `selenium < 4.3`, as the main branch is no longer supposed to support nbextensions.

This issue should be fixed in  the branch `0.7.x`.

jtpio:
This seems to be related to the removal of previously deprecated methods:

> * Deprecated find_element_by_* and find_elements_by_* are now removed (#10712)

https://github.com/SeleniumHQ/selenium/blob/2fbfc62f565026ba57cf700e14179012c6ebef4c/py/CHANGES#L2

Probably we can remove the pin and update the code to the new API as mentioned in the docstring:

> Please use find_element(by=By.CSS_SELECTOR, value=css_selector) instead",

brichet:
Fixed in https://github.com/jupyter/nbgrader/pull/1677

PR #1677, opened 2022-09-30
Fixes selenium tests

brichet:
EDIT : Fix https://github.com/jupyter/nbgrader/issues/1612

All the methods like `find_element_by_*` and `find_elements_by_*` have been removed from selenium 4.3.0.

This PR replaces them by `find_element(By.*, ...)` and `find_elements(By.*, ...)`.

### DIFF
--- a/nbgrader/tests/nbextensions/formgrade_utils.py
+++ b/nbgrader/tests/nbextensions/formgrade_utils.py
@@ -31,11 +31,11 @@ def _check_url(browser, port, url):
 
 def _check_breadcrumbs(browser, *breadcrumbs):
     # check that breadcrumbs are correct
-    elements = browser.find_elements_by_css_selector(".breadcrumb li")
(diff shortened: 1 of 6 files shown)


brichet:
@meeseeksdev backport to main

jtpio (approved):
Looks good, thanks!

Looks like this should fix https://github.com/jupyter/nbgrader/issues/1612? If so then probably the pin added in https://github.com/jupyter/nbgrader/pull/1611 can be removed?

brichet:
> Looks like this should fix #1612?

Yes, forgot to mention it.



> If so then probably the pin added in #1611 can be removed?

It was not ported on this branch (`0.7.x`), only on `main`, but I'm backporting these changes on `main` and will remove the pin.

jtpio:
> It was not ported on this branch (`0.7.x`), only on `main`, but I'm backporting these changes on `main` and will remove the pin.

Ah right, sounds good :+1:

PR #1684, opened 2022-10-17
Compatibility with Notebook7 and jupyterLab4

brichet:
This PR brings compatibility to Notebook7 and JupyterLab4, both not yet released.

In Notebook7 the *formgrader*; *assignment list* and *course list* are opened in a tab in the `tree` view, while the notebook files and feedbacks are opened in a new browser tab.

image
image

#### Remains to be fixed:
- [x] some links in *formgrader* and *assignment list* open a specific path in file browser. In `Notebook<7`, this is opening a new browser tab with a tree view. In this version, it changes the current path in file browser tab. Unfortunately, it does not activate the file browser tab, which makes it look like nothing is happening.
- [x] the feedback files are displayed as plain HTML text instead of being interpreted.
   EDIT: fixed in https://github.com/jupyter/notebook/pull/6583

Fell free to try it and discuss about expected behaviors.


tuncbkose:
Is there any reason not to remove `nbgrader/tests/nbextensions`?

jtpio:
JupyterLab 4 final is now released, and Notebook 7 beta is available. Maybe this PR could be resumed?

brichet:
> JupyterLab 4 final is now released, and Notebook 7 beta is available. Maybe this PR could be resumed?

Sure, I was waiting for some opinions about https://github.com/jupyter/nbgrader/discussions/1749 first, but in any case, I should have time to work on it in the next few weeks.

brichet:
> Is there any reason not to remove `nbgrader/tests/nbextensions`?

There is no reason, except that I want to clean properly everything related to nbextensions (not only the tests).

nbgrader/tests/ui-tests/course_list.spec.ts
```diff
@@ -0,0 +1,252 @@
...
+let test = jupyterLabTest;
+let mainPanelId = '#jp-main-dock-panel';
+let menuPanelId = '#jp-menu-panel';
+let mainPanelTabCount = 1;
+
+const baseTestUse = {
+  tmpPath: tempPath,
+  mockSettings: {
+    '@jupyterlab/apputils-extension:notification': {
+      fetchNews: 'false'
```
jtpio, line 31: Quick heads-up there were some updates on this config in core JupyterLab in https://github.com/jupyterlab/jupyterlab/pull/14609.

We can check whether the default `mockSettings` from JupyterLab is already automatically applied, if we should add `checkForUpdates: false` as well.

jtpio:
Thanks @brichet for working on this :+1: 

Diff looks good at first glance. Do you think it would be ready to be merged so it can be tested in a `0.9.0a0` release?

brichet:
Thanks @jtpio for looking at it.
Yes, I think we can merge it and test it in a pre-release.

jtpio:
Sounds good, thanks!

PR #1567, opened 2022-05-03
Upgrade nbconvert

brichet:
This PR upgrades dependencies to `nbconvert>=6` ,and `jinja2>=3`.

- preprocessing cell is now done with the native `nbclient` method (instead of overriding the `nbconvert<6` one)
- `feedback` and `formgrade` templates are now using the `nbconvert>6` template system
- uses `ClearMetadataPreprocessor` in `exporter` when running tests to remove the new `execution` which which was breaking comparison tests

Closes https://github.com/jupyter/nbgrader/issues/1373, https://github.com/jupyter/nbgrader/issues/1382

PR #1690, opened 2022-10-24
Make execute preprocessor traits configurable

tmetzl:
The following traits are not configurable:
```
interrupt_on_timeout = Bool(True)
allow_errors = Bool(True)
raise_on_iopub_timeout = Bool(True)
timeout = Integer(30)
```
This means you can't adjust the timeout via ``nbgrader autograde <My_Assignment> --Execute.timeout=120``.  This will throw an error ``"[AutogradeApp | WARNING] Config option `timeout` not recognized by `Execute`.  Did you mean one of: `iopub_timeout, startup_timeout, timeout_func`?" ``

By making the traits configurable the command line or config file option is used correctly.


tmetzl:
I copied the original help messages from [nbclient](https://github.com/jupyter/nbclient/blob/main/nbclient/client.py)

brichet:
Thinking again about this, I'm not sure about these changes.

Some of these parameters need to be set:
- `allow_errors = Bool(True)` is necessary to execute the whole Notebook when validating or auto-grading. Setting it to `False` will break validation and auto-grading process.
- `raise_on_iopub_timeout = Bool(True)` has been included in https://github.com/jupyter/nbgrader/commit/9d0f32d8b13323dc5ada4da89c630cbea971ce5e, related to `nbconvert` dependency
- `timeout = Integer(30)` has been introduced in https://github.com/jupyter/nbgrader/pull/1567, and as far as I remember it can be change but should not be set to `None`. Sometime the kernel didn't respond (i don't remember the reason) and `nbgrader` waited endlessly.

tmetzl:
This PR is not changing any of the default values. It just makes them configurable. I don't see harm in letting the user override the sensible default values picked here. What I really need is the ability to configure the timeout.

Currently I have to use my own ExecutePreprocessor since otherwise everything times out in 30s which is not enough time to autograde my students.

brichet:
I agree that the timeout should be configurable.

I remember why it was implemented the first time. It is necessary when the cell contains code to redirect the traceback (eg. https://github.com/jupyter/nbgrader/blob/main/nbgrader/tests/apps/files/submitted-cheat-attempt.ipynb). In this specific case, the execution never ended, and the timeout is necessary.

Perhaps we could add a minimal value > 0 and do not allow none value to guarantee a positive timeout.

shreve:
> In this specific case, the execution never ended, and the timeout is necessary.

I was looking into why this is the case, and I realized it doesn't have to be. This attack causes an unhandled exception in the kernel, which is why execution hangs indefinitely. I just issued a PR upstream in iPython to handle (read: ignore) that exception so we can still return a result.

I tested this alongside allowing `Execute.timeout = None` and the tests pass!

I have no idea how long that will take to merge upstream, but once it does, I think we're safe on that front. The upstream `nbclient.NotebookClient.timeout` is `Integer(allow_none=True)` so I don't think this is too wild of an idea.

nbgrader/preprocessors/execute.py
```diff
@@ -41,10 +89,15 @@ class Execute(NbGraderPreprocessor, ExecutePreprocessor):
         """)
     ).tag(config=True)
 
-    def on_cell_executed(self, **kwargs):
+    @validate('timeout')
+    def _validate_timeout(self, proposal):
+        value = int(proposal['value'])
```
shreve, line 94: This current validation throws an exception when the value is None.

nbgrader/preprocessors/execute.py
```diff
@@ -14,10 +14,58 @@ class UnresponsiveKernelError(Exception):
...
+        ),
+    ).tag(config=True)
+
+    timeout = Integer(
+        30,
+        help=dedent(
+            """
+            The time to wait (in seconds) for output from executions.
+            If a cell execution takes longer, a TimeoutError is raised.
+            ``None`` or ``-1`` will disable the timeout. If ``timeout_func`` is set,
```
shreve, line 64: It would be my preference to only support None and positive integers. If that's the case, this docstring needs to be updated. If not, more code changes are needed.

shreve:
Thanks for your work on this so far, @tmetzl !

brichet:
Thanks @shreve, it works well with `ipython>=8.10.0`.
Maybe we could:
- add a dependency to `ipython>=8.10.0`
- remove the traits `timeout` (and the associated `validate` decorator), since any value should be allowed now.

I'm still confused about the value of being able to set `allow-errors=False`, as it would break some functionalities, but maybe I missed a point on that.

tmetzl:
Once you decided how to move forward please let me know what I need to change to make it work.

shreve:
> I'm still confused about the value of being able to set allow-errors=False

I think from an end-user perspective it makes sense. If you have an assignment where each test must pass before the next one should be attempted, why waste the resources running cells that you know will fail?

Do you have more detail about what systems/functionality will fail in that scenario? I'm not incredibly familiar with the whole codebase yet.

brichet:
> Do you have more detail about what systems/functionality will fail in that scenario?

This would break the rendering of the validation dialog.
The validation process catches all errors and displays them with a specific formatting, which is lost if the error is raised by `nbclient`.
Here are the 2 renderings, the one on the left is an extract of the current and the one on the right with `allow_errors=False`.
 

I think it deserves more discussion since it breaks a part of the rendering of `nbgrader`.
Maybe you should open an issue if this feature is really necessary.

shreve:
> I think it deserves more discussion since it breaks a part of the rendering of nbgrader.
> Maybe you should open an issue if this feature is really necessary.

@brichet I agree. The change to allow_errors is more influential than I thought. This would be great before a 1.0, but deserves it's own ticket. 

@tmetzl I made a PR against yours with the changes I think we need here.

I was expecting to be able to remove the `timeout` and `interrupt_on_timeout` traits, but they need to be included because their default values are different from the parent default values. 🙃

shreve:
Oof, ok. iPython dropped support for python 3.7 a little over a year ago, so setting the minimum version broke nbgrader's compatibility. 

https://github.com/ipython/ipython/commit/5b9aa9c88325012439384b5a704e2b3d183c3366

In that commit, they cite [NEP 29 — Recommend Python and NumPy version support as a community policy standard](https://numpy.org/neps/nep-0029-deprecation_policy.html) which suggests a time table for dropping support for older pythons which NumPy recommends to the rest of the community as well. 

In order to merge this change, we now need to figure out a strategy for dealing with python 3.7 and implement it. I would personally be in favor of upping the minimum from 3.7 to 3.8, but we could pick another way forward. @brichet what would your preference be? Does a decision like this need wider approval?

brichet:
Thanks @shreve for changes and @tmetzl for merging it.

> @brichet what would your preference be?

I don't have a strong opinion on dropping unmaintained Python version.
I opened https://github.com/jupyter/nbgrader/pull/1750 to drop Python 3.7 in tests and include Python 3.11 instead. Let's wait for it to rebase the current branch.

> I was expecting to be able to remove the `timeout`

I think it can be removed, since it was introduced to avoid infinite loop in case of a cheat attempt using `showTraceback()` (fixed in https://github.com/ipython/ipython/pull/13934).
This can also be done in a follow up PR.

shreve:
Great! I see that change was merged. I think that means with a rebase off main, this PR should be good to go!

shreve:
@tmetzl Are you able to do this update? Either rebase off main or merge main in. It would be awesome to get this merged in time for 0.8.2.

tmetzl:
> @tmetzl Are you able to do this update? Either rebase off main or merge main in. It would be awesome to get this merged in time for 0.8.2.

@shreve I think I did now. Please let me know if there is anything left to do.

brichet:
Thanks @tmetzl and @shreve

shreve:
Beautiful! Thanks all!

PR #1698, opened 2022-10-28
Allow current working directory for nbgrader config file only

brichet:
This PR is to (re)allow loading configuration file from current working directory.
This feature has been removed from [jupyter_core](https://github.com/jupyter/jupyter_core/commit/1fd79f9c192d4c56235ec1847f18d3b079b722b9) for security reasons (https://github.com/advisories/GHSA-pq7m-3gw7-gq5x).

This will fix the tests on extensions which are broken since the last release of jupyter_core.

For what I understand it should only load configuration file for nbgrader, it should not reintroduce the security error.

### DIFF
--- a/nbgrader/apps/baseapp.py
+++ b/nbgrader/apps/baseapp.py
@@ -351,12 +351,17 @@ class NbGrader(JupyterApp):
         if self.config_file:
             paths = [os.path.abspath("{}.py".format(self.config_file))]
         else:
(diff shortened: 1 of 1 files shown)


jtpio:
Thanks @brichet :+1: 

I think that makes sense if a specific file is explicitly loaded and not any in `cwd` like it was the case before.

Looks like the `jupyter_core` advisory has now been published: https://github.com/jupyter/jupyter_core/security/advisories/GHSA-m678-f26j-3hrp

brichet:
Thanks for sharing the correct link.

PR #1718, opened 2022-12-30
Update dependency to jupyter-server>=2

brichet:
This PR updates the dependency to `jupyter-server>=2`.

It also adds a new dependency to `pytest-jupyter` (when installing in test) as it is not included anymore in `jupyter-server`


brichet:
Closed in favor of https://github.com/jupyter/nbgrader/pull/1776

Issue #1418, opened 2021-03-10
"validate" cannot detect error in R code


jnishii:
I'm trying to use nbgrader for R language, but the "Validate" button and autograder doesn't detect errors in answer cells.
I tried the following commands to throw error, but "validate" button always returns the message "Success! Your notebook passes all the tests" and I always got full score.

```
raise NotImplementedError()
```
```
fail()
```
```
e <- simpleError("test error")
stop(e)
```
```
throw('NotImplementedError') 
```

When I run the cells above, they return Error, so I think nbgrader seems not to detect the thrown error.
I'm currently using nbgrader for python on the same system without problems.
Is there some solution?

### Operating system
Ubuntu 16.04.6 LTS

### `nbgrader --version`
nbgrader version 0.6.1

### `jupyterhub --version` (if used with JupyterHub)
0.9.6

### `jupyter notebook --version` 
6.0.2

### Expected behavior

Detect errors in notebook

### Actual behavior

"Validate" button doesn't detect erros.

perllaghu:
Two questions here:

1) are you expecting the error to stop execution, or
2) are you getting _pass_ when you expect _fail_ ?

For the latter - I had a similar problem when trying to do stuff with a Stata kernel (see https://github.com/kylebarron/stata_kernel/issues/373 for my embarrassment)

The bottom line is you need a fairly recent version on nbgrader to get the updated logic for catching the `output.output_type == "stream" and output.name == "stderr"` errors

perllaghu:
Hmmm ok - so the `output.output_type` is `error`, which should be picked up by  ..... no, wait

We need to check what _validate_ is supposed to do - I think it just runs the notebook, and shows any output errors.

It certainly does that in a python3 notebook.

I'm doing some testing now - but I'll try adding some errors to an R notebook once I'm done, and report back

perllaghu:
OK.... here's what I have.

1. We need to set up the converter to produce the correct failure conditions in the _release_ notebook
In `nbgrader_config.py` I have:
```
c.ClearSolutions.code_stub = dict(
    python = '# YOUR CODE HERE\nraise NotImplementedError()',
    R = '# YOUR CODE HERE\nstop("NotImplementedError")'
    )
```

2. In the _source_ notebook, my _autograded answer_ looks like this:
```
squares <- function (n) {
    ### BEGIN SOLUTION
    if (!(is.numeric(n) & n > 0)) {
        stop("ValueError")
        }
    data <- 1:n; n
[...]

PR #1731, opened 2023-02-01
Fix validation for failed cells with zero points

tuncbkose:
If a cell gets full points during validation, check if max_points is 0 and the cell actually failed. Fixes (part of) #1418.

If this looks good, I can also add some related tests.

### DIFF
--- a/nbgrader/utils.py
+++ b/nbgrader/utils.py
@@ -61,6 +61,18 @@ def is_locked(cell: NotebookNode) -> bool:
     else:
         return cell.metadata['nbgrader'].get('locked', False)
 
+
+def has_failed(cell: NotebookNode) -> bool:
+    """Returns True if the cell contains an output indicative of an error"""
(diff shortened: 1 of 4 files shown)


nbgrader/utils.py
```diff
@@ -61,6 +61,18 @@ def is_locked(cell: NotebookNode) -> bool:
     else:
         return cell.metadata['nbgrader'].get('locked', False)
 
+
+def has_failed(cell: NotebookNode) -> bool:
+    """Returns True if the cell contains an output indicative of an error"""
+    if not cell.cell_type == 'code':
+        raise ValueError("cell is not a code cell, so it can't fail")
```
brichet, line 68:
```suggestion
        return False
```
Maybe we don't need to raise here

brichet (commented):
Thanks @tuncbkose.
I just have a comment.

brichet:
Thanks @tuncbkose

Issue #1736, opened 2023-02-24
Failing tests


tuncbkose:
It seems to me that there are two issues causing tests to fail.
1- JupyterLab [announcements](https://jupyterlab.readthedocs.io/en/stable/user/announcements.html) pop-up is interfering with `test_formgrader.spec.ts`, and
2- `node-canvas` is requested for a version that doesn't have a build for the `node` used in the testing environment, and it can't be built. (Related?: https://github.com/jupyterlab/jupyterlab/pull/13722) 

I am writing them down in case someone else encounters them too. Maybe I can create a pr as well soon.

brichet:
Thanks @tuncbkose for reporting this.
I think the second point should been fixed by https://github.com/jupyter/nbgrader/pull/1737

I'll take a look at the first issue.

brichet:
I close it as fixed by https://github.com/jupyter/nbgrader/pull/1737 and https://github.com/jupyter/nbgrader/pull/1743.
Feel free to re-open it if necessary.

PR #1737, opened 2023-02-27
Fixes for testing CI

tuncbkose:
This pr fixes a broken link in the docs and updates `node-canvas` so that a binary is actually available.

A testing error caused by JupyterLab Announcement extension remains unfixed because I don't know what is the best way to disable that (isolated to the testing process).


tuncbkose:
The cancellations are all timeouts (which is also present in other recent prs) and the labextension fails are due to the announcement extension, so no surprises there.

I don't really understand how check_release works yet, so I am not sure why it failing.

brichet:
Thanks @tuncbkose, the tests that fail are indeed not related

brichet:
Fixes a part of https://github.com/jupyter/nbgrader/issues/1736

PR #1741, opened 2023-03-03
Remove codecov badge

brichet:
This PR removes the broken `codecov` badge in the `README.md` file.
I think some right (that I don't have) are necessary to fix it.

I'll open an issue to restore it.

cc. @jhamrick


brichet:
The tests that fail are not related.

PR #1746, opened 2023-03-03
Assignment names breaking regex

### LINKED ISSUE #1738: Assignments_ids with `()` in them fail in formgrader
Discovered when using  `notebook: 6.4.11` and `nbgrader: 0.7.1` under linux

Tested & confirmed using `notebook: 6.4.12` and `nbgrader: 0.8.1` under linux

### Expected behavior

When collecting notebooks known to be submitted, the `# Submissions` column should give a value, and a link to be able to grade submissions

### Actual behavior

When collecting notebooks known to be submitted, the `# Submissions` column remains 0

### Cause of the behavior

There are a number of places that use `regexp` to disassemble a path to find details of course-code, assignment-code, student-code, and/or notebooks.

There is a problem if any of those elements have characters in them that can be interpreted as a "regex special character" [. $ * + ? {} \ [] | () ... and various combinations involving the above.]

For example, if an assignment is called `123 (abc)` - it will happily flow through the system, however formgrader will always count `0` students as NbGrader.api.get_submitted_students() has a regexp `<snip>/submitted/(?P<student_id>.*)/123 (abc)` ... and this matches `123 abc`, with the `abc` in a MatchObject, **not* `123 (abc)`

### Tested solution

Increase the number of characters tested for validation in `assignment_id`

- Edit `server_extensions/formgrader/static/js/manage_assignments.js` in the `var createAssignment` section (used in the formgrader UI)
- Edit `coursedir.py` in `def _validate_assignment_id(..` (needed for the CLI commands)

I suggest the following characters will confuse the regexp system, and should be excluded: `+()/\\[]{}$^`

perllaghu:
Extends the list of characters that need to be excluded from assignment names

closes #1738

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -124,7 +124,9 @@ package-lock.json
 .goutputstream*
 .idea
 .mypy_cache
-
+.envrc
+.node-version
+.direnv/
 
 # playwright output tests directory
 playwright-tests/
--- a/nbgrader/coursedir.py
+++ b/nbgrader/coursedir.py
@@ -78,14 +78,17 @@ class CourseDirectory(LoggingConfigurable):
             The assignment name. This MUST be specified, either by setting the
             config option, passing an argument on the command line, or using the
             --assignment option on the command line.
+
+            The assignment name cannot include any of the characters +()/\\[]\{\}$^
             """
         )
     ).tag(config=True)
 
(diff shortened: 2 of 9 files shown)


perllaghu:
Closing to rework....

Issue #1202, opened 2019-08-26
Consider how to prevent access to feedback from other students


jhamrick:
@rkdarst noted in #1200:

> One idea I had was to, at submission time, generate a random token and put it into the submitted notebook and some cache. Then, there is less coupling between a hash algorithm and retrieving the notebook later (which is probably a good thing)...

> On the other hand, tracking tokens separately from the notebook makes more work. They could be embedded in the notebook metadata, but you have the same problem as the next paragraph.

> All this actually this makes me think: if a student A shares their notebook with student B, is B then able to retrieve A's results and grade? It would be best to avoid this if possible...

As I responded there, I'm not really sure of a way around this access issue, barring having better access controls (which we should have once hubshare exists...). But I'm opening an issue here to at least track that this is a problem, and maybe brainstorm ways to deal with it.

jhamrick:
Also moving discussion from #1120 over here since that thread is already so long 🙂Ping @rkdarst and @Sefriol .

I guess with @Sefriol 's PR (#1186) the concern with being able to list and read other students' feedback is alleviated (with the exception of #1202), now that the feedback directory in the exchange is not listable and the hash is computed from the cache.

@Sefriol brought up the additional concern that:

> currently fetch assignment is meant for students only (according to documentation), but it accepts student ID or no ID specified as a parameter

I *think* with #1200 this should be ok now, since the hash is now computed from the student ID and the original notebook. You can try to override the student id if you want, but it won't work unless you actually have the original notebook submitted by that student.

Sefriol:
hubshare seems pretty dead project to me or is there any other alternatives on jupyterlab side?

Anyway, I think using cache and random string seems like the best option to me, unless we really want to store additional information into the hash.

rkdarst:
One interesting thing is that, if grading is a function of only the input state, knowing the notebook (plus the tests from your own notebook) is the same as knowing the grade. Except manual grading.  And I doubt data protection authorities would be impressed by that argument.  Anyway I don't consider this a top-priority problem, but maybe I should...

We can *almost* use the random string added to the filename when it's submitted as a token to retrieve it (either directly or part of the nb hash).  But it's not stored in the student cache (per #1104, if it was students could more easily find their submitted assignments and modify them).

One _could_ add two random strings, one to be used when retrieving and one not saved...

rkdarst:
[...]

PR #1753, opened 2023-03-14
Duplicate grade_ids

tuncbkose:
Here is a proposal on how to deal with duplicate `grade_id`s during autograding. Feedback is very welcome. Note that three of these commits might be worth considering merging independent of this PR:
- 3c51d4d fixes an old testing bug unrelated to this PR (I can make a pr for this separately if requested).
- c28a52a changes CheckCellMetadata error message to add duplicate cells as a possible reason why the error is encountered. This might be desirable as it would allow users to notice the duplication problem during validation and try to fix it before submission.
- bac7c76 tweaks the error handling in the base `converter.py` to allow continuing with the other submitted notebooks (from the same student) in an assignment even if one of them is problematic. This way, for example, if an assignment had multiple notebooks and only one contained a duplicated cell,  the rest would still get autograded.

## Dealing with duplicate grade_ids
The main idea of this PR is to add a `duplicate` flag to cells that were observed to have duplicate `grade_id`s during autograding, so that they can be ignored and the rest of the notebook can be autograded. At the end of autograding, the `duplicate` flag is caught in order to raise an error and let the instructor know that some cells were ignored. These ignored cells are also marked with a comment for easy identification, however, the problem of the instructor editing the notebook remains unfixed (as it breaks the feedback workflow, https://github.com/jupyter/nbgrader/issues/1202).

3b3c672 changes the order cells are processed by `DeduplicateIds´ with the idea that often if one is copy-pasting a cell, the pasted cell goes below the original one. This makes it more likely for the original to be more important for the grading, implying that we should be ignoring the cells lower in the notebook.


Gehock:
Rebased on top of the current main branch

Gehock:
Updated the schema version on a couple of missed notebooks (didn't expect test notebooks to live outside of the `nbgrader/tests` folder).

PR #1755, opened 2023-03-15
Copy integrity

tuncbkose:
- 11da69c: Adds error messages for when `fetch/fetch_feedback` fails
- 783e0d2: Wraps copying function into `try/except/finally` to ensure that in the event of an error, already copied files have correct permissions
- 8d8d898: Wraps copying function into `try/except` to ensure that even if not all files were copied, `timestamp.txt` is still created and feedback workflow doesn't break later
  - Right now, the student receives a "Submission Error" and no indication that something is submitted. However, files are still copied to the exchange folder.
  - I think it is fine to not prevent any copying (i.e. delete already-copied files if an error is encountered at any point), because if a student sees "Submission Error", they will likely try submitting again, hopefully after the problem is fixed.
  - The error message user sees could be clearer.


tuncbkose:
Is there a way to set some nbgrader configuration for a single test? Only https://github.com/jupyter/nbgrader/commit/783e0d25e5814c557414029a269447779b8faed8 remains untested, but it applies only when `CourseDirectory.groupshared == true`.

PR #1768, opened 2023-03-28
Assignment names breaking regex

### LINKED ISSUE #1738: Assignments_ids with `()` in them fail in formgrader

perllaghu:
This extends the list of banned assignment-name characters beyond `+` to include a bunch that break the path regex code.

closes #1738

### DIFF
--- a/.github/workflows/test-docs-python-nbextensions.yml
+++ b/.github/workflows/test-docs-python-nbextensions.yml
@@ -17,7 +17,7 @@ defaults:
 jobs:
   test_nbgrader:
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 25
+    timeout-minutes: 45
 
     env:
       # NOTE: UTF-8 content may be interpreted as ascii and causes errors
@@ -28,19 +28,19 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [windows-latest, ubuntu-20.04]
+        os: [windows-latest, ubuntu-22.04]
         group: ["docs", "nbextensions", "python"]
-        python: ["3.7", "3.8", "3.9", "3.10"]
+        python: ["3.8", "3.9", "3.10", "3.11"]
         exclude:
           - os: windows-latest
-            group: docs
-          - python: "3.7"
             group: docs
           - python: "3.8"
             group: docs
           - python: "3.9"
             group: docs
-          - python: "3.7"
+          - python: "3.10"
+            group: docs
+          - python: "3.8"
             group: nbextensions
     steps:
       # This is how you set an environment variable in a GitHub workflow that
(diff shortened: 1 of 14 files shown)


brichet:
Thanks @perllaghu.

I wonder if there could not be a way to manage special characters when running the regex match. Something like `re.escape`, to avoid constraining the assignment name.

brichet:
On my side there is some confusion in this PR, I see some modifications already in the branch `main` (e.g. *zero point validation*).

perllaghu:
> Thanks @perllaghu.
> 
> I wonder if there could not be a way to manage special characters when running the regex match. Something like `re.escape`, to avoid constraining the assignment name.

This is essentially an extension of the existing routine - which prevent `+` in assignment names.

perllaghu:
> On my side there is some confusion in this PR, I see some modifications already in the branch `main` (e.g. _zero point validation_).

I'll rebase on master :)

smith-kyle:
Just a heads up, you can review notebook changes like this using [GitNotebooks](https://gitnotebooks.com).

e.g. this pull request [https://gitnotebooks.com/jupyter/nbgrader/pull/1768](https://gitnotebooks.com/jupyter/nbgrader/pull/1768)

It's free for public repos, so might be helpful for y'all.

lahwaacz:
The last commit https://github.com/jupyter/nbgrader/pull/1768/commits/88915ab776de97b52d754d479709fbc36f84fa7b is very confusing: it includes changes from https://github.com/jupyter/nbgrader/commit/67069391deaf4ac390952e633926631b5271a328 (in `main`), reverts 2 previous commits in this pull request and maybe also some unrelated changes (I can't tell at this point).

Please do an [interactive rebase](https://thoughtbot.com/blog/git-interactive-rebase-squash-amend-rewriting-history), _drop_ the reverted commits, _squash_ corrections into previous commits and overall make sure that this pull request can be properly reviewed.

brichet:
> This is essentially an extension of the existing routine - which prevent `+` in assignment names.

Agree, let's move forward with this PR and open an issue to allow these character in future development.
@perllaghu can you clean this PR please ?

perllaghu:
The sane move now will be to drop this & redo the work on a fresh `master`

Issue #1682, opened 2022-10-09
Students deleting grade cells causes formgrader to report "needs manual grading" forever


trevorcampbell:
Hi nbgrader team,

(I'm using nbgrader 0.7.1 on ubuntu 22.04)

I've found that students sometimes accidentally delete grade cells (yes, they're read-only; with a class of 800 students, *someone* always manages to do this, even after disabling keyboard shortcuts, educating them about the perils of changing read-only cells, etc).

If this happens, formgrader will say "needs manual grading". But the TAs cannot do anything about it, since they can't see the cell in formgrader as it was removed from the notebook. To them it looks like grading is done, even though formgrader thinks it isn't. 

Proposed solution idea:
- during autograding, check if any grade cells are missing in the notebook
- try to re-insert them if the cells are otherwise all there and in the right order (or sufficiently many around the place where that missing cell should be)
- if nbgrader can't determine where the cell should go (e.g. if the student has mangled the notebook sufficiently), then assign a score of 0 and move on. Or perhaps add the cell at the end of the notebook.

tuncbkose:
@trevorcampbell If this is still an issue, would you mind taking a look at [this branch](https://github.com/AaltoSciComp/nbgrader/tree/missing_cells) and autograde using `nbgrader autograde {whatever} --OverwriteCells.add_missing_cells=True`?

tuncbkose:
I've done this quick-test already, and it seems to work fine. 

One clarification: I am guessing that you are referring to the "Needs Manual Grading" check mark. I observe that the problem being that when a grade cell is missing, the notebook is always marked "Need Manual Grading", even if all the present grade cells are graded. So we want the check mark to disappear, which is the opposite of what you are describing.

Am I misunderstanding/missing something? In either case, I'll make a pr to hopefully get more feedback from others and link it to this issue.

trevorcampbell:
> One clarification: I am guessing that you are referring to the "Needs Manual Grading" check mark. 

Yes, sorry -- that was my mistake. I haven't used it in a bit :) your description of it is correct.

nthiery:
I have been regularly struck by this issue with my 200+ course, and #1770 
sounds very promising! I am looking forward having it in production for the
next iteration of the course in Fall.

PR #1770, opened 2023-03-29
Add missing grade_cells before autograding

tuncbkose:
Add a new flag `OverwriteCells.add_missing_cells`, that, when set to `True`, compares the submitted notebook to the database before autograding and adds missing grade cells to (approximately) correct locations.

Fixes https://github.com/jupyter/nbgrader/issues/1682 (with the provided flag).

## Notes
- Missing grade cells are added immediately after the previous grade/solution cell, because presumably the notebook structure is repeated blocks of
```
question students should answer (solution cell)
test for the above (grade cell)
test for the above (grade cell)
...
```
- The above doesn't apply to manually graded answers, and currently I can't think of a good way to find their correct location based on the database
- Missing cells could be added based on the notebook in the `released/` directory, but there is no guarantee they are the same. Instructors can overwrite it, in which case the original notebook might not be recoverable. I am not sure if the database is completely safe, but it should be safer.

## To-dos
If this looks good, I will also
- [x] add tests
- [x] extend to restoring task cells as well ("Needs Manual Grade" behaves the same as [#1682](https://github.com/jupyter/nbgrader/issues/1682) if task cells are missing). 
- [x] Make the notification configurable?

### DIFF
--- a/nbgrader/preprocessors/overwritecells.py
+++ b/nbgrader/preprocessors/overwritecells.py
@@ -3,13 +3,83 @@ from nbformat.v4.nbbase import validate
 from .. import utils
 from ..api import Gradebook, MissingEntry
 from . import NbGraderPreprocessor
+from ..nbgraderformat import MetadataValidator
 from nbconvert.exporters.exporter import ResourcesDict
 from nbformat.notebooknode import NotebookNode
+from traitlets import Bool, Unicode
 from typing import Tuple, Any
+from textwrap import dedent
 
 
 class OverwriteCells(NbGraderPreprocessor):
     """A preprocessor to overwrite information about grade and solution cells."""
+
+    add_missing_cells = Bool(
+        False,
+        help=dedent(
+            """
+            Whether or not missing grade_cells should be added back
+            to the notebooks being graded.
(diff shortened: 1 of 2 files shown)


brichet:
Thanks @tuncbkose.

> Fixes #1682 (with the provided flag).

On my side the cell is indeed added, but the status is still *need manual grading*. I tried deleting an autograded cell.
Is that expected or have I missed something ?

tuncbkose:
> ...  but the status is still need manual grading.

On a quick attempt, I couldn't reproduce this. I will investigate, but it would be great if you can let me know what you did step by step.

tuncbkose:
On a more involved attempt, still couldn't reproduce this status mismatch.

brichet:
> On a more involved attempt, still couldn't reproduce this status mismatch.

My fault, it was actually needed a manual grading.
Sorry for the confusion and extra attempts.

Can you try rebasing on main ?

tuncbkose:
This pr shouldn't change anything for ui tests, so failure is unrelated?

brichet:
Thanks @tuncbkose

PR #1775, opened 2023-04-11
Toggle visibility of all student names

### LINKED ISSUE #951: Ability to show/toggle all student names in formgrader
Follow up on:
https://github.com/jupyter/nbgrader/issues/575 
... which led to:
https://github.com/jupyter/nbgrader/pull/617

I would highly appreciate the possibility to **toggle the visibility of all student names--with one click**. We do grading anonymously and I think it's a good default.

However, when students want to discuss individual issues **after they've received their grades/feedback**, I currently have to toggle up to 200 submission names to find the formgrader version of their submission. 
Thanks for any suggestions!

tuncbkose:
Adds a new button in gradebook_notebook_submissions view to toggle visibility of all student names, closes #951.

### DIFF
--- a/nbgrader/server_extensions/formgrader/static/js/gradebook_notebook_submissions.js
+++ b/nbgrader/server_extensions/formgrader/static/js/gradebook_notebook_submissions.js
@@ -37,11 +37,13 @@ var SubmittedNotebookUI = Backbone.View.extend({
     showName: function () {
         this.$reveal.parent().find(".name-shown").show();
         this.$reveal.parent().find(".name-hidden").hide();
+        updateToggleNamesButtonLabel();
     },
 
     hideName: function () {
         this.$reveal.parent().find(".name-hidden").show();
         this.$reveal.parent().find(".name-shown").hide();
+        updateToggleNamesButtonLabel();
     },
 
     render: function () {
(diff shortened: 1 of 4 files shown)


nbgrader/server_extensions/formgrader/static/js/gradebook_notebook_submissions.js
```diff
@@ -187,6 +187,20 @@ var loadSubmittedNotebooks = function () {
     });
 };
 
+// button to toggle all student names
+var names_hidden = true;
```
brichet, line 191: If all the names have already been displayed using the eye icons, clicking on the button will do nothing the first time.
Maybe instead of a global variable we could check in the function if some names are hidden.

nbgrader/server_extensions/formgrader/templates/gradebook_notebook_submissions.tpl
```diff
@@ -17,6 +17,11 @@ var notebook_id = "{{ notebook_id }}";
 </ol>
 {%- endblock -%}
 
+<!-- Button to toggle all student names -->
+{%- block messages -%}
+<button onclick="toggleAllNames()">Toggle student names</button>
```
brichet, line 22: Can text be dynamic ?
Something like **Show all names** if some names are still hidden and **Hide all names** when all the names are already displayed.
This status could also define what should do the button (see previous comment)

brichet (commented):
Thanks @tuncbkose.
I have some comment on the PR.

And also can you add test in Notebook and Jupyterlab ?

nbgrader/server_extensions/formgrader/static/js/gradebook_notebook_submissions.js
```diff
@@ -187,6 +189,33 @@ var loadSubmittedNotebooks = function () {
     });
 };
 
+// at least one hidden name -> label="Show All Names"
+// no hidden name -> label="Hide All Names"
+let updateToggleNamesButtonLabel = function() {
+    const button = document.getElementById("toggle_names");
+    const icons = document.getElementsByClassName("glyphicon name-hidden");
+    const icon_array = [...icons];
+    if (icon_array.some(el => el.checkVisibility())){
```
tuncbkose, line 198: `checkVisibility()` might not be the best option because it is [not supported in safari](https://caniuse.com/mdn-api_element_checkvisibility). I can look for an alternative if that is a concern.

nbgrader/server_extensions/formgrader/static/js/gradebook_notebook_submissions.js
tuncbkose: I changed the logic so that the button label is updated properly when pressing individual eye icons. The button's logic then depends on its label.

tuncbkose:
The labextension tests are not implemented properly yet. I need to fix them (and my local playwright setup). I'll request a review when they are ready.

On a related note, for the Lab 4 tests, would the tests work immediately? If needed, I can take a look into that pull request and prepare something that will work.

nbgrader/server_extensions/formgrader/static/js/gradebook_notebook_submissions.js
tuncbkose: The playwright browsers seemed too old for this anyway, so I changed it to `$(el).is(':visible')`.

brichet:
> ~The labextension tests are not implemented properly yet. I need to fix them (and my local playwright setup). I'll request a review when they are ready.~
> 
> On a related note, for the Lab 4 tests, would the tests work immediately? If needed, I can take a look into that pull request and prepare something that will work.

Thanks, I'll do it for this test.

PR #1776, opened 2023-04-12
Upgrade to Jupyter Server 2

shreve:
My strategy is to update the dependency spec then fix whatever tests start failing.


luisarturopena:
Any updates on this? 😃

brichet:
Thanks @shreve.

That would be a regression from https://github.com/jupyter/nbgrader/pull/1586, not sure this is what we want.
Maybe it will be the way to go for this branch (Jupyterlab 3.x and Notebook 6.x).

cc. @jtpio @blink1073 who may have an opinion on it.

nbgrader/server_extensions/assignment_list/handlers.py
```diff
@@ -9,10 +9,10 @@
 from textwrap import dedent
 
 from jupyter_server.utils import url_path_join as ujoin
-from jupyter_server.base.handlers import JupyterHandler
 from traitlets import Unicode, default
 from traitlets.config import LoggingConfigurable, Config
 from jupyter_core.paths import jupyter_config_path
+from notebook.base.handlers import IPythonHandler as JupyterHandler
```
jtpio, line 15: Why import the handler from the `notebook` and not `jupyter_server`?

blink1073:
I'd say the notebook 6.x compat branch should keep its pin on `jupyter_server<2`, and then use `>=2.0` for notebook 7.x going forward.

lahwaacz:
The main branch already has `jupyter_server>=2`: https://github.com/jupyter/nbgrader/commit/1b5ffa93db15012f15ac17bf475eab2420fe73bf#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711

PR #1778, opened 2023-04-13
Upgrade to support jupyter_client 8

shreve:
When upgrading the dependency spec to include v8, some python tests broke (see https://github.com/jupyter/nbgrader/pull/1728). The tests were failing with a timeout due to indefinitely blocked IO.

This originates in an infinitely-looping notebook and is caused by a misconfiguration of the client connection. Currently using `nbclient.NotebookClient` with a `jupyter_client.manager.KernelManager` results in a deadlock. nbclient doesn't seem aware of this issue because there is no test covering it. `NotebookClient` uses `AsyncKernelManager` by default.

So why isn't nbgrader using it? nbgrader uses the NotebookClient via nbconvert which just realized the same issue. Up until v7.3.1, nbconvert has been hard-coding the default kernel manager to the sync version. (see https://github.com/jupyter/nbconvert/issues/1964)

Rather than require nbconvert>=7.3.1, this commit sets our own default value to the async kernel manager.


shreve:
I'm not familiar but it looks like the `codecov` dependency got pulled from pypi.

https://github.com/codecov/codecov-python

tuncbkose:
opened #1780 for `codecov`

brichet:
Kicking the CI

brichet:
@shreve can you rebase to check the tests ?

PR #1788, opened 2023-05-18
Upgrade to sqlalchemy>=2

tuncbkose:
Most of the required changes so far were fairly innocent.

One of the warnings when running on `sqlalchemy<2` was [this one](https://docs.sqlalchemy.org/en/14/changelog/migration_14.html#change-5150) about changing cascade_backref behavior. It is still a bit unclear to me what consequences this change might have, but tests were passing on `sqlalchemy<2` with using `future=True` in `Session` and `Engine`, and they are passing now.

### DIFF
--- a/nbgrader/api.py
+++ b/nbgrader/api.py
@@ -7,11 +7,10 @@ import subprocess as sp
 
 from sqlalchemy import (create_engine, ForeignKey, Column, String, Text,
                         DateTime, Interval, Float, Enum, UniqueConstraint,
-                        Boolean, inspect)
+                        Boolean, inspect, text)
 from sqlalchemy.orm import (sessionmaker, scoped_session, relationship,
-                            column_property)
+                            column_property, DeclarativeBase)
 from sqlalchemy.orm.exc import NoResultFound, FlushError
-from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.ext.associationproxy import association_proxy
 from sqlalchemy.exc import IntegrityError, StatementError
 from sqlalchemy.sql import and_, or_
@@ -26,7 +25,8 @@ from typing import List, Any, Optional, Union
 from .auth import Authenticator
 
 
-Base = declarative_base()
+class Base(DeclarativeBase):
+    pass
 
 
 def new_uuid() -> str:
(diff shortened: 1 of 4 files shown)


tuncbkose:
Failures are unrelated.

nbgrader/api.py
```diff
@@ -7,11 +7,10 @@
 
 from sqlalchemy import (create_engine, ForeignKey, Column, String, Text,
                         DateTime, Interval, Float, Enum, UniqueConstraint,
-                        Boolean, inspect)
+                        Boolean, inspect, text)
 from sqlalchemy.orm import (sessionmaker, scoped_session, relationship,
-                            column_property)
+                            column_property, declarative_base)
```
shreve, line 12: The preferred method for 2.0 is to use the `DeclarativeBase` class. This still works, but switching to the class makes a lot of sense in this PR.
see: https://docs.sqlalchemy.org/en/20/orm/mapping_api.html#sqlalchemy.orm.declarative_base

tuncbkose: Thank you for letting me know. I don't recall seeing a mention of this in the migration document (maybe I missed it), but I agree that it makes sense to switch. Will fix accordingly.

pyproject.toml
```diff
@@ -46,7 +46,7 @@ dependencies = [
     "python-dateutil>=2.8",
     "rapidfuzz>=1.8",
     "requests>=2.26",
-    "sqlalchemy>=1.4,<2",
+    "sqlalchemy>=1.4,<3",
```
brichet, line 49: Will it still work with `sqlalchemy<2` ?

brichet (commented):
Thanks @tuncbkose, that looks good to me.
I don't really know how to be sure everything will work as before, not sure if the current tests cover it.
cc @jhamrick

pyproject.toml
shreve: The tests are still passing for me locally with sqlalchemy 1.4.46. That might be helpful to add to the test matrix if we want to continue supporting <2 moving forward.

tuncbkose: My understanding is that `sqlalchemy==1.4` supports all the big changes for `sqlalchemy==2`. Only the `future=True` flags might change behavior (like [this](https://docs.sqlalchemy.org/en/14/changelog/migration_14.html#cascade-backrefs-behavior-deprecated-for-removal-in-2-0)).

brichet:
@tuncbkose can you please rebase it to check if all the tests pass ?

brichet:
Thanks @tuncbkose

nbgrader/api.py
tuncbkose: @shreve @brichet `DeclarativeBase` is apparently a sqlalchemy 2.x addition (without being present in 1.4), so I should have bumped up the minimum version. 

The previous `declarative_base` imported from `sqlalchemy.orm` should work with both 1.4 and 2.x.

The difference between them [is](https://docs.sqlalchemy.org/en/20/orm/mapping_api.html#sqlalchemy.orm.DeclarativeBase) for type-checkers, which at least right now is not a big deal for `nbgrader`. So I think it is worth going back to `declarative_base` over `DeclarativeBase`. What do you think?

PR #1797, opened 2023-07-05
Human readable course titles for multi-class nbgrader on JupyterHub

### LINKED ISSUE #1795: Introduce (human readable) course titles in addition to course IDs
Courses in nbgrader have the `course_id` property with several limitations (at least URL safe if used with JupyterHub). Would be great to have an additional property `course_title`. At the moment course list extension and assignment list extension show `course_id` to users (instructors and students). Showing a `course_title` would allow for whitespace, special characters, Unicode aso. in GUI.

**Concrete use case:** We (a small team at a university) are working on LMS integration of JupyterHub and nbgrader based on LTI. Coming to JupyterHub from the LMS our scripts and tools automatically sync course structures and enrolement lists from LMS to nbgrader. At the moment we have to convert LMS course titles to nbgrader course IDs losing information (if the LMS title is in Chinese, generating an equivalent nbgrader ID is almost impossible). Having identical titles in LMS and nbgrader would improve user experience and acceptance of the system.

**Some details/ideas on how to implement this:**
* If running multiple formgrader services on a JupyterHub, one might add something like
  ```
  c.NbGrader.course_titles = {
      'course_1': 'Mathematics I for Engineers',
      'course_2': 'Mathematics II for Engineers'
  }
  ```
  to global nbgrader config file.
* If running with a local formgrader (no JupyterHub), then
  ```
  c.CourseDirectory.course_id = 'testcourse'
  c.CourseDirectory.course_title = 'My Test Course'
  ```
  in the user's nbgrader config file would be reasonable.
* The course list server extenion's `formgraders` API endpoint returns a list of course dicts. Each dict would get an additional item `course_title` based on above config values.
* The course list labextension (and also nbextension) then use `course_title` instead of `course_id` from the API endpoint's dicts.

I've implemented above ideas and it works quite well. Similar should be possible for assignment list extension (not implemented yet).

If next to me there are other people looking for better course titles in nbgrader, I could make a PR. But not sure whether there are better ideas how to implement such a feature than mine.

jeflem:
Course list and assignment list extensions show nbgrader's `course_id` (and/or the formgrader service's name) in the GUI. A course's `course_id` property has to be (at least) URL save if nbgrader is running in a multi-class JupyterHub setting. Thus, no white space, no specical characters (German umlauts, Chinese,...) allowed.

This PR adds a `course_title` property to nbgrader courses and modifies course list and assignment list extensions (server/lab/nb) to show `course_title` where appropriate. Using `course_title` for GUI (and for nothing else) allows for arbitrary course titles.

If the new `course_title` property is not set, then behavior is as before (show `course_id`). In single-class environments set `c.CourseDirectory.course_title` to some string. In multi-class JupyterHub environments set
```
c.NbGrader.course_titles = {
    'course_id_1': 'Title of Course 1',
    'course_id_2': 'Title of Course 2'
}
```

This PR closes #1795.

Tests pass up to one: `tests/nbextensions/test_course_list.py::test_no_formgrader`. But that test even fails without my modifications. My interpretation is that that test has to fail if everything is okay (test disables fromgrader extension and then fails due to timeout waiting for formgrader to load). Would be a different PR to make this test pass if everything is okay (pass if `TimeoutException` has been raised).

Would be great if someone could carefully review this PR. It's my first contribution to nbgrader, maybe I missed something important.

### DIFF
--- a/demos/demo_multiple_classes/global_nbgrader_config.py
+++ b/demos/demo_multiple_classes/global_nbgrader_config.py
@@ -2,3 +2,7 @@ from nbgrader.auth import JupyterHubAuthPlugin
 c = get_config()
 c.Exchange.path_includes_course = True
 c.Authenticator.plugin_class = JupyterHubAuthPlugin
+c.NbGrader.course_titles = {
+    'course101': 'Course 101',
+    'course123': 'Course 123'
+}
--- a/nbgrader/apps/baseapp.py
+++ b/nbgrader/apps/baseapp.py
@@ -12,7 +12,7 @@ import traitlets.log
 from jupyter_core.application import JupyterApp
 from textwrap import dedent
 from tornado.log import LogFormatter
-from traitlets import Unicode, List, Bool, Instance, default
+from traitlets import Unicode, List, Bool, Instance, default, Dict
 from traitlets.config.application import catch_config_error
 from traitlets.config.loader import Config
 
@@ -65,6 +65,16 @@ class NbGrader(JupyterApp):
     flags = nbgrader_flags
 
     _log_formatter_cls = LogFormatter
+
+    course_titles = Dict(
+        {},
+        help=dedent(
+            """
+            Dict mapping course IDs to human readable course titles. If there is
+            no title for a course, ID is shown.
+            """
+        )
+    ).tag(config=True)
 
     @default("log_level")
     def _log_level_default(self) -> int:
--- a/nbgrader/coursedir.py
+++ b/nbgrader/coursedir.py
@@ -31,6 +31,21 @@ class CourseDirectory(LoggingConfigurable):
             self.log.warning("course_id '%s' has trailing whitespace, stripping it away", proposal['value'])
         return proposal['value'].strip()
 
+    course_title = Unicode(
+        '',
+        help=dedent(
+            """
+            A human readable course name for display purposes.
+            """
+        )
+    ).tag(config=True)
+
+    @validate('course_title')
+    def _validate_course_title(self, proposal):
+        if proposal['value'].strip() != proposal['value']:
+            self.log.warning("course_title '%s' has trailing whitespace, stripping it away", proposal['value'])
+        return proposal['value'].strip()
+
     student_id = Unicode(
         "*",
         help=dedent(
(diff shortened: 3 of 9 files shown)


lahwaacz:
The `nbgrader/nbextensions` directory is gone on the `main` branch, so this will need to be updated for the [0.9.x version](https://github.com/jupyter/nbgrader/discussions/1749)

jeflem:
Thanks for considering this PR! I rebased on main. I'm not a git pro. So if something went wrong please let me know.

> I tested it only in single course environment (not with jupyterhub), it worked well for course list, but not for assignment list.

For non-local formgraders (JHub) course titles are shown in the assignement list's drop-down where you select a course. For local formgraders this does not work (course ID is shown). I'll try to explain why, as far as I understand the design of course list and assignment list extensions:
* The course list server extension lists all formgraders. In https://github.com/jupyter/nbgrader/blob/ba2a40491ff0f9ae0a08baeb9cf0c92ab670a05c/nbgrader/server_extensions/course_list/handlers.py#L56 we have access to the local formgrader's config file and may read `c.CourseDirectory.course_title`.
* The assignment list server extension extracts course IDs from assignments, see https://github.com/jupyter/nbgrader/blob/ba2a40491ff0f9ae0a08baeb9cf0c92ab670a05c/nbgrader/server_extensions/assignment_list/handlers.py#L184 without knowing whether the course belongs to a local or a non-local formgrader. Thus, we cannot access the config file. Only knowledge from global configuration is available, that is, the `c.NbGrader.course_titles` dict.

Current solution (not showing title in assignment list for local formgraders) is not optimal, but good enough for my use case (no local formgraders). If we want to see titles of local formgraders in the assignment list, too, one solution would be to always use the `c.NbGrader.course_titles` config option. Even for local formgraders (and course specific config files) we would have to provide the mapping of IDs to titles in this dict. So instead of `c.CourseDirectory.course_title = 'Course 101'` one could write
```
c.NbGrader.course_titles.update({'course101':  'Course 101'})
```
It's important to use `update` here to not overwrite items defined in other config files. Don't know whether this is a better solution. Now we lost the symmetry between `c.CourseDirectory.course_id` and `c.CourseDirectory.course_title`.

All other solutions I thought about require lots of code changes (listing all formgraders and compare their course IDs to course IDs extracted from assignments, for example). Ideas welcome!

perllaghu:
> All other solutions I thought about require lots of code changes (listing all formgraders and compare their course IDs to course IDs extracted from assignments, for example). Ideas welcome!

Bear in mind, as you work your way through your thinking, that not everyone uses the default exchange mechanism.... so make sure you can cope with exchange handlers that use the current data formats :)

PR #1807, opened 2023-08-05
Remove jupyter_notebook_config.json since notebook server is not supported anymore

lahwaacz:
There is already `jupyter_server_config.d/nbgrader.json` and I don't see why `jupyter_notebook_config.json` should be any different.


lahwaacz:
@brichet What do you think about this change?

brichet:
Thanks @lahwaacz.
As far as I understand this config file is now useless since the main branch is not using `NotebookApp` but only `ServerApp` (see https://jupyter.org/enhancement-proposals/28-jupyter-server/jupyter-server.html).
It was useful to keep both for compatibility with Notebook<7 and Jupyterlab (branch `0.8.x`).
We can probably try to remove it instead.

lahwaacz:
@brichet Oh, right. I found a few more NotebookApp instances and removed them too.

brichet (approved):
LGTM
Thanks @lahwaacz

brichet:
@lahwaacz I've taken the liberty of renaming the PR for the sake of consistency.

PR #1815, opened 2023-08-21
Add unused arg for docs/cleandocs

### LINKED ISSUE #1814: Generating docs does not work
If I run

```
python3 tasks.py docs
```

from the repository root, I get the following output:

```
  File "/home/trevor/repos/dsci/nbgrader/tasks.py", line 176, in <module>
    ns.func(ns, args)
TypeError: docs() takes 1 positional argument but 2 were given
```

This appears to be related to this recent commit:

https://github.com/jupyter/nbgrader/blame/af47e5eaf880d456f79030e68c0feb0de1fb6313/tasks.py#L176

where these two functions were not updated:
https://github.com/jupyter/nbgrader/blame/af47e5eaf880d456f79030e68c0feb0de1fb6313/tasks.py#L35
and
https://github.com/jupyter/nbgrader/blame/af47e5eaf880d456f79030e68c0feb0de1fb6313/tasks.py#L46

It should be fairly straightforward to fix these by adding the `args` argument to `docs()` and `cleandocs()`

trevorcampbell:
Closes #1814

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -32,7 +32,7 @@ def _check_if_directory_in_path(pth, target):
     return False
 
 
-def docs(ns):
+def docs(ns, args):
     del ns  # unused
     run('git clean -fdX nbgrader/docs')
     if not WINDOWS:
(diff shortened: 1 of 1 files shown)


trevorcampbell:
@brichet one fix and this seems to be good now. Thanks for triaging so quickly!

brichet:
Thanks @trevorcampbell, and that make me realize that the documentation is probably not up to date.

Issue #1816, opened 2023-08-22
CI ui-tests are not actually passing


trevorcampbell:
The current commit af47e5e looks like it is passing all `ui-tests`, but there are quite a few tracebacks / errors in the ui-tests logs that aren't being caught as failures. I'm not sure if these are actual problems that weren't noticed because they're failing silently, or they are known and being ignored on purpose, but I figured I'd point it out here in case!

For example, see the log for the CI run on af47e5e :

https://github.com/jupyter/nbgrader/actions/runs/5925665890/job/16065530603

Under the "Run Playwright Tests" section, you will see quite a few errors. The first is related to the exchange being unwriteable:

```
[WebServer] [E 2023-08-21 11:34:21.644 ServerApp] Traceback (most recent call last):
      File "/home/runner/work/nbgrader/nbgrader/nbgrader/server_extensions/assignment_list/handlers.py", line 85, in list_released_assignments
        assignments = lister.start()
      File "/home/runner/work/nbgrader/nbgrader/nbgrader/exchange/abc/list.py", line 23, in start
        super(ExchangeList, self).start()
      File "/home/runner/work/nbgrader/nbgrader/nbgrader/exchange/default/exchange.py", line 133, in start
        self.ensure_root()
      File "/home/runner/work/nbgrader/nbgrader/nbgrader/exchange/default/exchange.py", line 59, in ensure_root
        self.fail("Unwritable directory, please contact your instructor: {}".format(self.root))
      File "/home/runner/work/nbgrader/nbgrader/nbgrader/exchange/abc/exchange.py", line 58, in fail
        raise ExchangeError(msg)
    nbgrader.exchange.abc.exchange.ExchangeError: Unwritable directory, please contact your instructor: /usr/local/share/nbgrader/exchange
```

And then there follows quite a few 404s / file not found errors:
```
[WebServer] [W 2023-08-21 11:34:22.071 ServerApp] wrote error: 'No such file or directory: nbgrader_config.py'
    Traceback (most recent call last):
      File "/opt/hostedtoolcache/Python/3.10.12/x64/lib/python3.10/site-packages/tornado/web.py", line 1786, in _execute
        result = await result
      File "/opt/hostedtoolcache/Python/3.10.12/x64/lib/python3.10/site-packages/jupyter_server/services/contents/handlers.py", line 266, in put
        await self._save(model, path)
[...]

PR #1817, opened 2023-08-23
New Feature: Auto-generation of test code (compatible with 0.9.x)

trevorcampbell:
This PR supercedes #1650 and is compatible with the upcoming 0.9.x release. We can close that one in favour of this new PR, but see the old discussion there and at #1633.

Some notes:
- I chose not to rebase the old PR because it was too far out of date.
- This new PR has much-improved kernel execution code for instantiating tests compared with the old PR.
- There are quite a few new tests implemented to validate AutoTest functionality. But note that AutoTest doesn't modify any UI elements or require any special treatment in the UI. So in the `ui-tests` I decided to just remove the new `problem3.ipynb` notebook from the `user_guide/source/ps1/` directory during `ui-tests` to avoid triggering AutoTest.
  - I originally tried to include this in the UI tests with the other notebooks, but I really struggled to debug the tests given all the pollution of the logs with existing errors (see [this discussion](https://github.com/jupyter/nbgrader/discussions/1749#discussioncomment-6786738) and [this issue](https://github.com/jupyter/nbgrader/issues/1816)).


nbgrader/apps/quickstartapp.py
brichet, line 1: Would it be possible to use the autotests only if a flag is provided (e.g. `nbgrader quickstart --autotests`) ?

Currently the quickstart generate *ps1/Problem3.ipynb*, which is a copy of the *ps1/Problem1.ipynb* with autotests. It can be confusing. We could generate only one of these example, depending on the flag above.
Same for the `tests.yml` file.

nbgrader/docs/source/user_guide/tests.yml
brichet, line 1: We should rename it `autotests.yml` for consistency.

nbgrader/preprocessors/clearsolutions.py
```diff
@@ -15,6 +15,7 @@ class ClearSolutions(NbGraderPreprocessor):
 
     code_stub = Dict(
         dict(python="# YOUR CODE HERE\nraise NotImplementedError()",
+             R="# YOUR CODE HERE\nfail()",
```
brichet, line 18: Is this related to this PR ?

nbgrader/preprocessors/instantiatetests.py
```diff
@@ -0,0 +1,420 @@
...
+        # get the setup code if it's there
+        self.setup_code = tests.get('setup', None)
+
+    # -------------------------------------------------------------------------------------
+    def _instantiate_tests(self, snippet, salt=None):
+        # get the type of the snippet output (used to dispatch autotest)
+        template = j2.Environment(loader=j2.BaseLoader).from_string(self.dispatch_template)
+        dispatch_code = template.render(snippet=snippet)
+        dispatch_result = self._execute_code_snippet(dispatch_code)
+        self.log.debug('Dispatch result returned by kernel: ', dispatch_result)
```
brichet, line 345: There is a warning about arguments:
```suggestion
        self.log.debug(f'Dispatch result returned by kernel: {dispatch_result}')
```

nbgrader/preprocessors/instantiatetests.py
```diff
@@ -0,0 +1,420 @@
...
+        tests = tests[resources["kernel_name"]]
+
+        # get the test templates
+        self.test_templates_by_type = tests['templates']
+
+        # get the test dispatch code template
+        self.dispatch_template = tests['dispatch']
+
+        # get the success message template
+        self.success_code = tests['success']
```
brichet, line 325: Shouldn't it be optional ?

nbgrader/preprocessors/instantiatetests.py
```diff
@@ -0,0 +1,420 @@
...
+                if kernel_name not in self.comment_strs:
+                    raise ValueError(f"Kernel {kernel_name} has not been specified in InstantiateTests.comment_strs")
+                if kernel_name not in self.sanitizers:
+                    raise ValueError(f"Kernel {kernel_name} has not been specified in InstantiateTests.sanitizers")
+                self.log.debug(f"Found kernel {kernel_name}")
+                resources["kernel_name"] = kernel_name
+
+                # load the template tests file
+                self.log.debug('Loading template tests file')
+                self._load_test_template_file(resources)
```
brichet, line 123: Since the autotests is optional, is it possible to avoid running the preprocessor if the config file does not exist ?

nbgrader/preprocessors/instantiatetests.py
```diff
@@ -0,0 +1,420 @@
...
+        self.test_templates_by_type = tests['templates']
+
+        # get the test dispatch code template
+        self.dispatch_template = tests['dispatch']
+
+        # get the success message template
+        self.success_code = tests['success']
+
+        # get the hash code template
+        self.hash_template = tests['hash']
```
brichet, line 328: I think it should be optional, and should generate an error if `HASHED` is required and this not filled.

nbgrader/docs/source/user_guide/tests.yml
```diff
@@ -0,0 +1,305 @@
+python3:
+    setup: "from hashlib import sha1"
+    hash: 'sha1({{snippet}}.encode("utf-8")+b"{{salt}}").hexdigest()'
```
brichet, line 3: Don't you think this could be something like:
```suggestion
    hash:
        setup: "from hashlib import sha1"
        command: 'sha1({{snippet}}.encode("utf-8")+b"{{salt}}").hexdigest()'
```
to have a special setup only for hash, keeping also a general setup for other configuration than hash.
This would allow importing the hash function only if hash is required.

brichet (commented):
Thanks @trevorcampbell for updating this PR.

In a general case, can it (and should it) be combined with `HIDDEN TESTS`, to prevent students from knowing the test cases ?

I made some comments/suggestions below. I did not check at the tests you added yet.

nbgrader/preprocessors/instantiatetests.py
trevorcampbell: Agreed -- I'll make the change

nbgrader/preprocessors/instantiatetests.py
trevorcampbell: Ah yes -- this is related to a new change I made compared with the old PR. I call `_load_test_template_file` in `preprocess` now, but only after checking if `self.autotest_delimiter in cell.source`. So indeed we can catch an error sooner.

(But we should raise an exception here. If there is a `self.autotest_delimiter` in a cell, it's clear the instructor wants autotest to run -- moving on silently without running autotest might create notebooks without tests the instructor expects to be there.)

nbgrader/preprocessors/clearsolutions.py
trevorcampbell: Yes, it's there because `ps2/problem.ipynb` needs it currently.

I can separate all the R stuff into its own PR if you want?

nbgrader/docs/source/user_guide/source/ps2/problem.ipynb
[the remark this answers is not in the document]
trevorcampbell: I'm happy to split the R stuff into its own PR. I'll do that

nbgrader/apps/quickstartapp.py
trevorcampbell: > Currently the quickstart generate ps1/Problem3.ipynb, which is a copy of the ps1/Problem1.ipynb with autotests. It can be confusing. 

I agree. But the point of this example was to show something identical to `problem1.ipynb` except implemented using Autotests (so instructors can compare the manual one to the automatic one). But having it named `problem3.ipynb` is indeed confusing.

Maybe I can create `ps1_autotests/` as a new assignment entirely, and make it clear in the documentation that it's a duplicated assignment except implemented using automatic tests?

> Would it be possible to use the autotests only if a flag is provided (e.g. nbgrader quickstart --autotests) ?

I suspect if we make it clearer that it's meant to duplicate `ps1` but show off autotest functionality, we don't need a separate flag for it.

nbgrader/docs/source/user_guide/tests.yml
trevorcampbell: I personally don't think it's worth the change--the `setup` is general anyway, it can be used for whatever. I don't feel strongly, though, so if you feel strongly I can change it. Just let me know

nbgrader/preprocessors/instantiatetests.py
trevorcampbell: Ah, actually I'll use something like the below for consistency with the rest of the code:

```
./nbgrader/converters/base.py:        self.log.debug("Student: %s", gd['student_id'])
```

nbgrader/preprocessors/clearsolutions.py
trevorcampbell: (But really this is a fairly small improvement, and I'm not sure why R wasn't listed here in the first place)

trevorcampbell:
Thanks for the comments @brichet !

Re: hidden tests -- yes, autotest works as expected with hidden tests. In `generate_assignment`, the `InstantiateTests` preprocessor runs before `ClearHiddenTests` -- once the tests are instantiated, the notebook is processed just like any other notebook. 

So far I have done the following, per your request:
- renamed `tests.yml` to `autotests.yml`
- made `hash` and `success` optional
- avoid running the cell preprocessor if an `autotest_delimiter` is found but `autotests.yml` cannot be loaded. This raises an exception to avoid silently missing tests that instructors intended to insert.
- fixed warning message string interpolation for logging messages

Two remaining things:
- Regarding `problem3.ipynb`: see my comment above -- the intention was to show users a duplicate assignment implemented using autotests (so they can compare them). Would it work for you if I made a new assignment called `ps1_autotests` (I'm not sure I like that folder name, but something like that) instead of adding an `--autotests` flag?
- Regarding R: We need both R and Py in our course at UBC, so I would like to keep the R stuff if possible. If not I can separate `ps2` into its own PR, but I still think we should keep the `code_stub` change at the very least.

Let me know what you think about these two, and I'll get that sorted as quickly as possible.

nbgrader/preprocessors/instantiatetests.py
brichet: I missed the check about `self.autotest_delimiter in cell.source`. Anyway, it's probably better to raise if the file doesn't exist, rather than warning of something that will happen anyway (since *AUTOTEST* has already been found in the file).

brichet:
> * Regarding R: We need both R and Py in our course at UBC, so I would like to keep the R stuff if possible. If not I can separate `ps2` into its own PR, but I still think we should keep the `code_stub` change at the very least.

I agree about keeping the `code_stub`.
About the courses in Python and R together, it's related to the use case of UBC. 
I'm more in favor of a new PR to select the language(s) in the `quickstart` using a flag. It could also generate multiple assignments using different languages, using multiple flags.

nbgrader/apps/quickstartapp.py
brichet: I like the idea of the `quickstart` generating an assignment as little as possible.
My opinion is that *autotests* is an advanced feature, and bringing a new configuration file (*autotest.yml*) in the quickstart can be confusing for people who just try to understand the basics of `nbgrader`.

cc. @jhamrick @BertR @perllaghu if you have an opinion on it

nbgrader/tests/apps/files/autotest-multi-unchanged.ipynb
brichet, line 1: Can you reset the outputs and execution count ?

brichet (commented):
Thanks for your quick update @trevorcampbell.
I added a few comments about syntax.

I didn't find documentation about the YAML file ? 
Can you add some (if missing), maybe in *advanced topic*, with a link to it in *“Autograder tests” cells with automatically generated test code*.

nbgrader/apps/quickstartapp.py
trevorcampbell: @brichet I'm happy to do as you suggest -- you're right, the quickstart should produce a very simple example to start from, and the flags can be used for advanced users.

trevorcampbell:
OK @brichet :
- I wrote a new documentation section in the *Advanced Topics* area that details `autotests.yml` and more advanced parts of autotest in general. Cross linked with the initial section in the *Creating and Grading Assignments*. 
- The quickstart app now accepts an `--autotest` flag. Without the flag, it reproduces only the old `ps1` problem set. With the flag, it produces the same looking problem set, except `ps1/problem1.ipynb` has autotests in it. There is no `ps2` any more.
- added tests for the `--autotest` flag in the quickstart tests file
- The R `ps2` assignment is now removed from the `source/` in docs
- Added `Jinja2 >= 3.0` to `pyproject.toml` (it was missing before -- autotest needs jinja for template substitution)

I think that was everything from your suggestions. Let me know what you think!

pyproject.toml
```diff
@@ -47,6 +47,8 @@ dependencies = [
     "rapidfuzz>=1.8",
     "requests>=2.26",
     "sqlalchemy>=1.4,<3",
+    "PyYAML>=6.0",
+    "Jinja2>=3.0"
```
brichet, line 51: `Jinja2>=3` is already imported above.

```suggestion
```

brichet (commented):
Thanks again @trevorcampbell for being so responsive.
The documentation looks great, and I like the `--autotest` option in the quickstart.

Let me know if you think it's ready to be merged, it looks good to me.

trevorcampbell:
Thanks @brichet -- FYI: I added two more tests (just to catch the kernel working directory behaviour and release notebook consistency).

I think it's ready to merge now.

PR #1855, opened 2024-01-03
Remove use of deprecated typing.io

shreve:
The typing.io namespace was deprecated in python 3.8, which is the lowest version currently supported by nbgrader.

https://docs.python.org/3/library/typing.html#aliases-to-other-concrete-types


brichet (approved):
LGTM, thanks @shreve

PR #1871, opened 2024-03-26
Fix the confusing server errors during UI tests

brichet:
This PR addresses https://github.com/jupyter/nbgrader/issues/1816.

It suppresses server error messages during user interface tests, which suggest that the test is not passing, and make the stdout output difficult to analyze.


brichet:
Yes, some traceback are expected (test with missing file or directory).

PR #1874, opened 2024-03-28
Update Release Scripts

blink1073:
See https://github.com/jupyter-server/jupyter_releaser/pull/557 for context.

### DIFF
--- a/.github/workflows/prep-release.yml
+++ b/.github/workflows/prep-release.yml
@@ -1,0 +1,49 @@
+name: "Step 1: Prep Release"
+on:
+  workflow_dispatch:
+    inputs:
+      version_spec:
+        description: "New Version Specifier"
+        default: "next"
+        required: false
+      branch:
+        description: "The branch to target"
+        required: false
+      post_version_spec:
+        description: "Post Version Specifier"
+        required: false
+      silent:
+        description: "Set a placeholder in the changelog and don't publish the release."
+        required: false
+        type: boolean
+      since:
+        description: "Use PRs with activity since this date or git reference"
+        required: false
+      since_last_stable:
+        description: "Use PRs with activity since the last stable git tag"
+        required: false
+        type: boolean
+jobs:
+  prep_release:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: jupyterlab/maintainer-tools/.github/actions/base-setup@v1
+
+      - name: Prep Release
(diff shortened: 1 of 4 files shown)


blink1073:
@brichet can you please add `blink1073` to the PyPI project so I can set up trusted publishing?

blink1073:
Also, you'll need to create an npm token and replace the NPM_TOKEN secret.

brichet:
Thanks @blink1073 for quick update.

Look like I can't add you on PyPI


Maybe @jtpio  or @SylvainCorlay can ? And if you can please also add me as owner, it would be great.

brichet:
> Also, you'll need to create an npm token and replace the NPM_TOKEN secret.

Not sure to understand that one, the current NPM_TOKEN secret is still valid.
Do U have to change something when created it again ?

blink1073:
> Not sure to understand that one, the current NPM_TOKEN secret is still valid.

There wasn't one set, I just set it to the one for the `@jupyter` org, but this package isn't in that namespace.  I don't have access to the npm package.

brichet:
> > Not sure to understand that one, the current NPM_TOKEN secret is still valid.
> 
> There wasn't one set, I just set it to the one for the `@jupyter` org, but this package isn't in that namespace. I don't have access to the npm package.

Strange, I used the releaser to push on NPM 2 days ago, using a token that I created
Screenshot from 2024-03-28 15-55-44

No matter, I updated the value of the NPM_TOKEN.

Perhaps we should move the package to the `@jupyter` namespace later. We could then use the token from the `@jupyter` org, no ?

blink1073:
> Perhaps we should move the package to the @jupyter namespace later. We could then use the token from the @jupyter org, no ?

Yes, all you'd have to do is rename it in `package.json` and ping me to update the token.

blink1073:
Trusted publishing is set up, ready to go!

brichet:
Thanks a lot @blink1073

blink1073:
No problem!

brichet:
> > Perhaps we should move the package to the @jupyter namespace later. We could then use the token from the @jupyter org, no ?
> 
> Yes, all you'd have to do is rename it in `package.json` and ping me to update the token.

Do you know if it will be possible to bring the history back to the new package ?

blink1073:
> Do you know if it will be possible to bring the history back to the new package ?

I don't know, I haven't dealt with npm packages directly in a few years.

PR #1902, opened 2024-07-12
Eliminate compatibility issues in supporting notebook and jupyterlab

### LINKED ISSUE #1888: Click on the course name in formgrader but does nothing
### Operating system
       ubuntu 22 
### `nbgrader --version`

# nbgrader --version
Python version 3.9.17 (main, Jul  5 2023, 20:41:20) 
[GCC 11.2.0]
nbgrader version 0.8.5
# conda list |grep nbgrader
nbgrader                  0.8.5              pyhd8ed1ab_0    conda-forge

### `jupyterhub --version` (if used with JupyterHub)

root@jpserver:/usr/local/etc# conda list |grep nbgrader
nbgrader                  0.8.5              pyhd8ed1ab_0    conda-forge
root@jpserver:/usr/local/etc# jupyterhub --version
4.1.5
root@jpserver:/usr/local/etc# conda list |grep  jupyterhub
jupyterhub                4.1.5              pyh31011fe_0    conda-forge
jupyterhub-base           4.1.5              pyh31011fe_0    conda-forge

root@jpserver:/usr/local/etc# jupyter notebook --version
6.5.6
root@jpserver:/usr/local/etc# conda list |grep notebook
notebook                  6.5.6              pyha770c72_0    conda-forge
notebook-shim             0.2.4              pyhd8ed1ab_0    conda-forge

### Expected behavior
A notebook opens upon clicking onto the name of an assignment. (formgrader page and there managing assignments)
i login jupyterhub, switch to notebook, then open page of Manage Assignments , when i click on the course name in formgrader , then should open assignment folder

### Actual behavior

I login jupyterhub, switch to notebook, then open page of Manage Assignments , when i click on the course name in formgrader but does nothing 

Have JupyterHub with nbgrader and one classes of ps1 setup.
Go to formgrader page and there managing assignments and add an assignment or try to use the preexisting ps1 that comes with the quickstart (following the demo-setup-steps for multiple classes).

issue

shreve:
Formgrader could be loaded in one of two different contexts: on its own from a notebook, or inside an iframe a lab. When navigating to a page outside the formgrader application, this presents an issue in that these two contexts need different navigation mechanisms.

Previously, the mechanism was decided by looking at the name of the jupyter-server application. If it was `jupyter-notebook` (notebook<7), use href. Otherwise, use postMessage. However, it's possible to use the notebook interface without `jupyter-notebook`, e.g. `jupyter-server`.

Now the mechanism is decided by context clues inside the browser, i.e. whether formgrader is inside an iframe.

This commit adds a `linkTo` function which is applied to jQuery elements via `.map` and applies either href or postMessage nav mechanisms based on whether the current window is the top window.

Additionally, it completely removes the `is_lab` and `is_jlab` checks as they were relatively naive and add complexity that doesn't need to be there.

Fixes #1888


nbgrader/server_extensions/formgrader/base.py
```diff
@@ -41,7 +42,10 @@ def gradebook(self):
 
     @property
     def mathjax_url(self):
-        return self.settings['mathjax_url']
+        url = self.settings.get("mathjax_url", "")
+        if not url or url_is_absolute(url):
+            return url
+        return url_path_join(self.base_url or "/", url)
```
shreve, line 48: This method exists in the parent `JupyterHandler` and could be removed here completely if base_url didn't include `.rstrip("/")`. 

https://github.com/jupyter-server/jupyter_server/blob/5fa7ceb3f1a5b103b9eea1231480ebddd9ae76fb/jupyter_server/base/handlers.py#L248-L261

shreve:
The two "Check Release" failures and the "Make SDist" failure are related to Automattic/node-canvas. I think this might need a yarn.lock bump, but I'm not sure how safe that is.

The test_nbgrader(docs) failure is a timeout because checking GitHub links in the changelog keeps getting rate limited and eventually times out. It seems like it might be worth adding a filter which auto-approves links that start with github.com/search.

brichet (approved):
Thanks @shreve, it looks good to me.

Do you think that you can add a test on it ? Otherwise I can do it in a follow up PR.

nbgrader/server_extensions/formgrader/base.py
brichet: Do you think that `rstrip` is necessary ? Does it break something if we remove it ?

shreve: The `rstrip` was added really early in development, so the whole server extension is built based on it. I don't think it will be difficult to change, but every usage of `{{ base_url }}/` in the templates would need to be changed to `{{ base_url }}`, and similar updates need to be made in the javascript too.

PR #1904, opened 2024-07-16
New Autograde Preprocessor: IgnorePattern

AlirezaT99:
Creating a new preprocessor, which is useful for removing specific warning cells that might result in a false-negative autograde result. This is used when the assignment is already out and released, but you notice a problem in a test after the fact. This provides a much easier way of recovering compared to editing the assignment.

We at @aaltoscicomp came across an issue where Torch raised a specific warning on stderr (due to some avx1 incompatibility in the node on which the grading would happen), and the auto-grader would mark the solution as wrong. Since the warning was not a Python Warning object (NNPACK.cpp warning), it cannot be suppressed the usual way. Since this warning can't be easily predicted in advance, the assignment couldn't be made to compensate for it.

We addressed the problem by inserting this preprocessor in the `c.Autograde.autograde_preprocessors` list immediately after `Execute` to detect and remove pre-defined warning (stderr) outputs from GradeCells. The module uses a configured regular expression (`c.IgnorePattern.pattern`) to find and delete the matching outputs before the grading process.  The default value is empty, which short-circuits the processing so has no effect (and anyway an empty value would have no effect by default).

It shouldn't be needed often, but when it's needed it will save a lot of time.

It has tests and all. We haven't used it live since we made it too late, but we definitely had a need for it.

### DIFF
--- a/nbgrader/converters/autograde.py
+++ b/nbgrader/converters/autograde.py
@@ -9,7 +9,7 @@ from nbconvert.preprocessors import ClearMetadataPreprocessor
 from .base import BaseConverter, NbGraderException
 from ..preprocessors import (
     AssignLatePenalties, ClearOutput, DeduplicateIds, OverwriteCells, SaveAutoGrades,
-    Execute, LimitOutput, OverwriteKernelspec, CheckCellMetadata)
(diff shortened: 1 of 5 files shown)


brichet (approved):
Thanks @AlirezaT99, LGTM

brichet:
@meeseeksdev backport to 0.8.x

Issue #660, opened 2017-01-15
Add `grant_extension` method to Gradebook


jhamrick:
There is currently support for giving students extensions (see http://nbgrader.readthedocs.io/en/stable/api/models.html#nbgrader.api.SubmittedAssignment.extension ) but it's not particularly discoverable in the documentation nor particularly easy to use. There should be a method added to the Gradebook class that allows instructors to more easily grant extensions, and we should create better documentation for it.

nhiquach:
What is the current procedure for modifying the extension value of a submitted assignment? I'm having some trouble finding the interface to modify this attribute in the db.

PR #1906, opened 2024-07-29
Grant Extension interface in the ManageSubmissions section

AlirezaT99:
Long ago, the `grant_extension` method was added to the gradebook following #660—however, no API handler or user interface utilized this functionality.

In this PR, a new modal has been introduced to the “Manage Submissions” view in FormGrader, allowing viewing and granting individual extensions to auto-graded submissions.

image

As mentioned, the modal opens only for the auto-graded submissions. This limitation is an attempt to minimize the changes to the existing API.

image


Gehock:
Rebased on top of current main branch

Gehock:
Fixed an oversight in tests, now passes locally at least

PR #1909, opened 2024-07-30
Show Deadlines in Assignment List

### LINKED ISSUE #943: Ability for students to see assignment duedate
It seems that the current Student use case of `nbgrader list` is fairly limited in what information can be seen. Is there any way to expose the duedate of an assignment on the student side of nbgrader?

AlirezaT99:
In this PR, we introduce a workaround for storing and displaying deadlines to students, which fixes #943.

Currently, the due date is stored in the Gradebook and inaccessible in the student's view. To tackle this issue, we store the deadlines in a file in the *exchange* (fs-based) at a configured path as well. This allows us to display the deadlines in a new column within the AssignmentList. The solution respects the existing configurations for time zones and group sharing.

new-assignment-list-look

Moreover, the Submit button now triggers a dialog that asks for confirmation in the case of a late submission.

late-submission-warning-dialog

**Note** that the deadline file manipulation is attached to API so it will only be set for assignments created/editted after the upgrade. Assignments released before the update won't have deadline files made, and won't be shown to the users (which is fixed by simply pressing the edit button with no changes for each).

According to the feedback received from instructors who use nbgrader, this will enhance the assignment submission and collection processes, and reduce the need for manual intervention.

### DIFF
--- a/nbgrader/coursedir.py
+++ b/nbgrader/coursedir.py
@@ -225,6 +225,11 @@ class CourseDirectory(LoggingConfigurable):
         )
     ).tag(config=True)
 
+    deadline_file = Unicode(
+        "deadlines.json",
+        help="File for storing deadlines"
+    ).tag(config=True)
+
     groupshared = Bool(
         False,
         help=dedent(
--- a/nbgrader/exchange/abc/exchange.py
+++ b/nbgrader/exchange/abc/exchange.py
@@ -36,11 +36,6 @@ class Exchange(LoggingConfigurable):
         help="Format string for timestamps"
     ).tag(config=True)
 
-    deadline_file = Unicode(
-        "deadlines.txt",
-        help="File for storing deadlines"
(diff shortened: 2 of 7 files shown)


lahwaacz:
@AlirezaT99 Thanks for working on this! :tada:  It looks good to me.

brichet (approved):
Thanks @AlirezaT99, and sorry for the very late review.

Maybe, for a follow PR, we could also display that the assignment has been submitted late in formgrader, for ease instructor check ? 

Image

Otherwise it looks good to me, can you merge main in it to ensure it does not break anything (looks like I don't have permission to do it) ?

AlirezaT99:
Thanks @brichet for the review!
It might take me a while to set up the codebase as I no longer work on this. Can you @Gehock kindly take care of this?

Gehock:
I'll have a look

Issue #1870, opened 2024-03-20
JupyterHub 4.1.0 breaks Formgrader iFrame embed


milandeepbassi:
### Operating system 
Ubuntu 22

### `nbgrader --version`
0.9.1

### `jupyterhub --version` (if used with JupyterHub) 
4.1.0

### `jupyter notebook --version`  
7.1.2

### Expected behavior

![image](https://github.com/jupyter/nbgrader/assets/59866350/46e632b3-86d9-45f3-af7d-a1221955ac69)

### Actual behavior

![image](https://github.com/jupyter/nbgrader/assets/59866350/373e7952-0575-44c9-b443-27466246f088)

### Steps to reproduce the behavior

Created an instance with the The Littlest Jupyterhub that fetches JupyterHub 4.1.0 and installed NBGrader. Tried to access Formgrader and get the following error on browsers.

It is due to this update on Jupyterhub here: https://jupyterhub.readthedocs.io/en/stable/explanation/websecurity.html#mitigating-same-origin-deployments

```
The default Content-Security-Policy for single-user servers is

frame-ancestors: 'none'
which prohibits iframe embedding, but not pop-ups.
```

lahwaacz:
Confirmed!

Zercerium:
workaround if you are using jupyterhub in Kubernetes:
edit the helm chart and add the key extraFiles with the following content: pls replace the domain with your domain
```
singleuser:
  [...]
  extraFiles:
    server-config:
      mountPath: /etc/jupyter/jupyter_server_config.json
      data: {
        ServerApp: {
          tornado_settings: {'headers': {'Content-Security-Policy': "frame-ancestors self https://example.com;"}}
        }
      }
```

lahwaacz:
You can have just `"frame-ancestors self"` without having to specify a domain. However, note that this obviously impacts security of the whole JupyterHub.

goekce:
I would like to use the workaround until the problem is fixed in nbgrader. Jupyterhub is installed directly on my server OS and I modified `jupyterhub_config.py` as follows:

```python
c.JupyterHub.tornado_settings = {                                                                                                                                                                                                            
        "headers":{"Content-Security-Policy": "frame-ancestors 'self' "}                                                                                                                                                                     
}                                                                                                                                                                                                                                            
c.JupyterHub.ServerApp.tornado_settings = {                                                                                                                                                                                                  
        "headers":{"Content-Security-Policy": "frame-ancestors 'self' "}
```
[...]

PR #1915, opened 2024-08-26
Set CSP frame-ancestors 'self' for nbgrader handlers

jeflem:
This PR solves issue #1870 (formgrader does not show in JupyerLab tab due to JupyterHub >=4.1.0 security settings in HTTP headers).

Starting with JupyterHub 4.1.0 HTTP header
```
Content-Security-Policy:  frame-ancestors 'none'
```
is the default setting instead of
```
Content-Security-Policy: frame-ancestors 'self'
```
See [Mitigating same-origin deployments](https://jupyterhub.readthedocs.io/en/stable/explanation/websecurity.html#mitigating-same-origin-deployments) for some background on this decision and [CSP: frame-ancestors](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors#self) for details on the header.

The `none` header prevents loading of formgrader in a tab of JupyterLab.

The JupyterHub `none` setting overwrites the `self` setting of Jupyter Server running without JupyterHub. See
* [jupyter_server/base/handlers.py#L101](https://github.com/jupyter-server/jupyter_server/blob/74655ce66f36ed85a83591e6658e70ba91232580/jupyter_server/base/handlers.py#L101) for default setting `self` in Jupyter Server,
* [jupyterhub/singleuser/extension.py#L658](https://github.com/jupyterhub/jupyterhub/blob/99b32dd372ae41c14473e543be4c302b80f613aa/jupyterhub/singleuser/extension.py#L658) for default setting `none` in JupyterHub's Jupyter Server extension,
* [jupyterhub/singleuser/mixins.py#L688](https://github.com/jupyterhub/jupyterhub/blob/99b32dd372ae41c14473e543be4c302b80f613aa/jupyterhub/singleuser/mixins.py#L688) for default setting `none` in JupyterHub's code for classic notebook.

To allow embedding of nbgrader's formgrader (and possibly other nbgrader components) without affecting security of other JupyterHub components this PR sets `frame-ancestors` to `self` for responses of nbgrader handlers only.

The class `BaseHandler` modified by this PR is a subclass of [`tornado.web.RequestHandler`](https://www.tornadoweb.org/en/stable/web.html#tornado.web.RequestHandler), which provides the [`set_header`](https://www.tornadoweb.org/en/stable/web.html#tornado.web.RequestHandler.set_header) method.

### DIFF
--- a/nbgrader/server_extensions/formgrader/base.py
+++ b/nbgrader/server_extensions/formgrader/base.py
@@ -60,6 +60,10 @@ class BaseHandler(JupyterHandler):
         api.log_level = level
         return api
 
+    def initialize(self):
+        super().initialize()
+        self.set_header("Content-Security-Policy", "frame-ancestors 'self'")
+
     def render(self, name, **ns):
         template = self.settings['nbgrader_jinja2_env'].get_template(name)
         return template.render(**ns)


brichet:
@jeflem I took the liberty to merge main is this PR to try to fix the `check_release` test, as mentioned at https://github.com/jupyter-server/jupyter_releaser/issues/593#issuecomment-2478628377

LukasMueller187:
Hi,
just wanted to check in here because we do have problems with clients using iPads (browser doesn't matter). Feedback html files are rejected because of the frame-ancestors directive.
I thought 0.9.4 would fix it but it doesn't.
Manually setting the header to 'self' again in (our case) apache does solve the issue, but as stated in the first post this is discouraged.

Any ideas?

brichet:
Is this only failing on IPads ?

> Manually setting the header to 'self' again in (our case) apache does solve the issue, but as stated in the first post this is discouraged.

What do you mean by manually ? in the config ?
What is the "first post" you are mentioning ?

LukasMueller187:
Yes it seems to only fail on iPads (and maybe iPhones), so I'm guessing it's an iOS issue.
I narrowed it down to being an issue with html files, as they are always served as a blob. As stated in the error message, iOS seems to reject this blob. In my understanding, it shouldn't (like all desktop browsers do).

By manually I mean configuring apache to set headers to 'self' for frame-ancestors and by first post I mean the first post in this issue.

However, this doesn't seem to be an issue with nbgrader. But I don't know where to file the bug (Lab or Hub or webkit iOS itself).

brichet:
Thanks for the clarification

PR #1934, opened 2024-11-05
Submission scores in Assigment List

KrKOo:
With this PR, the scores for each assignment are displayed directly on the assignment list page, significantly enhancing the course overview from the student's perspective.

- **Updated feedback template**
The only source of scores accessible to students is the feedback.html file. To facilitate easier parsing, the default template has been modified by adding <meta> tags to store the student's earned score. This change also allows custom templates to be parsed conveniently, provided they incorporate these meta tags.

- **Updated Exchange**
The default exchange now parses the feedback file and retrieves the score and max_score attributes for each submission. If multiple feedbacks are available, the score from the last one is returned.

- **Backward Compatibility**
  - Since this update relies on the new <meta> tags in feedback files, scores will not be displayed for older assignments.
  - If a custom exchange does not return scores, they will not appear in the UI.

image

### DIFF
--- a/nbgrader/exchange/default/list.py
+++ b/nbgrader/exchange/default/list.py
@@ -3,6 +3,7 @@ import glob
 import shutil
 import re
 import hashlib
+from lxml import html
 
 from nbgrader.exchange.abc import ExchangeList as ABCExchangeList
 from nbgrader.utils import notebook_hash, make_unique_key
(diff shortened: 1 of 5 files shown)


lahwaacz:
Why do you need to contact the exchange to get the score for display? When the feedback is already downloaded by the user, the `assignment_list` JupyterLab extension can just contact its Jupyter server counterpart that would extract the values from the downloaded file. This would avoid the need to upgrade any external exchange code to support the new functionality.

PR #1938, opened 2024-11-15
Prevent dependabot from using yarn 4

brichet:
This PR should prevent dependabot from using yarn 4, and change the whole dependencies in `yarn.lock`.

See https://github.com/jupyterlab/extension-template/pull/79 for more information.


brichet:
Failure is not related

PR #1940, opened 2024-11-15
Fix the mit license link

brichet:


lahwaacz:
The [LICENSE file](https://github.com/jupyter/nbgrader/blob/main/LICENSE) is BSD-3-Clause, not MIT, though...

lahwaacz:
> It is the license of a component used in formgrader.

Hmm what is the upstream of the component? Why do you update in this git repo? The top-level LICENSE file should still apply to the whole repo, but in this case it doesn't - a [classical issue of a monorepo](https://css-tricks.com/from-a-single-repo-to-multi-repos-to-monorepo-to-multi-monorepo/#aa-issues-with-the-monorepo)...

brichet:
I agree, the choice of embedding javascript libraries seems to date back to the origins. Some notes about it in the doc: https://nbgrader.readthedocs.io/en/latest/contributor_guide/js_dependencies.html

We should probably get rid of these committed dependencies. But we should also avoid using `bower` as package management for the formgrader frontend, it is now unmaintained.

I'm closing this PR since it doesn't solve the problem anyway.

PR #1942, opened 2024-11-18
Replace npm-run-all by npm-run-all2

brichet:
The current version of `npm-run-all` block updating `cross-spawn`.

This leads to a security issue https://github.com/jupyter/nbgrader/security/dependabot/52.

looks like this package is not necessary, this PR removes it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,6 +76,7 @@
     "eslint-config-prettier": "^8.0.0",
     "eslint-plugin-prettier": "^4.0.0",
     "mkdirp": "^1.0.4",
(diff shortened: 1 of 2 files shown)


brichet:
Looks like it is used in https://github.com/jupyter/nbgrader/blob/1836e322057bd8f46f1e2818d4b4ec24a2c3dd26/package.json#L41 with the `run-p` CLI.

This [fork](https://github.com/bcomnes/npm-run-all2) has updated the dependencies, the initial project doesn't seems to be maintained anymore.

PR #1947, opened 2025-01-17
Revert #1915 for security reason

brichet:
This PR reverts #1915 for security reason.

Jupyter application running in JupyterHub must not allow iframes, as described at [mitigating-same-origin-deployments](https://jupyterhub.readthedocs.io/en/stable/explanation/websecurity.html#mitigating-same-origin-deployments).

This PR also updates the demo and documentation to make nbgrader work with `jupyterhub>=4.1`.


lahwaacz:
So nbgrader should be revamped to avoid using iframes altogether...

brichet:
> So nbgrader should be revamped to avoid using iframes altogether...

AFAIK only the `formgrader` UI is using an iframe, and there are solutions to use it with JupyterHub (see updated [doc](https://nbgrader--1947.org.readthedocs.build/en/1947/configuration/jupyterhub_config.html#formgrader-with-jupyterhub-4-1)).

Revamping the formgrader UI would be the best solution indeed, but it would take some time.

nbgrader/docs/source/configuration/jupyterhub_config.rst
```diff
@@ -331,3 +345,55 @@ API
...
+
+Although this solution isn't the most practical, it does allow to use ```formgrader``
+without having to update the configuration and without adding vulnerabilities to the application.
+
+Enabling JupyterHub subdomains
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Enabling per-user and per-service subdomains with ``JupyterHub.enable_subdomains = True``
+allows to securely use iframes with JupyterHub.
+Then, even if embedding in an IFrame is allowed, the host page does not have access to the
```
minrk, line 378: This second sentence isn't quite accurate - `frame-ancestors: self` allows same-origin iframes which _do_ still have access to the page contents. That's not strictly the problem. The important change is that when `enable_subdomains` is True,  `frame-ancestors self` allows embedding the iframe _only_ on pages served by the user's own server but not on _other_ users' servers, which is the problem case for a single-origin JupyterHub deployment.

minrk (approved):
Thanks! One suggestion to improve the accuracy of the docs a bit.

brichet:
Thanks again @minrk for the advisory, the possible solutions and the review.

PR #1962, opened 2025-04-28
fix: Allow null kernelspec in OverwriteKernelspec

shreve:
The `Notebook.kernelspec` field is nullable, so in the event of a null value, OverwriteKernelspec should not fail. This commit checks for that case before trying to `json.loads` the value.


brichet (approved):
Thanks @shreve

PR #1963, opened 2025-04-29
Resolve warnings in SQLAlchemy field defs

### LINKED ISSUE #1946: Some code uses constructs that will fail under Python 3.12
I'm testing a plugin for NBGrader, and getting the following errors [backslash wrapping by me]:

```
..../python3.11/site-packages/nbgrader/nbgraderformat/v1.py:119: DeprecationWarning: typing.io is deprecated, \
import directly from typing instead. typing.io will be removed in Python 3.12.
    def read_v1(source: typing.io.TextIO, as_version: int, **kwargs: typing.Any) -> NotebookNode:
```

and

```
..../python3.11/site-packages/nbgrader/api.py:1013: SADeprecationWarning: User-placed attribute Grade.cell_type \
on Mapper[Grade(grade)] is replacing an existing class-bound attribute of the same name.  Behavior is not fully \
defined in this case.  This use is deprecated and will raise an error in a future release
    Grade.cell_type = column_property(
```

shreve:
Most notably, Grade.cell_type was defined as an association_proxy in the class, but was later overwritten with a column_property. Because the mapper had already been made aware of the association_proxy, this raised a warning. Grade.max_score also had a similar problem, with a column_property overwriting a regular @​property.

This commit fixes the warnings by replacing the in-class field definitions with `field = None`.

Additionally, remove duplicate definitions for `cell_type_gradecell` and `cell_type_taskcell`.

Fixes #1946

### DIFF
--- a/nbgrader/api.py
+++ b/nbgrader/api.py
@@ -766,8 +766,8 @@ class Grade(Base):
     cell_id = Column(String(32), ForeignKey('base_cell.id'))
 
     #: The type of cell this grade corresponds to, inherited from
-    #: :class:`~nbgrader.api.GradeCell`
-    cell_type = association_proxy('cell', 'cell_type')
+    #: :class:`~nbgrader.api.GradeCell` or :class:`~nbgrader.api.TaskCell`
+    cell_type = None
 
     #: The student who this grade is assigned to, represented by a
     #: :class:`~nbgrader.api.Student` object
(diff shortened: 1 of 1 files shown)


brichet:
Thanks @shreve for working on this.

As far as I understand: 
- `cell_type` and `max_score` attributes were always overwritten
- `cell_type_from_taskcell` and `cell_type_from_gradecell` were similar to `cell_type_taskcell` and `cell_type_gradecell`, respectively, and used only to set `cell_type`

If so, your change looks good to me.

shreve:
That is my understanding as well. 

- `cell_type` and `max_score` were both always overwriting the already-mapped methods/attrs in the classes, and this was detected by SQLAlchemy and firing off a soon-to-be-an-error warning.
- `cell_type_from_gradecell` and `cell_type_from_taskcell` were exactly the same code as their non-`from` counterparts.

shreve:
I guess this is more maintenance than bugfix because it's cleanup and removal of a deprecation warning.

PR #1964, opened 2025-04-29
fix: Use cross-platform shortcuts in ui-tests

shreve:
Previously, UI tests failed on macOS because the required keyboard shortcuts are slightly different on macOS from Windows and Linux. This commit updates them to use the ControlOrMeta key, which handles both cases.


brichet (approved):
Thanks @shreve, it looks good to me.

PR #1967, opened 2025-05-09
Use NPM packages instead of bower for formgrader extension dependencies

### LINKED ISSUE #1966: Do not depend on bower for frontend packages
The formgrader extension relies on `bower` for the frontend dependencies that is serves.

Bower is deprecated and should not be used anymore, we can't update the dependencies.
This leads to security issues like https://github.com/jupyter/nbgrader/security/dependabot/60.

The 2 ways I can see to not use bower anymore are:

1. Using CDN for the frontend packages
    This would easily solve the problem as all the packages seems to be available over CDN.
    The drawback is that we need network access to run the application, and also to properly display the feedbacks
    https://github.com/jupyter/nbgrader/blob/99e145bf019d841a12a1b794e4bc29689ca267b3/nbgrader/server_extensions/formgrader/templates/feedback/index.html.j2#L11

    Reference to https://github.com/jupyter/nbgrader/issues/718#issuecomment-314077192

2. Using NPM to download the dependencies in the static directory.
    Then the `*node_modules* directory would replace the current *components* directory, and could be included with the formgrader extension.
    That way it would be the same behavior as currently.

I would go for the 2nd way to reduce changes.

References:
- https://github.com/jupyter/nbgrader/issues/1348
- https://github.com/jupyter/nbgrader/pull/1350

brichet:
Fixes #1966 
Should fix https://github.com/jupyter/nbgrader/security/dependabot/60

This PR replaces the use of deprecated `bower` by NPM.
It also updates `underscore` dependency to >=1.13.1, which should fix a security error.


.gitignore
```diff
@@ -96,20 +96,25 @@ nbgrader/docs/source/user_guide/source/ps1_autotest/problem1.html
...
+nbgrader/server_extensions/formgrader/static/node_modules/bootstrap/js
+nbgrader/server_extensions/formgrader/static/node_modules/bootstrap/grunt
+nbgrader/server_extensions/formgrader/static/node_modules/jquery/external
+nbgrader/server_extensions/formgrader/static/node_modules/jquery/src
+nbgrader/server_extensions/formgrader/static/node_modules/underscore/amd
+nbgrader/server_extensions/formgrader/static/node_modules/underscore/cjs
+nbgrader/server_extensions/formgrader/static/node_modules/underscore/modules
+nbgrader/server_extensions/formgrader/static/node_modules/underscore/underscore-esm*
+nbgrader/server_extensions/formgrader/static/node_modules/underscore/underscore-node*
+nbgrader/server_extensions/formgrader/static/node_modules/underscore/underscore-umd*
```
shreve, line 112: Because we're using the exceptions below, we can replace all of these specific rules with a more general one:

```
nbgrader/server_extensions/formgrader/static/node_modules/**/*
```

shreve: Ah, I see what you mean. I think the change I want to see goes well beyond the scope of this PR. Thanks for clarifying.

PR #1978, opened 2026-01-09
Fix Path Traversal in validate_assignment via Path Normalization

yueyueL:
## Description
This PR fixes a path traversal vulnerability in `nbgrader/server_extensions/validate_assignment/handlers.py`.

Following the merge of #1984 (Python ≥3.10), the fix has been updated to use `pathlib` as suggested.

## Changes
- Replaced `os.path.normpath` approach with `pathlib.Path` for path resolution.
- Used `Path.resolve()` to handle `../` sequences and symlinks.
- Used `Path.is_relative_to()` to ensure the path stays within `root_dir`.
- Requests outside the root directory now raise a `403 Access denied` error.

## Testing
- Valid notebook paths within `root_dir` work correctly.
- Malicious paths (e.g., `../../../etc/passwd`) are blocked with a 403 response.
- Updated tests to reflect the new pathlib-based implementation.

### DIFF
--- a/nbgrader/server_extensions/validate_assignment/handlers.py
+++ b/nbgrader/server_extensions/validate_assignment/handlers.py
@@ -38,7 +38,10 @@ class ValidateAssignmentHandler(JupyterHandler):
         return app.config
 
     def validate_notebook(self, path):
-        fullpath = os.path.join(self.root_dir, path)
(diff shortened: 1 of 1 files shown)


brichet (approved):
Thanks @yueyueL, it looks good to me.

